PR #103, opened 2017-08-14
Use non-binary patching to work around different line endings than we expect.

crummel:
Default git behavior has changed over time and people can have different `core.eol` and `core.autocrlf` settings, which can cause issues when we try to apply patches.  This change makes us use text patching instead of binary, which handles differing line endings.


karajas:
I think it's safe to assume we will never have binary files that need to be patched in.

Issue #100, opened 2017-08-13
CI is broken due to missing libfxr


karajas:
/cc @crummel @dleeapho 

This seems to be due to the missing libhost*.so files that reproed locally also. Jenkins CI looks for strings that match error messages.

karajas:
Updating title to reflect missing file.

karajas:
Think I have a fix, it seems to be a pathing issue. Testing now,

karajas:
This is due to the csproj not publishing the fixed version assets of the host and hostresolver for the specific rid.
After trying to fix the publish step without building, i am exploring the option of trying to reuse depproj's instead.

stevedesmond-ca:
Should this be resolved now that #128 is merged?

karajas:
Yup this error should be resolved with that pr

PR #114, opened 2017-08-23
only generate rootfs for cross-compile if it doesn't exist

### LINKED ISSUE #113: Don't generate the rootfs every time when cross-compiling
Steps to reproduce:
1. `sudo ./build.sh /p:Platform=arm`
1. let it build all the way
1. `sudo ./build.sh /p:Platform=arm` again

Expected result:
- Not much should happen, since it just got build and nothing changed

Observed result:
- The rootfs is regenerated right off the bat, which takes a really long time and downloads an entire OS

We should add a condition to only generate the rootfs if it doesn't already exist.

stevedesmond-ca:
Resolves #113.

Sorry for the wonky `Command`, but since the `ROOTFS_DIR` environment variable is passed in as part of the `Command`, we can't use the `Condition` attribute to check for it there.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -30,8 +30,8 @@
   </Target>
 
   <Target Name="GenerateRootFs">
-     <Exec Condition="$(Platform.Contains('arm')) AND '$(Platform)' != 'armel'" Command="$(ArmEnvironmentVariables) sh -c 'if [ ! -d &quot;$ROOTFS_DIR&quot; ]; then ./cross/build-rootfs.sh; fi'" />
-     <Exec Condition="'$(Platform)' == 'armel'" Command="$(ArmEnvironmentVariables) sh -c 'if [ ! -d &quot;$ROOTFS_DIR&quot; ]; then ./cross/armel/tizen-build-rootfs.sh; fi'" />
+     <Exec Condition="!Exists($(CrossRootFsDir)) AND $(Platform.Contains('arm')) AND '$(Platform)' != 'armel'" Command="$(ArmEnvironmentVariables) ./cross/build-rootfs.sh;" />
+     <Exec Condition="!Exists($(CrossRootFsDir)) AND '$(Platform)' == 'armel'" Command="$(ArmEnvironmentVariables) ./cross/armel/tizen-build-rootfs.sh" />
   </Target>
 
   <Target Name="BuildDummyPackages">
--- a/build.proj
+++ b/build.proj
@@ -30,8 +30,8 @@
   </Target>
 
   <Target Name="GenerateRootFs">
-     <Exec Condition="$(Platform.Contains('arm')) AND '$(Platform)' != 'armel'" Command="$(ArmEnvironmentVariables) ./cross/build-rootfs.sh" />
-     <Exec Condition="'$(Platform)' == 'armel'" Command="$(ArmEnvironmentVariables) ./cross/armel/tizen-build-rootfs.sh" />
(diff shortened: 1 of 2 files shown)


omajid:
Might be a bit simpler to patch `build-rootfs.sh` to not rebuild rootfs?

stevedesmond-ca:
Maybe, though I'm assuming there's an implicit goal of keeping the cross
directory in-line with how all the submodules do it. Also, it might be an
unexpected/breaking change that calling the build-rootfs.sh script
explicitly is a no-op if the rootfs directory exists.

omajid:
Fair enough. Thanks for clarifying.

stevedesmond-ca:
Alternatively, we can try refactoring the below lines in `dir.props` to expose `ROOTFS_DIR`, so it can be used in the `Condition`:
https://github.com/dotnet/source-build/blob/f4cdf7fa8cc786dea41c4760cd34682fdc8ae6d6/dir.props#L61-L64

Let me know if you think that's cleaner.

stevedesmond-ca:
Also, is it possible to get access to the CI result details (or make them public) so I can see why things are failing?

tannergooding:
@ellismg, looks like the repolist file wasn't updated when the repo was made public. It is still listed as Private.

jaredpar:
CC @dleeapho about the CI results being private.

stevedesmond-ca:
Thinking about it more last night, I just pushed a change that does what I suggested [above](https://github.com/dotnet/source-build/pull/114#issuecomment-324469113) -- I believe it is much cleaner now.

Note that, along with my comment about keeping in-line with the submodules' cross-compile workflow, the default rootfs has been moved to `./cross/rootfs/{platform}` -- now that it's not intended to be as transient, I believe this is a more appropriate location than the intermediate path.

jaredpar:
@stevedesmond-ca I just changed CI so it should be publicly visible now. I'm unsure if Jenkins will pick that up for an existing PR though. It's possible closing / opening the PR will cause it to make the CI visible. But it may take opening a PR for it to all hook up.

stevedesmond-ca:
Thanks, @jaredpar -- I'm not seeing any change in visibility when I click the `Details` CI link after closing and reopening this PR...should I close this and create an identical new PR?

tannergooding:
@stevedesmond-ca, the generators are still waiting to run. I'll ping this thread once they've finished

tannergooding:
Generator is blocked on https://github.com/dotnet/source-build/pull/116

tannergooding:
They should show now.

jaredpar:
@tannergooding I checked in an incognito window and couldn't see them. I think we may have to opn a new PR to see it.

tannergooding:
@jaredpar, they don't display at all if you aren't even logged in.

tannergooding:
test this please

tannergooding:
@stevedesmond-ca, all 6 jobs should be visible now. Please let me know if that isn't the case.

omajid:
I don't work at Microsoft and I can see them now. Thanks for fixing CI!

jaredpar:
@omajid great! thanks for verifying

stevedesmond-ca:
It looks like all the Linux CI builds are failing on all open PRs for the same reasons. Should those be ignored for now, or do we want to get those working (separate PR) before merging anything else in?

dleeapho:
@karajas is actively on the CI issue #100

stevedesmond-ca:
@dotnet-bot run tests again please :)

crummel:
@dotnet-bot test this please

weshaggard:
@stevedesmond-ca thanks for the interest in helping improve this but I've merged https://github.com/dotnet/source-build/pull/158 which should make this incremental along with all the other initialization steps like ApplyPatches, build dummy packages, etc.

PR #124, opened 2017-08-30
Remove PR legs we currently expect to fail.

crummel:
These will be re-enabled as we fix issues.
Leave the non-PR legs in for now as these are less visible.


crummel:
@dotnet-bot test ci please

PR #128, opened 2017-09-01
Add patches to unblock CI.

karajas:
This changeset introduces the fixes required to get a clean master and a passing CI. The struggle mainly comes from the fact that there are now 2.0.0 versioned packages from products in the official feeds.

### DIFF
--- a/patches/cli/0012-allow-intermediate-package-cache.patch
+++ b/patches/cli/0012-allow-intermediate-package-cache.patch
@@ -1,0 +1,52 @@
+From d8d39ebf8c98a5339089cb89d5838e4ece71780f Mon Sep 17 00:00:00 2001
+From: karajas <karajas@microsoft.com>
+Date: Wed, 6 Sep 2017 18:33:13 -0400
+Subject: [PATCH] Allow cli to restore to an intermdiate package cache to pick up source built packages 
+
+---
+ build/Prepare.targets | 22 +++++++++++++++++++++-
+ 1 file changed, 21 insertions(+), 1 deletion(-)
+
+diff --git a/build/Prepare.targets b/build/Prepare.targets
+index 214b1f7..b94b60a 100644
+--- a/build/Prepare.targets
++++ b/build/Prepare.targets
+@@ -83,6 +83,26 @@
+                    ProjectPath="&quot;%(RestoreToolsPackagesInput.FullPath)&quot;"
+                    AdditionalParameters="/p:UsePortableLinuxSharedFramework=$(UsePortableLinuxSharedFramework)" />
+ 
++
++      <PropertyGroup>
++        <IntermediateToolsPackages>$(MSBuildThisFileDirectory)packages</IntermediateToolsPackages>
++        <FilesToCopyPattern>$(MSBuildThisFileDirectory)packages/**/*</FilesToCopyPattern>
++      </PropertyGroup>
++  
++      <DotNetRestore ToolPath="$(Stage0Directory)"
++                     ProjectPath="&quot;$(RepoRoot)/tools/CrossGen.Dependencies/CrossGen.Dependencies.csproj&quot;"
++                     Source="$(SourceBuiltPackages)"
++                     Runtime="$(SharedFrameworkRid)"
++                     Packages="$(IntermediateToolsPackages)"
++                     AdditionalParameters="/p:UsePortableLinuxSharedFramework=$(UsePortableLinuxSharedFramework)" />
++                     
++      <ItemGroup>
++        <FilesToCopy Include="$(FilesToCopyPattern)" />
++      </ItemGroup>
++  
++    <Message importance="high" text="KARTHIKDEBUG '@(FilesToCopy)' '$(NuGetPackagesDir)'" />
++    <Copy SourceFiles="@(FilesToCopy)" OverwriteReadOnlyFiles="true" DestinationFolder="$(NuGetPackagesDir)/%(RecursiveDir)"/>
++
+   </Target>
+ 
+   <Target Name="CleanToolsLockFiles" >
+@@ -94,7 +114,7 @@
+ 
+   <Target Name="SetupRestoreToolsPackagesInputsOutputs">
+     <ItemGroup>
+-      <RestoreToolsPackagesInput Include="$(RepoRoot)/tools/**/*.csproj" />
++      <RestoreToolsPackagesInput Include="$(RepoRoot)/tools/**/*.csproj" Exclude="$(RepoRoot)/tools/CrossGen.Dependencies/CrossGen.Dependencies.csproj" />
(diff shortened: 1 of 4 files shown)


karajas:
fyi @crummel @dleeapho

karajas:
This got past the core-setup error. I added the xliff tasks that Tom added to his PR. Trying to determine the cause of failure for sdk.

```15:48:33 (ApplyPatches target) -> 
15:48:33   EXEC : error : patch failed: build/DependencyVersions.props:7 [/mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/targets/repository.proj]
15:48:33   EXEC : error : build/DependencyVersions.props: patch does not apply [/mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/targets/repository.proj]
15:48:33   /mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/targets/repository.proj(49,5): error MSB3073: The command "git apply --ignore-whitespace --whitespace=nowarn /mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/patches/sdk/0002-Unify-versions.patch" exited with code 1.
```

karajas:
Newest error happens in cli
```
/mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/src/cli/src/redist/redist.csproj(230,5): error MSB6006: "chmod" exited with code 1. [/mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/targets/repository.proj]
17:02:01   /mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/src/cli/src/redist/redist.csproj(230,5): error MSB6006: "chmod" exited with code 1. [/mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/targets/repository.proj]
17:02:01   /mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/targets/repository.proj(75,5): error MSB3073: The command "/mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/src/cli/build.sh '/p:CLITargets="Prepare;Compile;Package"' /v:detailed /p:SkipBuildingInstallers=true /p:IncludeNuGetPackageArchive=false /p:IncludeAdditionalSharedFrameworks=false /flp:v=detailed /p:CoreSetupBlobRootUrl=file:///mnt/resource/j/workspace/dotnet_source-build/master/RHEL7.2_Release_prtest/bin/obj/x64/Release/blobs/ /p:UsePortableLinuxSharedFramework=false /p:IncludeSharedFrameworksForBackwardsCompatibilityTests=false /p:IncludeAspNetCoreRuntime=false /p:IncludeMSBuildSdkResolver=false" exited with code 1.
```

karajas:
Passed - https://ci.dot.net/job/dotnet_source-build/job/master/job/RHEL7.2_Release_prtest/15/console

karajas:
This passed, just rebased the commits into a more readable manner.

patches/core-setup/0008-add-intermediate-packages.patch
```diff
@@ -0,0 +1,106 @@
...
+diff --git a/src/sharedFramework/framework/framework.csproj b/src/sharedFramework/framework/framework.csproj
+index 52e6553..d4bcf1d 100644
+--- a/src/sharedFramework/framework/framework.csproj
++++ b/src/sharedFramework/framework/framework.csproj
+@@ -4,4 +4,8 @@
+     <OutputType>Exe</OutputType>
+   </PropertyGroup>
+ 
++  <ItemGroup>
++    <PackageReference Include="Microsoft.NETCore.Platforms" Version="2.0.0" />
```
eerhardt: Why is this necessary?  It shouldn't be.  This project already reference Microsoft.NETCore.App, which references this package.

karajas: This isn't necessary but adding a direct reference, helped me understand which Platforms was being included in the assets file in an isolated restore. I can remove this.

eerhardt: Yes, please remove this.

patches/cli/0012-allow-intermediate-package-cache.patch
```diff
@@ -0,0 +1,52 @@
...
+                    ProjectPath="&quot;%(RestoreToolsPackagesInput.FullPath)&quot;"
+                    AdditionalParameters="/p:UsePortableLinuxSharedFramework=$(UsePortableLinuxSharedFramework)" />
+ 
++
++      <PropertyGroup>
++        <IntermediateToolsPackages>$(MSBuildThisFileDirectory)packages</IntermediateToolsPackages>
++        <FilesToCopyPattern>$(MSBuildThisFileDirectory)packages/**/*</FilesToCopyPattern>
++      </PropertyGroup>
++  
++      <DotNetRestore ToolPath="$(Stage0Directory)"
```
eerhardt: This is a pretty harsh hack, and it doesn't address the underlying problem.

The real problem here is that during `source-build`, we are getting NuGet cache polluted with packages that weren't built during `source-build`.  For example, Microsoft already built and shipped `Microsoft.NETCore.App v2.0.0`.  But when building `source-build`, it ALSO produces `Microsoft.NETCore.App v2.0.0`.  These two packages have the same identity, but they have different contents in them.

During the build of `source-build`, the Microsoft shipped package of `Microsoft.NETCore.App v2.0.0` is coming from nuget.org and getting put in the NuGet package cache.  Later, when we try to use the package we built, it sees it already has `Microsoft.NETCore.App v2.0.0` in the cache and doesn't use the one built from the repo.

We need to ensure the build is always using the package that just got built.  We will get the wrong assets in places if we don't - for example, for the native built assets we can get the linux-x64 portable version instead of the distro-specific version that was just built.

karajas: Agree, this was a quick workaround NOT the final solution, I wanted to unblock folks who were looking into this. The tough part is the versioning being the same for both the package being built in source-build and the one being produced officially. I'll open up an issue about this. Do you have any suggestions moving forward to address this conflict?

eerhardt: One suggestion would be that we use different package cache locations for "tools" and for "product" packages.

ex. When restoring BuildTools, or any other NuGet packages that come from external sources, use a directory named `externalPackages`.  When restoring actual product projects that depend on packages that were built earlier in the source-build process, use a directory named `productPackages`.

The names can be anything, as long as they are different.  But the idea being split the package cache for packages that are "tooling/bootstrapping" and packages that are used in the end product.

Issue #105, opened 2017-08-19
Build broken on Windows 10 platform


4creators:
Build is broken on Windows platform. 

After cloning repo for commit 807953db8911c2a8b320b372a4f2ff74c2452046

running .\build.ps1 throws:

```
PowerShell version: 5.1.15063.502
Free physical memory: 11573 MB out of 16300 MB

Loading personal and system profiles took 1746ms.
PS C:\> E:
PS E:\> cd src\ms\dotnet\src-bld
PS E:\src\ms\dotnet\src-bld> .\build.ps1
Get-Content : Cannot find path 'E:\src\ms\dotnet\src-bld\.cliversion' because it does not exist.
At E:\src\ms\dotnet\src-bld\build.ps1:5 char:15
+ $SdkVersion = Get-Content (Join-Path $ScriptRoot ".cliversion")
+               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (E:\src\ms\dotnet\src-bld\.cliversion:String) [Get-Content], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetContentCommand

& : The term 'E:\src\ms\dotnet\src-bld\bootstrap\bootstrap.ps1' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path
was included, verify that the path is correct and try again.
At E:\src\ms\dotnet\src-bld\build.ps1:7 char:3
+ & "$ScriptRoot\bootstrap\bootstrap.ps1"
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (E:\src\ms\dotne...p\bootstrap.ps1:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

& : The term 'E:\src\ms\dotnet\src-bld\Tools\dotnetcli\dotnet' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path
was included, verify that the path is correct and try again.
At E:\src\ms\dotnet\src-bld\build.ps1:8 char:3
+ & "$ScriptRoot\Tools\dotnetcli\dotnet" restore "$ScriptRoot\tasks\Mic ...
[...]

PR #147, opened 2017-09-14
Fixing the build on windows.

eerhardt:


dagood:
Here's an issue tracking building on Win10 that this should fix: https://github.com/dotnet/source-build/issues/105.

build.ps1
```diff
@@ -1,12 +1,20 @@
...
 
-& "$ScriptRoot\bootstrap\bootstrap.ps1"
-& "$ScriptRoot\Tools\dotnetcli\dotnet" restore "$ScriptRoot\tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj"
-& "$ScriptRoot\Tools\dotnetcli\dotnet" build "$ScriptRoot\tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj"
-& "$ScriptRoot\Tools\dotnetcli\dotnet" "$ScriptRoot\Tools\dotnetcli\sdk\$SdkVersion\MSBuild.dll" "$ScriptRoot\build.proj" @args /t:WriteDynamicPropsToStaticPropsFiles /p:GeneratingStaticPropertiesFile=true
-& "$ScriptRoot\Tools\dotnetcli\dotnet" "$ScriptRoot\Tools\dotnetcli\sdk\$SdkVersion\MSBuild.dll" "$ScriptRoot\build.proj" @args
+Remove-Item -Recurse -Force "$NUGET_PACKAGES"
 
+& "$CLIPATH\dotnet" "$SDKPATH/MSBuild.dll" "$SCRIPT_ROOT/build.proj" "$@" /t:WriteDynamicPropsToStaticPropsFiles /p:GeneratingStaticPropertiesFile=true
+& "$CLIPATH\dotnet" "$SDKPATH/MSBuild.dll" "$SCRIPT_ROOT/build.proj" "$@" /t:GenerateRootFs
```
dagood: I think these `"$@"`s should be `@args`.

crummel:
Could you also re-enable Windows builds with this PR?  i.e. partially revert 2acdb825f022544d5759d7b1a2db80e69235a939?  I will look into why CI is still running the old ARM leg.  Other than that LGTM.

build.ps1
```diff
@@ -1,12 +1,20 @@
...
-$SdkVersion = Get-Content (Join-Path $ScriptRoot ".cliversion")
+& "$CLIPATH\dotnet" restore "$SCRIPT_ROOT\tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj"
+& "$CLIPATH\dotnet" build "$SCRIPT_ROOT\tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj"
 
-& "$ScriptRoot\bootstrap\bootstrap.ps1"
-& "$ScriptRoot\Tools\dotnetcli\dotnet" restore "$ScriptRoot\tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj"
-& "$ScriptRoot\Tools\dotnetcli\dotnet" build "$ScriptRoot\tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj"
-& "$ScriptRoot\Tools\dotnetcli\dotnet" "$ScriptRoot\Tools\dotnetcli\sdk\$SdkVersion\MSBuild.dll" "$ScriptRoot\build.proj" @args /t:WriteDynamicPropsToStaticPropsFiles /p:GeneratingStaticPropertiesFile=true
-& "$ScriptRoot\Tools\dotnetcli\dotnet" "$ScriptRoot\Tools\dotnetcli\sdk\$SdkVersion\MSBuild.dll" "$ScriptRoot\build.proj" @args
+Remove-Item -Recurse -Force "$NUGET_PACKAGES"
```
weshaggard: Why do we need to remove the packages here?

build.ps1
weshaggard: We don't need to call GenerateRootFs on windows.

build.ps1
eerhardt: The quick answer: Because we do in `build.sh`.

The longer answer relates to our discussion today.  Restoring BuildTools will pollute the NUGET_PACKAGES cache by installing packages that we are going to build ourselves during the build.  And then later, when downstream repos want to consume the package, they will get the one that was restored by BuildTools (which isn't what we built).

weshaggard: OK I'm fine with consistency between build.cmd/sh for now but we should isolate these and not have to clean these like this.

build.ps1
```diff
@@ -1,12 +1,20 @@
...
 
+& "$CLIPATH\dotnet" "$SDKPATH/MSBuild.dll" "$SCRIPT_ROOT/build.proj" "$@" /t:WriteDynamicPropsToStaticPropsFiles /p:GeneratingStaticPropertiesFile=true
```
jaredpar: not checking for failure on any of these commands

eerhardt: Fixed by copying the way https://github.com/dotnet/roslyn/blob/master/build/scripts/build.ps1 does it.

build.ps1
eerhardt: Removed.

eerhardt: doh

eerhardt:
@dotnet-bot test ci please

eerhardt:
I'm going to merge this to make progress.  I can follow up any remaining feedback in my next change.

weshaggard:
Thanks @eerhardt with this I assume the windows leg builds now? Are you planning to add it to CI?

eerhardt:
It has been added to CI now, but the builds still aren't clean.  This change got the general "build.cmd/ps1" working.  Now there are error when building individual repos.  The first one I'm working through is: https://github.com/dotnet/source-build/pull/148

PR #155, opened 2017-09-21
Move corefx to latest release/2.0.0 commit

weshaggard:
Remove unnecessary patch and update another patch to align with the move.

This will enable corefx to build on windows and should fix https://github.com/dotnet/source-build/issues/106.


weshaggard:
Looks like my move of corefx fixed the windows build for that project but broke the update dependencies on the other OS's. I will investigate further and update the other steps.

PR #158, opened 2017-09-22
Setup an init-build project that can be incremental

weshaggard:
currently the init-build script will do the following:
- Build the local tasks
- Write out the targget props file
- GenerateRootFs
- Apply Patches
- Build dummy packages
- Build reference assembly packages

It will do all of that once and afterwards a clean will be needed
in order for it to run again. Individual targets can also be run
on the init-builds.proj file in order to repeat just parts of the
project steps.

cc @eerhardt


jaredpar:
Can we get a README explaining how to use the repo now?

weshaggard:
>Can we get a README explaining how to use the repo now?

The main steps are still the same just call build.cmd/sh. This just make some things more incremental. It does changes some more advance steps like cleaning or reapplying patches etc, and I agree we should document those step, but I don't think it should be tied to this PR.

jaredpar:
Think we should doc the basic dev worcflow as we change it. Otherwise we end up with a repo that is only usable if you are on the slack channel

weshaggard:
I submitted PR https://github.com/dotnet/source-build/pull/159 to add some basic commands this PR doesn't actually change any of those things only advanced stuff. If you want to see further docs please ask in the other PR.

eerhardt:
>     public static class RuntimeEnvironment

Hmm, I didn't know this code was in here.... It is a duplicate of the code we have in core-setup.

@karajas @crummel - do either of you know what the plan for this code is? 

---
Refers to: tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/vendor/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs:9 in 18e41a6. [](commit_id = 18e41a6f102f64fc9976e55c3685fd3611c7cdce, deletion_comment = False)

eerhardt:
>     public class DiscoverProjectsForTFMs : Task

I don't see this task being used - it is project.json based.  Can it be removed?

---
Refers to: tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/DiscoverProjectsForTFMs.cs:16 in 18e41a6. [](commit_id = 18e41a6f102f64fc9976e55c3685fd3611c7cdce, deletion_comment = False)

eerhardt:
>     public class ChangeRuntimesSection : Task

This task doesn't appear to be used.

---
Refers to: tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/ChangeRuntimesSection.cs:18 in 18e41a6. [](commit_id = 18e41a6f102f64fc9976e55c3685fd3611c7cdce, deletion_comment = False)

eerhardt:
>     public class UpdateProjectJsonFrameworkDependencyVersions : Task

Can this be removed?

---
Refers to: tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UpdateProjectJsonFrameworkDependencyVersions.cs:16 in 18e41a6. [](commit_id = 18e41a6f102f64fc9976e55c3685fd3611c7cdce, deletion_comment = False)

eerhardt:
>     public class AddPublicSignToProjectJson : Task

project.json based - can this be removed?

---
Refers to: tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/AddPublicSignToProjectJson.cs:16 in 18e41a6. [](commit_id = 18e41a6f102f64fc9976e55c3685fd3611c7cdce, deletion_comment = False)

tools-local/init-build.proj
```diff
@@ -0,0 +1,323 @@
...
+  </Target>
+
+  <Target Name="Clean">
+    <Delete Files="$(BuildCompetedSuccessSemaphore)" />
+  </Target>
+
+  <Target Name="BuildTasks">
+    <Exec Command="$(DotNetCliToolDir)dotnet build tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj" />
+    <!-- Delete the packages directory to avoid polluting the repo builds -->
+    <RemoveDir Directories="$(ProjectDir)packages" />
```
eerhardt: Do we have a property for this path?  If not, should we make one?

tools-local/init-build.proj
```diff
@@ -0,0 +1,323 @@
...
+    <CallTarget Targets="BuildTasks;WriteDynamicPropsToStaticPropsFiles;GenerateRootFs;ApplyPatches;BuildDummyPackages;BuildReferenceAssemblies" />
+    <Touch Files="$(BuildCompetedSuccessSemaphore)" AlwaysCreate="true" />
+  </Target>
+
+  <Target Name="Clean">
+    <Delete Files="$(BuildCompetedSuccessSemaphore)" />
+  </Target>
+
+  <Target Name="BuildTasks">
+    <Exec Command="$(DotNetCliToolDir)dotnet build tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj" />
```
eerhardt: Does this project need to be restored first?  or are we using a new enough CLI that restore will implicitly happen?

tools-local/init-build.proj
```diff
@@ -0,0 +1,323 @@
...
+      <DummyPackage Include="xunit.extensibility.execution" Version="2.2.0" />
+      <DummyPackage Include="xunit.runner.visualstudio" Version="2.2.0" />
+    </ItemGroup>
+
+    <BuildDummyPackages Packages="@(DummyPackage)"
+                        OutputPath="$(SourceBuiltPackagesPath)" />
+  </Target>
+
+  <Target Name="BuildReferenceAssemblies">
+    <Exec Command="$(DotNetCliToolDir)/dotnet build $(SubmoduleDirectory)/reference-assemblies/build.proj" />
```
eerhardt: Since we are using `$(DotNetCliToolDir)/dotnet` to build this project, can these just be `<MSBuild` tasks?  Same for the ones above.

tools-local/init-build.proj
weshaggard: From my testing we don't need to restore so I assume we have a new enough CLI.

tools-local/init-build.proj
weshaggard: We don't currently have a property for it from my quick look I was a little surprised but I also didn't see any other places where we use this path that could benefit from the property. At this point I'm unclear actually who is setting that path.

tools-local/init-build.proj
weshaggard: I didn't attempt to change these but I suspect we will have the same issue I had with the task project build where it cannot successfully compile when the root is building via `dotnet msbuild.dll` as opposed to `dotnet msbuild`

eerhardt: oh right.  good point.

tools-local/init-build.proj
eerhardt: The way it works today is that we are setting the NUGET_PACKAGES env var, and expecting everyone to respect it:

https://github.com/dotnet/source-build/blob/dev/release/2.0/build.sh#L16
https://github.com/dotnet/source-build/blob/dev/release/2.0/patches/corefx/0001-Respect-NUGET_PACKAGES-if-it-is-set.patch  (and others)

weshaggard: That's write they are in the root scripts which is why we cannot easily share them. We should definitely working on moving that to the msbuild side of the house.

PR #159, opened 2017-09-22
Update some basic commands in the documentation

weshaggard:
@jaredpar @eerhardt 

These are the basic commands but please let me know if you want to see other things documented. 

skip ci

### DIFF
--- a/Documentation/README.md
+++ b/Documentation/README.md
@@ -6,16 +6,39 @@ The following document includes documentation for the build scripts. Over time, 
 
 The scripts are supported on Windows, macOS and Linux. The scripts are based on PowerShell on Windows and Bash on macOS and Linux.
 
-### Build on Windows
+### Syncing
 
-```console
-./build.ps1
+This repo uses submodules so remember on fresh clones or when updating/switching branches to run
+
+```
+git submodule update --init --recursive
 ```
 
-### Build on Linux or macOS
+### Building
 
 ```console
-./build.sh
+./build.{cmd|sh}
+```
+
+## Building one repo
+
+By default we build the cli and its dependencies but the default root repository to build can be passed in.
+
+```console
+./build.{cmd|sh} /p:DefaultRepository=core-setup
+```
+
+Sometimes you want to just iterate on a single repo build and not rebuild all depedencies that can be done passing another property.
+
+```console
+./build.{cmd|sh} /p:DefaultRepository=core-setup /p:BuildDependentRepositories=false
+```
+
+### Cleaning
+
+```console
+./clean.{cmd|sh}      # Cleans root binary directory and fully git cleans and hard resets all submodules
+./clean.{cmd|sh} -a   # Does a full git clean on the root repo as well as fully git cleans and hard resets all submodules
 ```
 
 ## Building .NET Core SDK and a Source tarball


Documentation/README.md
```diff
@@ -6,16 +6,25 @@ The following document includes documentation for the build scripts. Over time,
...
+./build.{cmd|sh}
 ```
 
-### Build on Linux or macOS
+### Cleaning
 
 ```console
-./build.sh
+./clean.{cmd|sh}      # Cleans root binary directory and resets all submodules
+./clean.{cmd|sh} -a   # Does a fill git clean on the root repo as well as resets all submodules
```
eerhardt: type-o `fill`

Documentation/README.md
```diff
@@ -6,16 +6,25 @@ The following document includes documentation for the build scripts. Over time,
 
 The scripts are supported on Windows, macOS and Linux. The scripts are based on PowerShell on Windows and Bash on macOS and Linux.
 
-### Build on Windows
+### Syncing
+
+This repo uses submodules so remember on fresh clones or when updating/switching branches to run
+
+```
+git submodule update --init
```
crummel: NuGet - this should include the `--recursive` flag.

Documentation/README.md
```diff
@@ -6,16 +6,25 @@ The following document includes documentation for the build scripts. Over time,
...
+```
+
+### Building
 
 ```console
-./build.ps1
+./build.{cmd|sh}
```
crummel: Do we want to mention that parameters will pass through to MSBuild?

Documentation/README.md
```diff
@@ -6,16 +6,25 @@ The following document includes documentation for the build scripts. Over time,
 
 The scripts are supported on Windows, macOS and Linux. The scripts are based on PowerShell on Windows and Bash on macOS and Linux.
 
-### Build on Windows
+### Syncing
+
+This repo uses submodules so remember on fresh clones or when updating/switching branches to run
```
dagood: The doc should probably also mention `git submodule sync [--recursive]` for when `.gitmodules` changes to point to different forks.

Documentation/README.md
```diff
@@ -6,16 +6,25 @@ The following document includes documentation for the build scripts. Over time,
...
-./build.ps1
+./build.{cmd|sh}
 ```
 
-### Build on Linux or macOS
+### Cleaning
 
 ```console
-./build.sh
+./clean.{cmd|sh}      # Cleans root binary directory and resets all submodules
```
dagood: I'm not sure how concerned to be about protecting people from accidentally cleaning away their work by only reading the readme, but "resets" is kind of vague. "fully git cleans and hard resets" would be explicit.

For best (imo) user experience, we would have `clean.sh/cmd` by default only run `git clean -xdn`, which lists everything it *would* delete (and do something similar for the `reset --hard`), then prompt. Then we could add `clean.sh/cmd -f` to do it without prompting.

Documentation/README.md
weshaggard: I'm not sure that will be a common issue so we can keep it out for now.

Documentation/README.md
weshaggard: I don't really like the two stage process. However if folks end up hitting trouble then we can reconsider.

PR #182, opened 2017-09-25
Remove standard repo patches and move sha forward

weshaggard:
cc @eerhardt 

Relies on https://github.com/dotnet/standard/pull/511 but the submodule pointer already points to that commit.


targets/standard.props
```diff
@@ -15,8 +15,6 @@
     <EnvironmentVariables Include="LatestCommit=$(LatestCommit)" />
     <EnvironmentVariables Include="OfficialBuildId=$(OfficialBuildId)" />
     <EnvironmentVariables Include="VersionSeedDate=$(OfficialBuildId)" />
-    <EnvironmentVariables Include="OverridePackageSource=$(SourceBuiltPackagesPath)" />
-    <EnvironmentVariables Include="OverridePackageSourceEx=$(PrebuiltPackagesPath)" Condition="'$(OfflineBuild)' == 'true'" />
-    <EnvironmentVariables Include="ExcludeInternetFeeds=true" Condition="'$(OfflineBuild)' == 'true'" />
+    <EnvironmentVariables Include="DotNetSourceBuildRestoreSources=$(SourceBuiltPackagesPath)" />
```
eerhardt: I think you should still support the OfflineBuild scenario.  Basically add `$(PrebuiltPackagesPath)` to `DotNetSourceBuildRestoreSources` if it is true.  And also set `$(DotNetBuildOffline)` when `$(OfflineBuild)` is true.

eerhardt:
>     <LatestCommit>d318b764a689cfe285d8484bda918646905664e7</LatestCommit>

Do we need to update these?  What are they for?

---
Refers to: targets/standard.props:9 in a3c8947. [](commit_id = a3c8947d83a1757d95b6ddc5cef9977af0dc2604, deletion_comment = False)

targets/standard.props
```diff
@@ -2,7 +2,7 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <ProjectDirectory Condition="'$(PathToRepo)' != ''">$(PathToRepo)</ProjectDirectory>
-    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -$(Configuration)</BuildCommand>
+    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -$(Configuration) -- /p:DotNetBuildFromSource=true</BuildCommand>
```
eerhardt: Today this is set in the dir.props EnvironmentVariables list.  I was assuming we'd move these all to properties when we started using the official repo api stuff.  If we started adding these now, we would just need to change it again everywhere later.

Thoughts?

targets/standard.props
weshaggard: Yeah I was thinking about that but wasn't sure if the semacolon delimiter would make it through to the env variable. 

As for the offline build my assumption is we would set it globally, although I can do that with this PR.

targets/standard.props
weshaggard: Fair I can do that.

weshaggard:
>Do we need to update these? What are they for?

Not sure what those are used for I'm pretty sure they are outdated at any rate. @crummel do you know if those sha's are used by the auto-update tool?

crummel:
I don't believe they are currently used at all, @dagood might know if they ever were.

dagood:
I think the idea was to tell the repo what commit it's at when `git` isn't an allowed dependency and/or when we don't have the `.git` data, for sourcelinking, embedding in DLL metadata, things like that.

I don't know how the property specifically is/was used for that, though. I can say for sure that auto-update doesn't need it--the updater depends on `git`.

PR #189, opened 2017-09-27
Change clean-submodules to clean in build-source-tarball.sh.

crummel:
Clean-submodules.sh has been replaced by clean.sh

### DIFF
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -27,11 +27,11 @@ if [ $SKIP_BUILD -ne 1 ]; then
         rm -rf "$SCRIPT_ROOT/bin"
     fi
 
-    $SCRIPT_ROOT/clean-submodules.sh
+    $SCRIPT_ROOT/clean.sh -a
     $SCRIPT_ROOT/build.sh /p:ArchiveDownloadedPackages=true /flp:v=detailed
 fi
 
-$SCRIPT_ROOT/clean-submodules.sh
+$SCRIPT_ROOT/clean.sh -a
 
 mkdir -p "$TARBALL_ROOT"
 


crummel:
cc @dseefeld

build-source-tarball.sh
```diff
@@ -27,11 +27,11 @@ if [ $SKIP_BUILD -ne 1 ]; then
         rm -rf "$SCRIPT_ROOT/bin"
     fi
 
-    $SCRIPT_ROOT/clean-submodules.sh
+    $SCRIPT_ROOT/clean.sh -a
     $SCRIPT_ROOT/build.sh /p:ArchiveDownloadedPackages=true /flp:v=detailed
 fi
 
-$SCRIPT_ROOT/clean-submodules.sh
+$SCRIPT_ROOT/clean.sh -a
```
dseefeld: -a doesn't seem correct, because later in the script, the Tools dir is copied to $TARBALL_ROOT and -a removes it.

crummel: You're right, I think this breaks with `--skip-build` on.  I used `-a` because clean.sh no longer has this behavior with `-a` off:
```
git clean -dxf "$SCRIPT_ROOT/src/javascriptservices/"
git clean -dxf "$SCRIPT_ROOT/src/netcorecli-fsc/"
git clean -dxf "$SCRIPT_ROOT/src/reference-assemblies/"
```
It looks like we do want that behavior, as later we copy from these directories as well.  I might need to add an option to clean.sh - say `-k` for clean-well-known?  @weshaggard

weshaggard: I would prefer you just fix the Clean target in build.proj the correctly clean these directories instead of adding another option to the clean script.

crummel: Ok, I'll take a look at that.

weshaggard: Ideally those sub-directories wouldn't be dropping files in the source tree and instead the paths of their outputs should be reset to be under the bin folder in the root and then it would bet cleaned when you clean the bin folder. If for some reason that cannot be done then doing a git clean on those sub-dirs might be acceptable until those issues can be addressed.

eerhardt:
test RHEL7.2 Release - want to see the current state of the repo

eerhardt:
test RHEL7.2 Release - want to see the current state of the repo

eerhardt:
test RHEL7.2 Release - want to see the current state of the repo

Last one I promise.  😄

repos/javascriptservices.proj
```diff
@@ -2,8 +2,8 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
   <PropertyGroup>
-    <BuildCommand>$(DotnetToolCommand) pack -c $(Configuration) --no-build Microsoft.DotNet.Web.Spa.ProjectTemplates.csproj /p:NuspecFile=Microsoft.DotNet.Web.Spa.ProjectTemplates.nuspec</BuildCommand>
-    <PackagesOutput>$(ProjectDirectory)/bin/$(Configuration)/</PackagesOutput>
+    <BuildCommand>$(DotnetToolCommand) pack -c $(Configuration) --output $(SourceBuiltPackagesPath) --no-build Microsoft.DotNet.Web.Spa.ProjectTemplates.csproj /p:NuspecFile=Microsoft.DotNet.Web.Spa.ProjectTemplates.nuspec /p:MSBuildProjectExtensionsPath=$(IntermediatePath)$(RepositoryName)/ /p:IntermediateOutputPath=$(IntermediatePath)$(RepositoryName)/ /p:OutputPath=$(OutputPath)$(RepositoryName)/ /p:RestoreOutputPath=$(IntermediatePath)$(RepositoryName)</BuildCommand>
```
eerhardt: I think you can just set `BaseIntermediateOutputPath` instead of having to set `MSBuildProjectExtensionsPath`, `IntermediateOutputPath`, `RestoreOutputPath`.

src/reference-assemblies/build.proj
```diff
@@ -1,15 +1,15 @@
...
   <Target Name="Restore" />
   <Target Name="Build">
+    <Copy SourceFiles="$(MSBuildProjectDirectory)/_._" DestinationFolder="$(IntermediatePath)reference-assemblies/" />
     <MSBuild Projects="@(AllProjects)" Targets="Build" BuildInParallel="true" />
   </Target>
 
-  <Target Name="Pack">
-    <MSBuild Projects="nuspecs/pack.csproj" Targets="Pack" Properties="NuspecFile=%(AllNuSpecs.Identity)" />
-  </Target>
+  <Target Name="Pack"/>
```
eerhardt: ?? Do we need this target anymore?  Can it just be deleted?

src/reference-assemblies/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
eerhardt: `DefaultTargets="Build"` is unnecessary.  This .props file doesn't contain any targets.

(Technically `ToolsVersion` and `xmlns` are unnecessary as well with MSBuild v15)

src/reference-assemblies/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
...
+    <Import Project="../../dir.props" />
+    <PropertyGroup>
+        <OutputPath>$(IntermediatePath)$(MSBuildProjectDirectoryNoRoot.Substring($(MSBuildThisFileDirectoryNoRoot.LastIndexOf("reference-assemblies"))))</OutputPath>
```
eerhardt: You can extract `$(MSBuildProjectDirectoryNoRoot.Substring($(MSBuildThisFileDirectoryNoRoot.LastIndexOf("reference-assemblies"))))` into a separate property and then reuse it here and in IntermediateOutputPath.

src/reference-assemblies/build.proj
```diff
@@ -1,15 +1,15 @@
 <Project>
+  <Import Project="../../dir.props" />
```
eerhardt: A better way to do this is to say:

`<Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />`

Which tells MSBuild to find the first `dir.props` file in this directory, or any directory above this directory.  This allows for a new `dir.props` file to be added in the directory directly above this file, and this code doesn't need to change - it automatically gets picked up.

src/reference-assemblies/build.proj
dseefeld: No. I removed it.

repos/javascriptservices.proj
dseefeld: Yes.  I changed it to use BaseIntermediateOutputPath.

src/reference-assemblies/Directory.Build.props
dseefeld: Removed it.

src/reference-assemblies/Directory.Build.props
dseefeld: Yes.  I created a separate property.

src/reference-assemblies/build.proj
dseefeld: I saw @weshaggard added in his restructuring PR.  I'll add it here.

PR #195, opened 2017-09-29
Remove corefx patches

weshaggard:
cc @eerhardt


targets/corefx.props
```diff
@@ -2,7 +2,7 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <ProjectDirectory Condition="'$(PathToRepo)' != ''">$(PathToRepo)</ProjectDirectory>
-    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -$(Configuration) -buildArch=$(Platform) -portable=false -- /p:ILLinkTrimAssembly=false /p:BuildAllPackages=true</BuildCommand>
+    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -$(Configuration) -buildArch=$(Platform) -portable=false -allConfigurations -- /p:ILLinkTrimAssembly=false</BuildCommand>
```
eerhardt: To me, it seems like corefx should be checking the `$(DotNetBuildFromSource)` flag here, instead of source-build having to specify `-allConfigurations`.

Do you plan on doing this later?

targets/corefx.props
```diff
@@ -14,11 +14,13 @@
...
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(OfflineBuild)' == 'true'">
     <EnvironmentVariables Include="LatestCommit=$(LatestCommit)" />
     <EnvironmentVariables Include="OfficialBuildId=$(OfficialBuildId)" />
     <EnvironmentVariables Include="VersionSeedDate=$(OfficialBuildId)" />
-    <EnvironmentVariables Include="OverridePackageSource=$(SourceBuiltPackagesPath)" />
-    <EnvironmentVariables Include="OverridePackageSourceEx=$(PrebuiltPackagesPath)" Condition="'$(OfflineBuild)' == 'true'" />
-    <EnvironmentVariables Include="ExcludeInternetFeeds=true" Condition="'$(OfflineBuild)' == 'true'" />
+    <EnvironmentVariables Include="DotNetSourceBuildRestoreSources=$(SourceBuiltPackagesPath)%3B$(PrebuiltPackagesPath)" />
```
eerhardt: This is duplicated in `targets/standard.props`.  Should it be move to a common place?

targets/corefx.props
```diff
@@ -14,11 +14,13 @@
   <ItemGroup>
     <RepositoryReference Include="coreclr" />
     <RepositoryReference Include="standard" />
+    <EnvironmentVariables Include="DotNetSourceBuildRestoreSources=$(SourceBuiltPackagesPath)" Condition="'$(OfflineBuild)' != 'true'" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(OfflineBuild)' == 'true'">
```
eerhardt: I think we want the `LatestCommit` `OfficialBuildId` `VersionSeedDate` parameters even if we aren't building offline.  Why the change?

weshaggard: They are only interesting when building from the tarball. For the LatestCommit we will get the live one from git and for OfficialBuildId/VersionSeedDate we only want that if we are producing an official build and we need it to exactly match one we shipped which again is only interesting for the tar-ball.

targets/corefx.props
weshaggard: I agree and plan to do it later, when we start to solidify the exact command lines we pass to the repos.

targets/corefx.props
weshaggard: Yes it probably is and I will move these into the common place as part of my coreclr work.

weshaggard:
test RHEL7.2 Release

weshaggard:
test RHEL7.2 Release

targets/corefx.props
```diff
@@ -14,11 +14,13 @@
   <ItemGroup>
     <RepositoryReference Include="coreclr" />
     <RepositoryReference Include="standard" />
+    <EnvironmentVariables Include="DotNetSourceBuildRestoreSources=$(SourceBuiltPackagesPath)" Condition="'$(OfflineBuild)' != 'true'" />
```
dagood: To make sure my plan is known: I'm planning on removing this and using the [`DotNetRestoreSourcePropsPath`](https://github.com/dotnet/source-build/blob/dev/release/2.0/Documentation/auto-dependency-flow/api.md#-pdotnetrestoresourcepropspathpath) Repo API when I have my changes through.

Re: https://gitter.im/dotnet/source-build?at=59d29381614889d47567f95b
@eerhardt @weshaggard

weshaggard: Yep.

targets/corefx.props
eerhardt: Since this is a servicing branch, don't you want `/p:BuildAllPackages=true`?  Or does `-allConfigurations` do the same thing?

weshaggard:
test CentOS7.1 Release
test RHEL7.2 Release

weshaggard:
@eerhardt corefx patches are green on RHEL now so I"m going to merge this finally.

PR #202, opened 2017-10-02
Remove core setup patches

eerhardt:


eerhardt:
Test CentOS7.1 Release
Test Windows_NT Release

eerhardt:
PTAL - This is now "green" and I plan to merge soon.

dir.props
```diff
@@ -81,6 +82,7 @@
     <EnvironmentVariables Include="BUILDTOOLS_SKIP_CROSSGEN=1" />
     <!-- New env-var names to replace the ones above over time -->
     <EnvironmentVariables Include="DotNetBuildFromSource=true" />
+    <EnvironmentVariables Include="DotNetRestorePackagesPath=$(PackagesDir)" />
```
dagood: Interesting that the conventions clash here. I see that NuGet's MSBuild code does call this `PackagesPath` in its code. The NuGet tasks don't seem to stick to `Path`/`File`, using `Directory` and even `Folders` too. Annoying that there isn't a solid convention here. I don't think this needs to be addressed in this PR but pointing it out in case it matters for the Repo API.

targets/core-setup.props
```diff
@@ -14,18 +14,17 @@
...
     <RepositoryReference Include="newtonsoft-json" />
     <RepositoryReference Include="standard" />
+    <EnvironmentVariables Include="DotNetRestoreSources=$(SourceBuiltPackagesPath)" Condition="'$(OfflineBuild)' != 'true'" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(OfflineBuild)' == 'true'">
+    <EnvironmentVariables Include="LatestCommit=$(LatestCommit)" />
+    <EnvironmentVariables Include="OfficialBuildId=$(OfficialBuildId)" />
+    <EnvironmentVariables Include="VersionSeedDate=$(OfficialBuildId)" />
+    <EnvironmentVariables Include="DotNetRestoreSources=$(SourceBuiltPackagesPath)%3B$(PrebuiltPackagesPath)" />
```
dagood: `%3B` hasn't worked in my experience without unescaping downstream: https://github.com/NuGet/Home/issues/5817#issuecomment-333857710.

This will be removed anyway when moving to the Repo API, so might not be worth trying to find a fix for now.

PR #215, opened 2017-10-09
Update dependency flow doc for not implemented

eerhardt:
Documenting which properties haven't been implemented yet.

/cc @dotnet/source-build-contrib @jaredpar

### DIFF
--- a/Documentation/auto-dependency-flow/api.md
+++ b/Documentation/auto-dependency-flow/api.md
@@ -3,7 +3,7 @@
 Repos must implement these arguments to participate in the orchestrated dependency flow.
 
 
-# Dependency version input arguments
+# Dependency version input arguments (Not yet implemented)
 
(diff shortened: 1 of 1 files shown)


Documentation/auto-dependency-flow/api.md
```diff
@@ -3,7 +3,7 @@
 Repos must implement these arguments to participate in the orchestrated dependency flow.
 
 
-# Dependency version input arguments
+# Dependency version input arguments (Not yet implemented)
```
dagood: What does it mean for an argument to be implemented? By who?

eerhardt: By the source-build repo.  Consumers of this doc were trying to use these properties, but no one was passing them in yet.

I need to update this doc again after #219

PR #219, opened 2017-10-09
Implement Dependency Flow for standard => corefx

eerhardt:
This is the first round of implementation of https://github.com/dotnet/source-build/tree/dev/release/2.0/Documentation/auto-dependency-flow.

/cc @dotnet/source-build-contrib


dir.props
```diff
@@ -44,13 +44,18 @@
     <LocalBlobStorageRoot>$(IntermediatePath)blobs/</LocalBlobStorageRoot>
     <LocalBuildInfoRoot>$(IntermediatePath)build-info/</LocalBuildInfoRoot>
     <LocalNuGetPackagesRoot>$(IntermediatePath)nuget-packages/</LocalNuGetPackagesRoot>
+    <SourceBuiltBlobFeedDir>$(IntermediatePath)blob-feed/</SourceBuiltBlobFeedDir>
+    <SourceBuiltPackagesPath>$(SourceBuiltBlobFeedDir)packages/</SourceBuiltPackagesPath>
```
jaredpar: Why path here and not dir?

dir.props
```diff
@@ -44,13 +44,18 @@
...
+    <SourceBuiltAssetsDir>$(SourceBuiltBlobFeedDir)assets/</SourceBuiltAssetsDir>
     <PrebuiltPackagesPath>$(ProjectDir)prebuilt/nuget-packages/</PrebuiltPackagesPath>
-    <SourceBuiltPackagesPath>$(IntermediatePath)source-built/</SourceBuiltPackagesPath>
     <SourceBuiltTarBallPath>$(OutputPath)</SourceBuiltTarBallPath>
-    <SourceBuiltSymbolsPackagesPath>$(SourceBuiltPackagesPath)symbols/</SourceBuiltSymbolsPackagesPath>
+    <SourceBuiltSymbolsPackagesPath>$(SourceBuiltAssetsDir)</SourceBuiltSymbolsPackagesPath>
```
jaredpar: Same

targets/newtonsoft-json.props
```diff
@@ -12,5 +12,6 @@
     <BuildPackagesCommand>$(DotNetCommand) pack $(NewtonsoftJsonProjectPath) $(DotNetCommandArguments)</BuildPackagesCommand>
     <CleanCommand>$(DotNetCommand) clean $(NewtonsoftJsonProjectPath) $(DotNetCommandArguments)</CleanCommand>
     <PackagesOutput>$(NewtonsoftJsonDirectory)bin/$(Configuration)/</PackagesOutput>
+    <RepoApiNotImplemented>true</RepoApiNotImplemented>
```
jaredpar: Rather than a negative property i think we should consider a positive one. 

``` xml
<RepoApiImplemented>true</RepoApiImplemented>
```

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/WriteBuildOutputProps.cs
```diff
@@ -0,0 +1,63 @@
+﻿using Microsoft.Build.Framework;
```
jaredpar: Do we require copyright headers in this repo?

targets/fsharp.props
```diff
@@ -7,6 +7,7 @@
     <BuildRevision>170630</BuildRevision>
     <BuildCommand>$(ProjectDirectory)/src/buildfromsource$(ShellExtension)</BuildCommand>
     <NuGetConfigFile>$(ProjectDirectory)/NuGet.Config</NuGetConfigFile>
+    <RepoApiNotImplemented>true</RepoApiNotImplemented>
```
weshaggard: FYI- All these updates to the props file for the repo's is going to clash pretty hard with changes I have locally.

targets/standard.props
```diff
@@ -2,9 +2,8 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <ProjectDirectory Condition="'$(PathToRepo)' != ''">$(PathToRepo)</ProjectDirectory>
-    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -$(Configuration)</BuildCommand>
+    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -$(Configuration) --</BuildCommand>
```
weshaggard: what is the "--" for here?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/WriteBuildOutputProps.cs
weshaggard: yes

targets/standard.props
dagood: This tells run.exe to pass all remaining args to msbuild directly so that repo API args can be passed. (This PR doesn't change the repos to  implement the "core" repo API framework.)

dir.props
dagood: I didn't rename some existing args to keep the delta smaller--would make sense to me to do renames for convention in a separate PR, depending on how much people mind merge conflicts.

targets/newtonsoft-json.props
jaredpar: Why not default to true, when unset, and then force all of the uncompliant projects to override to false? That is the same amount of churn as this change. It also has the benefit that the code reads much nicer and the final state is a lot cleaner

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/WriteBuildOutputProps.cs
eerhardt: I fixed this for all the task C# files, and removed the ones that don't appear to be used anymore.

targets/newtonsoft-json.props
eerhardt: Since this is just a temporary parameter (the end goal is that this property and the others go away when all repos are compliant), I don't think it matters too much either way.

@jaredpar is right that the code does read nicer when `Condition="'$(RepoApiImplemented)' == 'true'"` and `Condition="'$(RepoApiImplemented)' != 'true'"`.  I'll flip it to that for now.

dir.props
eerhardt: Agreed - I think we can do this on a separate PR to minimize churn for now.

dir.props
eerhardt: I actually don't see this property getting used....  I'll remove it if that's the case.

eerhardt:
@jaredpar and @weshaggard - I think all the feedback has been addressed.  Let me know if this can be merged.

targets/repository.proj
```diff
@@ -64,19 +98,42 @@
                             SourcePath="$(SourceBuiltPackagesPath)" />
```
weshaggard: Shouldn't we disable the mutation of the nuget.config with this change?

eerhardt: We are.  The condition on this target is `Condition="'$(NuGetConfigFile)' != ''"`.  And we are not setting that property when the repo supports the SourceOverrideRepoApi.

https://github.com/dotnet/source-build/pull/219/files#diff-fcefba7596e1fe07d9907df477c1384fR31

```xml
<PropertyGroup Condition="'$(SourceOverrideRepoApiImplemented)' != 'true'">
      <NuGetConfigFile Condition="'$(NuGetConfigFile)' == '' and Exists('$(ProjectDirectory)NuGet.config')">$(ProjectDirectory)NuGet.config</NuGetConfigFile>
```

weshaggard: cool

Issue #194, opened 2017-09-29
Determine Repo API framework recommendation for -p:Foo=Bar PowerShell passthrough


dagood:
A few people have encountered problems with the `-p:Foo=Bar` [Repo API format](https://github.com/dotnet/source-build/blob/dev/release/2.0/Documentation/RepoApi.md#parameters) when it comes to PowerShell:

@weshaggard https://github.com/dotnet/source-build/pull/152#discussion_r140380708
@natemcmaster https://github.com/dotnet/source-build/pull/181#issuecomment-333181763

This is because PowerShell parses `-p:Foo=Bar` as passing a parameter and splits it out into two arguments `-p:` and `Foo=Bar` when you use `$args` or `ValueFromRemainingArguments`. A caller-side workaround is to pass `'-p:Foo=Bar'` but that might be unreasonable.

We should have a recommendation for this, since it seems like it will be a common issue.

---

Alternatively:
 * Switch to `/p:Foo=Bar` which doesn't have the issue mentioned above (but introduces others)
 * Stop trying to use an MSBuild-compatible arg format and go with
   * PS arg style `-Foo bar`
   * Unixy `--foo=bar` or `--foo bar` (or `-foo bar`, although rarer (X11))
   * Something else that also works in more places than `-p:Foo=bar` without workarounds
 * Always use a response file
 * ...?

natemcmaster:
Another option to consider is passing in the [magic arg `--%`](https://ss64.com/ps/stop-parsing.html) which disables PowerShell's parameter binding for anything that follows.

dagood:
That's neat, didn't know about that arg. Just to make sure, you're saying it would make sense to you to recommend putting `--%` in a `build.cmd` if it's passing through to PowerShell?

(Also, I missed the parameter name collision issue you also mentioned in your PR comment, but it sounds like `--%` solves that too.)

natemcmaster:
The special arg helps, but it means you have to mirror powershell's parameter parsing if you want a powershell script parameters in addition `$args` or `ValueFromRemainingArguments`

For example, if we wanted a custom arg in powershell `-MyArg`. You would need something like this. 
```batch
REM build.cmd
@ECHO OFF
set LF=^

set PowerShellArgs=
set OtherArgs=
:loop
IF NOT "%1"=="" (
    IF "%1"=="-MyArg" (
        SET PowerShellArgs=%PowerShellArgs% -MyArg %2
        SHIFT
    ) ELSE (
[...]

PR #228, opened 2017-10-12
Update Repo API parameter passing.

eerhardt:
Addresses the concerns in #194 by following the recommended approach in that issue.

This approach isn't 100% POSIX compliant, but it close enough to work correctly on UNIX systems and works for Windows as well.

/cc @dotnet/source-build-contrib

### DIFF
--- a/Documentation/RepoApi.md
+++ b/Documentation/RepoApi.md
@@ -30,6 +30,8 @@ There will be a number of actions source-build will need the repos to implement.
 
 Actions will take the form of `ActionName` and are the first parameter passed to the build script. Action names are case insensitive.
 
+Note that actions don't have a leading dash `-`. Also note that individual repos can support a default action if one isn't specified. However, source-build will always specify an action explicitly.
+
 ### Parameters
 
 Each action above will take a set of parameters used to pass information into the action and control its behavior.
--- a/Documentation/RepoApi.md
+++ b/Documentation/RepoApi.md
@@ -28,17 +28,15 @@ There will be a number of actions source-build will need the repos to implement.
 
 1. `Build` - Assemble the product from the current source contained in the repo.
 
-source-build formats arguments using a style that is compatible with MSBuild. The reasoning for this is that the majority of repos have an MSBuild based build system. Using this format will allow for arguments to easily be passed through to MSBuild. Similarly, both action and parameter arguments are case-insensitive, just like MSBuild default behavior.
-
-Actions will take the form of `-t:ActionName`, just like MSBuild targets are passed to `msbuild.exe`.
+Actions will take the form of `ActionName` and are the first parameter passed to the build script. Action names are case insensitive.
 
 ### Parameters
 
 Each action above will take a set of parameters used to pass information into the action and control its behavior.
 
-Parameters will take the form of `-p:ParameterName=ParameterValue`, just like MSBuild properties are passed to `msbuild.exe`. One slight variation here is the usage of `-p` instead of `/p`.  MSBuild supports both forms, but source-build chooses `-p` because it meets [POSIX recommendations](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) for command line parameters and avoids any confusion with file paths on Unix machines.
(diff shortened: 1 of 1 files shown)


Documentation/RepoApi.md
```diff
@@ -28,25 +28,23 @@ There will be a number of actions source-build will need the repos to implement.
...
+In order to avoid conflicts between existing properties (MSBuild or otherwise) used by the repos and the properties passed in by source-build, source-build will append a prefix to every property name it passes in: `DotNet`.
 
 #### Detecting build from source
 
 Some repos build more projects than are necessary to build the .NET Core stack. For example, the `dotnet/roslyn` repo builds the C# compiler, which is needed, but it also builds Visual Studio extensions, which aren't needed. These VS extensions cannot be built without using an external dependency on tooling that isn't available when building from source.
 
 In order to tell repos that these projects shouldn't be built, source-build will pass the following parameter
 
-`-p:DotNetBuildFromSource=true`
+`-DotNetBuildFromSource true`
```
natemcmaster: Could this just be a switch without needing to also pass in `true`?

eerhardt: See the above line

> Every parameter passed by source-build will have a value, even if that value is just `true`.

My thinking is that will make the parsing easier - if there is a parameter name, then there is a parameter value.

The way POSIX solves this is to use `=` in between the parameter name and value.  If there is no `=`, then it is a switch.  

Also, when passing through to MSBuild, the parsing code won't need to figure out "which ones are switches?" and use `true` for their values going into MSBuild parameters.

This was my reasoning at least.  Thoughts?

weshaggard: >The way POSIX solves this is to use = in between the parameter name and value. If there is no =, then it is a switch.

Is there any harm with using that approach here? Or is that still and issue with PS parsing? I'm sort of hoping the PS issue is only because of the `/p` but perhaps not. 

Even if we don't have switches as part of the repo APIs I do think the arg parsing is likely going to need to handle them because a lot of repo's already support them, so I think the arg parsing complexity is already going to exist.

dagood: The `=` itself causes issues in the cmd file before passing through to the ps1:
https://github.com/dotnet/source-build/issues/194#issuecomment-333579905
(Complicated workaround:) https://github.com/dotnet/source-build/issues/194#issuecomment-333747630

natemcmaster: > My thinking is that will make the parsing easier - if there is a parameter name, then there is a parameter value.

I don't see any reason why that would be a problem. I was just wondering

Documentation/RepoApi.md
```diff
@@ -28,25 +28,23 @@ There will be a number of actions source-build will need the repos to implement.
 
 1. `Build` - Assemble the product from the current source contained in the repo.
 
-source-build formats arguments using a style that is compatible with MSBuild. The reasoning for this is that the majority of repos have an MSBuild based build system. Using this format will allow for arguments to easily be passed through to MSBuild. Similarly, both action and parameter arguments are case-insensitive, just like MSBuild default behavior.
-
-Actions will take the form of `-t:ActionName`, just like MSBuild targets are passed to `msbuild.exe`.
+Actions will take the form of `ActionName` and are the first parameter passed to the build script. Action names are case insensitive.
```
dagood: I think we should add that an actions won't start with `-` so that repos can detect that and have a default behavior for their "clone and type `build`" dev scenario.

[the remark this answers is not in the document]
eerhardt: I'm not sure.  I guess I would assume if we needed sub-actions, we would just create a new top-level action instead.

ex. if I needed to say `build.cmd Build Managed`, I would instead just create a `Build-Managed` action.

Or am I misunderstanding the question?  Did you mean would we need to specify multiple actions in one call?  ex `build.cmd Build;Package;Publish`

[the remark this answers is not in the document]
weshaggard: I was meaning your Build-Managed scenario. I guess a new action should be fine.

PR #233, opened 2017-10-12
Restructuring to use project files

weshaggard:
Instead of having a props/targets file for each repo now we have
a proj file for each repo to contain all the individual information
needed to handle that repo. This will help with diagnosing issues
as well letting msbuild handle the repo dependencies as project references.

cc @eerhardt @jaredpar 

@crummel @karajas can you guys also have a look to see if any of the tarball stuff will be broken by this refactoring.


build.ps1
```diff
@@ -36,4 +36,4 @@ $SDKPATH = "$CLIPATH\sdk\$SdkVersion"
 
 $captured_args = $args
 
-Exec-Block { & "$CLIPATH\dotnet" "$SDKPATH/MSBuild.dll" "$SCRIPT_ROOT/build.proj" /flp:v=detailed /clp:v=detailed $captured_args } | Out-Host
+Exec-Block { & "$CLIPATH\dotnet" "$SDKPATH/MSBuild.dll" "$SCRIPT_ROOT/build.proj" /flp:v=diag /clp:v=m $captured_args } | Out-Host
```
eerhardt: `diag` is going to get really, really big, isn't it?

eerhardt: Maybe instead we could pass in `/bl` for binlog?  http://msbuildlog.com/

build.proj
```diff
@@ -5,17 +5,16 @@
...
-      <_RootProject Include="$(ProjectDir)targets/repository.proj" Condition="'%(Repository.Identity)' == '$(DefaultRepository)'">
-        <AdditionalProperties>RepositoryName=%(Repository.Identity);PathToRepo=%(Repository.PathToRepo);RepositoryOrganization=%(Repository.Organization);RepositoryBranch=%(Repository.Branch)</AdditionalProperties>
-      </_RootProject>
-    </ItemGroup>
+    <MSBuild Projects="repos\$(RootRepo).proj" Targets="Build" BuildInParallel="$(BuildInParallel)" />
+  </Target>
 
-    <MSBuild Projects="@(_RootProject)" Targets="BuildRepositoryAndDependencies" BuildInParallel="$(BuildInParallel)" />
+  <Target Name="ApplyPatches">
+    <MSBuild Projects="repos\$(RootRepo).proj" Targets="ApplyPatches" BuildInParallel="$(BuildInParallel)" />
```
eerhardt: Doesn't this just apply patches to the one repo? 

Do we really need this target?

build.ps1
weshaggard: it will only be diag for the source-build itself not the other repos.

build.proj
weshaggard: we could remove it but I found it helpful when just working on one repo.

build.ps1
weshaggard: I don't want to force us to use a custom tool to read the logs, especially on non-windows.

repos/dir.props
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" InitialTargets="SetNuGetPackagesEnvironment" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., dir.props))/dir.props" />
+
+  <PropertyGroup>
+    <RepositoryName Condition="'$(RepositoryName)' == ''">$(MSBuildProjectName)</RepositoryName>
+    <ProjectDirectory Condition="'$(ProjectDirectory)' == ''">$(SubmoduleDirectory)$(RepositoryName)/</ProjectDirectory>
+    <RepositoryOrganization Condition="'$(RepostioryOrganization)' == ''">dotnet</RepositoryOrganization>
+    <RepositoryBranch Condition="'$(RepositoryBranch)' == ''">release/2.0.0</RepositoryBranch>
+    <MinimalConsoleLogOutput>true</MinimalConsoleLogOutput>
```
weshaggard: While I think this should be true local dev work as it greatly helps with the console reabability. Should we have it set to false for CI and official builds to push more of the information to build console?

repos/dir.props
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" InitialTargets="SetNuGetPackagesEnvironment" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory).., dir.props))/dir.props" />
+
+  <PropertyGroup>
+    <RepositoryName Condition="'$(RepositoryName)' == ''">$(MSBuildProjectName)</RepositoryName>
+    <ProjectDirectory Condition="'$(ProjectDirectory)' == ''">$(SubmoduleDirectory)$(RepositoryName)/</ProjectDirectory>
```
eerhardt: Is it possible to come up with a better name for this property?  `ProjectDirectory` is rather ambiguous to me.

weshaggard: Sure but that should be another change. Perhaps RepoSourceDirectory or something along those lines?

build.ps1
eerhardt: Using minimum for the console output doesn't work well in CI, we currently don't capture the log files at all.

If we are going to change the console output, can you also ensure we can get access to the log files in CI?

I think this is how you do it:  https://github.com/dotnet/core-setup/blob/2e5ec4b9ac8084c61fc4b8130c33f08f783aaae3/netci.groovy#L104-L108

eerhardt: > I don't want to force us to use a custom tool to read the logs, especially on non-windows.

Can we do both?  I don't know how much you've tried msbuildlog, but it is rather fantastic.

weshaggard: That is related to my other question. I would prefer for CI we just dump things to the console so there is no need for finding logs and downloading them. So perhaps we should have pass the extra detailed switch in CI to make it have higher logging.

weshaggard: >Can we do both? I don't know how much you've tried msbuildlog, but it is rather fantastic.

We could do both on windows I suppose. I tried it a number of times for corefx but the logs are so huge it cannot handle reading those diag logs. I personally find I always need diag logs to get property values when diagnosing most issues.

eerhardt: It also looks like msbuild supports "replaying" the binary log to a diag log file:  https://github.com/Microsoft/MSBuild/wiki/Binary-Log#replaying-a-binary-log.  Which means we could do binlog, and then if someone needed to view to the diag file log, they could just replay it into a text file.

eerhardt: > I would prefer for CI we just dump things to the console so there is no need for finding logs and downloading them.

For large logs, my browser can't display and search very well.  So I normally just download the console output anyway and then use a text editor to read/search it.

eerhardt: Looking at the logs from your run, it doesn't appear to get getting any logs from the sub-repos:

```
14:48:18   [21:48:18.56] Building 'common'
14:48:18   Running command:
14:48:18      
14:48:18     Log: /mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/bin/logs/common.log
14:48:18     With Enivornment Varibles:
14:48:18       DOTNET_RUNTIME_ID=rhel.7-x64
14:48:18       DOTNET_TOOL_DIR=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/Tools/dotnetcli/
14:48:18       BUILD_TOOLS_TOOL_DIR=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/Tools/
14:48:18       BUILDTOOLS_SKIP_CROSSGEN=1
14:48:18       DotNetBuildFromSource=true
14:48:18       DotNetRestorePackagesPath=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/packages/
14:48:18       DotNetCoreSdkDir=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/Tools/dotnetcli/
14:48:18       DotNetBuildToolsDir=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/Tools/
14:48:24   [21:48:24.26] Building 'common'...done
14:48:24   [21:48:24.27] Building 'nuget-client'
14:48:24   Running command:
14:48:24      
14:48:24     Log: /mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/bin/logs/nuget-client.log
14:48:24     With Enivornment Varibles:
14:48:24       DOTNET_RUNTIME_ID=rhel.7-x64
14:48:24       DOTNET_TOOL_DIR=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/Tools/dotnetcli/
14:48:24       BUILD_TOOLS_TOOL_DIR=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/Tools/
14:48:24       BUILDTOOLS_SKIP_CROSSGEN=1
14:48:24       DotNetBuildFromSource=true
14:48:24       DotNetRestorePackagesPath=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/packages/
14:48:24       DotNetCoreSdkDir=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/Tools/dotnetcli/
14:48:24       DotNetBuildToolsDir=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/Tools/
14:48:24       MS_PFX_PATH=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/keys/NuGet.Client.snk
14:49:15   [21:49:15.15] Building 'nuget-client'...done
```

You should try to make one fail to see what the debugging experience looks like.

eerhardt: It also really stinks when you look at the console output to see how far along the build is, and it just sits there for 9 minutes doing nothing:

```
14:49:15       DotNetBuildToolsDir=/mnt/resource/j/workspace/dotnet_source-build/dev_release_2.0/RHEL7.2_Release_prtest/Tools/
14:58:20   [21:58:20.20] Building 'coreclr'...done
```

repos/application-insights.proj
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+  <PropertyGroup>
+    <DotNetCommand>$(DotNetCliToolDir)dotnet</DotNetCommand>
```
eerhardt: This property is duplicated on a lot of these files.  Can it be refactored up to the dir.props?

build.ps1
weshaggard: As we chatted I fixed this in CI to have more logging but for local dev scenarios I like the cleaner logs but folks can enable more logging if they like.

weshaggard:
@dotnet test ci

weshaggard:
@dotnet test ci

weshaggard:
@dotnet test ci

build.proj
eerhardt: I think that it's confusing if I say

`build /t:ApplyPatches`

that it will only apply patches to the CLI repo.

build.proj
```diff
@@ -5,17 +5,16 @@
...
-    <ItemGroup>
-      <_RootProject Include="$(ProjectDir)targets/repository.proj" Condition="'%(Repository.Identity)' == '$(DefaultRepository)'">
-        <AdditionalProperties>RepositoryName=%(Repository.Identity);PathToRepo=%(Repository.PathToRepo);RepositoryOrganization=%(Repository.Organization);RepositoryBranch=%(Repository.Branch)</AdditionalProperties>
-      </_RootProject>
-    </ItemGroup>
+    <MSBuild Projects="repos\$(RootRepo).proj" Targets="Build" BuildInParallel="$(BuildInParallel)" />
+  </Target>
 
-    <MSBuild Projects="@(_RootProject)" Targets="BuildRepositoryAndDependencies" BuildInParallel="$(BuildInParallel)" />
+  <Target Name="ApplyPatches">
```
jaredpar: Shouldn this target condition on `'$(RootRepo)' != ''`?

repos/roslyn.proj
```diff
@@ -1,5 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
```
jaredpar: Seems like a lot of ceremony here. Should be able to simplify to 

``` 
<Import Project="dirs.props" />
```

weshaggard: Yes but it is copy and past-able in all projects and originally I didn't have a dir.props file. It is a nice convention that we use everywhere.

build.proj
weshaggard: OK. Would you suggest I remove it completely or make it apply all the patches?

build.proj
weshaggard: RootRepo is always defaulted so someone would need to explicitly pass empty for it to be empty and if they did they would get an error saying cannot find project. If I were to add a condition I would also want to add an error saying the project doesn't exist and that is basically what the msbuild task does for me if it doesn't exist.

build.proj
eerhardt: Either of those 2 things would work for me.

I'd lean towards removing it.  The init-build.proj will already apply all patches the first time a build is executed.

repos/roslyn.proj
jaredpar: Both are copy pastable. There is no semantic difference between the two.

weshaggard: except one requires a dir.props in every directory and the other doesn't as it will find the next one in the directory structure, so if we move this file around or move them into there own subdirectories or things like that we don't need to change the files at all, which is why I prefer the approach I used.

PR #237, opened 2017-10-13
Remove root ApplyPatches and clean-up dotnetcommmand properties

weshaggard:
cc @eerhardt


eerhardt:
Looks like source-build doesn't import `Build.Common.props`, which means this property isn't getting set.

Issue #239, opened 2017-10-16
CLI is always reinstalled by init-tools.sh


dagood:
This is making the CLI always install on my Fedora 26 runs (and probably other shells):

https://github.com/dotnet/source-build/blob/51f2c9ae3aabf8168df046fa0474be3dd2997822/init-tools.sh#L17-L20

> /home/dagood/git/here-source-build/init-tools.sh: line 19: return: can only `return' from a function or sourced script

Two fixes come to mind:
 1. Wrap the rest of the file with `else` rather than trying to early return.
 2. Put the code in a function and keep using `return`.

@weshaggard @eerhardt

weshaggard:
Most repo's are sourcing in the init-tools script which is why we needed it to be a return instead of exit. That said I was under the impression that return should turn into an exit if it is an outer script, perhaps my understand is incorrect on that front.

I will do some testing but if this is blocking you the easiest fix is to source in the script from build.sh.

dagood:
Looking at the PR that changed from executing to sourcing (https://github.com/dotnet/corefx/pull/19363), switching to sourcing is actually better anyway. The only remaining problem is anyone calling `./init-tools.sh`, but I don't have a reason to support that.

PR #240, opened 2017-10-16
Source (.) init-tools.sh rather than executing it

dagood:
Fixes https://github.com/dotnet/source-build/issues/239


build.sh
```diff
@@ -15,7 +15,11 @@ export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
 export DOTNET_MULTILEVEL_LOOKUP=0
 export NUGET_PACKAGES="$SCRIPT_ROOT/packages/"
 
-"$SCRIPT_ROOT/init-tools.sh"
+(
+    # init-tools.sh isn't compatible with these flags. Unset them in this subshell.
+    set +euo pipefail
```
dagood: CI failed and I had to add this workaround, still look ok @weshaggard?

weshaggard: Can we fix init-tools so it works with this option? What was the failure?

dagood: The first error is:

```
init-tools.sh: line 9: __BUILDTOOLS_SOURCE: unbound variable
```

I'll give fixing it a try (waiting for a different build to finish).

weshaggard: I wonder if we should just remove the "u" option in the root script.

build.sh
```diff
@@ -15,7 +15,7 @@ export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
 export DOTNET_MULTILEVEL_LOOKUP=0
 export NUGET_PACKAGES="$SCRIPT_ROOT/packages/"
 
-"$SCRIPT_ROOT/init-tools.sh"
+. "$SCRIPT_ROOT/init-tools.sh"
```
eerhardt: Does it work to use the actual `source "$DIR/common/_prettyprint.sh"` command?  That seems more obvious to me when reading the code than trying to see that little `.`.

dagood: `source` isn't part of POSIX, so `.` is in theory more portable.

eerhardt: We have #!/bin/bash at the top...

dagood: Hmm, more real than I thought: `dash` apparently doesn't support `source`. https://stackoverflow.com/a/13702876 https://wiki.ubuntu.com/DashAsBinSh#source

eerhardt: 👍

dagood: (Whoops, didn't refresh and missed your comment.)

You're right that we do have that direct dependency on `bash` right now, and it would be ok to use `source`. A while back someone pointed out some bashisms in BuildTools and I think CoreFX. I've pointed out obvious ones like `[[` since then, but I don't know if there's a higher-level drive to avoid a `bash` dependency.

weshaggard: Right now we are definitely dependent on bash.

dagood: I'll go ahead and change it to `source` then--the readability concern also came up in the CoreFX PR. It's not like this would be hard to spot/fix if we do want to work on non-`bash`.

build.sh
dagood: Personally, I really like how `-u` forces scripts to be clear if they expect the variable to be set at that point in time, and I think it's worth keeping. I haven't done a ton of work with it yet, though.

PR #249, opened 2017-10-18
Update BuildTools and core-setup to the latest.

### LINKED ISSUE #201: Implement auto-dependency flow Core-Setup

eerhardt:
Fix #201


dagood:
Can you update anything in `core-setup.proj` yet? I'd only say #201 is done once the Repo API implementation exceptions are gone.

eerhardt:
Good point.  Updated.

I guess the only "TODO" part that is left is that the `PublishCoreSetupBinaries` task is still running.  This is necessary because the CLI doesn't know to pick these binaries up from the transport feed's `assets` directory yet.

I'll log a separate issue for solving that later, when the CLI can uptake it.

eerhardt:
Thinking more, I should update Core-setup to start writing to the transport feed ‘assets’ directory now. That way CLI can use the assets when it gets updated.

repos/core-setup.proj
```diff
@@ -6,13 +6,12 @@
...
-    <PackagesOutput>$(ProjectDirectory)/Bin/$(TargetRid).$(Configuration)/packages/</PackagesOutput>
-    <BinariesOutput>$(PackagesOutput)</BinariesOutput>
     <LatestCommit>82b22fa44e46d624ebeddf156f33e9cfb317bd57</LatestCommit>
     <VersionSeedDate>2017-07-20</VersionSeedDate>
     <OfficialBuildId>20170720-05</OfficialBuildId>
-    <DependencyVersionInputRepoApiImplemented>false</DependencyVersionInputRepoApiImplemented>
-    <OutputPlacementRepoApiImplemented>false</OutputPlacementRepoApiImplemented>
+
+	<!-- Need to set $(PackagesOutput) so WriteVersions writes the versions file for cli, until cli respects auto-dependency flow -->
+    <PackagesOutput>$(SourceBuiltPackagesPath)</PackagesOutput>
```
dagood: Tab char snuck in before the comment.

repos/core-setup.proj
```diff
@@ -26,28 +25,14 @@
...
-      <UpdateStep Remove="@(UpdateStep)" />
-      <UpdateStep Include="@(FilteredUpdateStep)" />
-    </ItemGroup>
-  </Target>
-
+  <!-- Need to "publish" the tarballs for cli, until cli respects the assets blob -->
   <Target Name="PublishBinaries" AfterTargets="Package">
     <ItemGroup>
-      <CoreSetupTarballs Include="$(BinariesOutput)*$(TarBallExtension)" />
+      <CoreSetupTarballs Include="$(SourceBuiltAssetsDir)dotnet-runtime-*$(TarBallExtension)" />
```
dagood: I didn't realize it Core-Setup-side--this path means https://github.com/dotnet/source-build/blob/2c79047d76f6a1feb4cc70077102d3409cea9634/Documentation/auto-dependency-flow/api.md#-pdotnetassetrooturlurl-not-yet-implemented should be updated. Currently the recommended implementation is specifically for CLI and it says the installers will be in a `Runtime/{version}/` dir within the blob feed.

(It isn't important to me which way we go--tried both in the PRs for that doc and ended up where we are now, but all that matters is being consistent.)

eerhardt: This also impacts how we are going to do publishing in orchestrated builds.  I'm not sure I know which way is the "right" way here.

eerhardt: One thought is, if we have a single "assets root url" for a build, it doesn't really make sense to put things in a subfolder under the root it with a version in the folder.  The root url would already be unique for the given build, right?

(I know I may have said the opposite previously, but I don't think I was realizing all the intricacies here.)

dagood: The root url is unique for each product build, but we *could* find ourselves building multiple versions of Core-Setup within that product build at some point. As long as Core-Setup maps each file directly to a target publish location, we're fine, but I don't know if there's any globbing we expect to be able to do.

My gut agrees that we shouldn't need this "extra" `Runtime/{version}/` dir because the version is already in the filename at this point. (Only removed once we publish to latest on `dotnetcli` storage.)

repos/core-setup.proj
eerhardt: I fixed it, since I was updating doc anyway.

repos/core-setup.proj
eerhardt: I updated the doc to the current approach.  We can make adjustments as we go, if we decide this is a bad approach.

Documentation/auto-dependency-flow/api.md
```diff
@@ -141,22 +141,20 @@ Use the AssetRoot properties if they are defined. Example (adapted from [CLI's b
...
-    <CoreSetupRootUrl>$(CoreSetupBlobRootUrl)Runtime/</CoreSetupRootUrl>
+    <CoreSetupBlobRootUrl Condition="'$(CoreSetupBlobRootUrl)' == ''">https://dotnetcli.azureedge.net/dotnet/Runtime/$(SharedFrameworkVersion)/</CoreSetupBlobRootUrl>
 
     <CoreSetupBlobAccessTokenParam Condition="'$(DotNetAssetRootAccessTokenSuffix)' != ''">$(DotNetAssetRootAccessTokenSuffix)</CoreSetupBlobAccessTokenParam>
   </PropertyGroup>
 
   <ItemGroup>
     <_DownloadAndExtractItem Include="CombinedSharedHostAndFrameworkArchive">
-      <Url>$(CoreSetupRootUrl)$(SharedFrameworkVersion)/$(CombinedFrameworkHostCompressedFileName)$(CoreSetupBlobAccessTokenParam)</Url>
+      <Url>$(CoreSetupRootUrl)$(CombinedFrameworkHostCompressedFileName)$(CoreSetupBlobAccessTokenParam)</Url>
```
dagood: Should `CoreSetupRootUrl` be `CoreSetupBlobRootUrl` here? (Or rename `CoreSetupBlobRootUrl`.)

eerhardt: Yes.

Documentation/auto-dependency-flow/api.md
```diff
@@ -141,22 +141,20 @@ Use the AssetRoot properties if they are defined. Example (adapted from [CLI's b
 
   <PropertyGroup>
     <CoreSetupBlobRootUrl Condition="'$(DotNetAssetRootUrl)' != ''">$(DotNetAssetRootUrl)</CoreSetupBlobRootUrl>
-    <CoreSetupBlobRootUrl Condition="'$(CoreSetupBlobRootUrl)' == ''">https://dotnetcli.azureedge.net/dotnet/</CoreSetupBlobRootUrl>
-
-    <CoreSetupRootUrl>$(CoreSetupBlobRootUrl)Runtime/</CoreSetupRootUrl>
+    <CoreSetupBlobRootUrl Condition="'$(CoreSetupBlobRootUrl)' == ''">https://dotnetcli.azureedge.net/dotnet/Runtime/$(SharedFrameworkVersion)/</CoreSetupBlobRootUrl>
```
weshaggard: Why are we hard-coding this here? I would expect it to be in the build definition.

eerhardt: a) that's how the CLI works today.  https://github.com/dotnet/cli/blob/86732427fa598a72381a103dda83585e9938867b/build/BundledRuntimes.props#L27
b) This is just an example in docs.  The real code can do whatever it wants.

weshaggard: lol - I missed this is just a doc but still my question holds that I would like to avoid hard-coding these paths here as I don't like the duplication of this data in the build definitions and in code. We should just pass it from the build definition, particularly when we need to override it for our servicing builds.

eerhardt: IMO it is easier to understand what we are trying to illustrate here - how to use `$(DotNetAssetRootUrl)` - if we limit the number of concepts we are using. Trying to describe some other external "build definition parameter", which serves no purpose here just clutters the example.

eerhardt: Also, this URL is necessary for the dev machine scenario when the developer just types "build" and doesn't supply the URL.

weshaggard: I'm not sure if you agree or disagree with my suggestion :)

At any rate my primary concern is that we avoid duplication of this information in various places and I don't think it is actually needed for local dev builds so I would prefer they stay as a build definition parameter.

dagood: CLI's in a state where I think it's actually more maintainable if the default stays in the source code: CLI has no checked-in build defs and no central build definition to set the variable in. Moving to the build definitions would actually duplicate this URL across each platform leg.

eerhardt: I don't think you are understanding what is happening here.

The CLI needs to get the Shared Framework from Core-Setup during its build.  When a developer just has the CLI repo and types "build", the CLI build needs to acquire the Shared Framework from somewhere.

> I don't think it is actually needed for local dev builds

This is where you are misunderstanding.  It is 100% needed.  You can't build the CLI repo without a runtime.

eerhardt: Think of this like myget URLs.  We need to get the NuGet packages from somewhere.  You can't build without them.  By default, we hard-code the URLs to the feeds we are acquiring the NuGet packages from.  This is the exact same thing.

weshaggard: Yes I am miss-understanding :) I was thinking that this is where we are publishing stuff from core-setup, however now if I understand this is essentially like a checked in feed for folks like CLI to be able to pull the shared framework. I think that is OK. 

Hangs head in sham and will go read the docs closer :(

Documentation/auto-dependency-flow/api.md
eerhardt: fixed.

Issue #151, opened 2017-09-20
Make moving to new versions of submodules easy


crummel:
Updating to the latest version of submodules should be one command.

dagood:
No longer blocked: https://github.com/dotnet/corefx/pull/25208 upgraded BuildTools over the hump.

dagood:
Implemented in https://github.com/dotnet/source-build/pull/254. Now I can use it to hook up auto-submodule-update PRs (https://github.com/dotnet/source-build/issues/269).

Issue #107, opened 2017-08-22
Use NuGet's fork of NuGet.Client in source-build submodule


dagood:
We had to switch to my fork in https://github.com/dotnet/source-build/pull/94. Should be able to switch back if we move to a newer NuGet.Client commit that doesn't have the FileSystem submodule problem. (A branch disappeared that I had to revive in my own FileSystem fork.)

dagood:
I see--to connect things together, the NuGet.Client commit brought in by https://github.com/dotnet/source-build/pull/123 got us to a cloneable commit.

PR #254, opened 2017-10-21
Use new VersionTools submodule updaters

dagood:
Add submodule dependency tooling: part 1 of https://github.com/dotnet/source-build/issues/151. For part 2 (submodule update PRs), I'll be setting up a scheduled build in VSTS.

Configuration for submodules goes in `.gitmodules`. The commands here will only affect submodules with a `branch`. Auto-updates can be disabled without removing the `branch` by adding `sourceBuildAutoUpdate = false` to the submodule config.

The reason for using `.gitmodules` is that it lets devs use `git submodule update --remote` if they want to.

The new upgrade-related targets (not all necessarily useful, just stuff we get for free with VersionTools):

 * **`/t:PullSubmodules`** or `/t:UpdateToRemoteDependencies`
   * Check out the latest remote commit of each submodule.
 * `/t:UpdateDependencies`
   * Update each submodule to the hash in the index. (Same result as running `git submodule update`.)
 * `/t:VerifyDependencies`
   * Check if the commit of any checked-out submodule doesn't match the hash in the Git index of source-build.
 * **`/t:UpdateToRemoteDependencies /t:CreatePullRequestIfChanged`** or `/t:UpdateDependenciesAndSubmitPullRequest`
   * Updates to the *latest remote commit* of each configured submodule, then submits a PR.

All targets support `/p:TargetSubmodule=src/corefx` to specify a single submodule.

I'm able to run all the targets on Windows and Fedora 26.

@dotnet/source-build-contrib


.gitmodules
```diff
@@ -1,18 +1,21 @@
...
 	path = src/standard
 	url = https://github.com/dotnet/standard
+	branch = dev/release/2.0.0
 [submodule "src/cli"]
 	path = src/cli
-	url = https://www.github.com/dotnet/cli
+	url = https://github.com/dotnet/cli
 [submodule "src/nuget-client"]
 	path = src/nuget-client
 	url = https://github.com/dagood/NuGet.Client
```
eerhardt: Are we OK with this pointing to your fork?

.gitmodules
```diff
@@ -63,5 +67,6 @@
 	url = https://github.com/aspnet/common
 [submodule "src/sdk"]
 	path = src/sdk
-	url = https://github.com/dotnet/sdk.git
+	url = https://github.com/dotnet/sdk
 	branch = release/2.0.0
+	sourceBuildAutoUpdate = false
```
eerhardt: Is this added just for illustration purposes?  Or is there a reason `sdk` shouldn't be auto updated, but everything else should be.

dir.props
```diff
@@ -36,6 +36,7 @@
     <BaseOutputPath>$(ProjectDir)bin/</BaseOutputPath>
     <ToolsDir>$(ProjectDir)Tools/</ToolsDir>
     <TaskDirectory>$(ProjectDir)tools-local/tasks/</TaskDirectory>
+    <TargetDirectory>$(ProjectDir)tools-local/targets/</TargetDirectory>
```
eerhardt: Is there an advantage to having this new, separate directory for just one file?  Could we put the `update-submodules.targets` file in the root?

.gitmodules
dagood: No, this is tracked by https://github.com/dotnet/source-build/issues/107. Not trivial to switch to a new fixed commit: see https://github.com/dotnet/source-build/pull/231 for an attempt.

.gitmodules
dagood: There are currently a lot of `sdk` commits on that branch that aren't in source-build, and I didn't want to include them as part of this change (and the coming auto-updates). After this PR, another PR can be submitted to bring up `sdk` and enable auto-updates for it.

dir.props
dagood: That would be fine. I do consider it a local tool, though, and I figured the fewer things in the root, the better (usually).

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/ReadGitConfigFile.cs
```diff
@@ -0,0 +1,143 @@
...
+                .GroupBy(o => o.Subsection)
+                .Select(g =>
+                {
+                    string path = g.FirstOrDefault(o => o.Name == "path")?.Value;
+                    string url = g.FirstOrDefault(o => o.Name == "url")?.Value;
+
+                    if (string.IsNullOrEmpty(path) ||
+                        string.IsNullOrEmpty(url))
+                    {
+                        return null;
```
eerhardt: You may want/need to filter these `null` items out.  That way you aren't returning `null` items to MSBuild.  I've never tried this before.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/ReadGitConfigFile.cs
```diff
@@ -0,0 +1,143 @@
...
+                    }
+
+                    string branch = g.FirstOrDefault(o => o.Name == "branch")?.Value ?? string.Empty;
+
+                    // "git --list" converts entries to lowercase, but camelCase is standard
+                    // capitalization within the config file.
+                    string sourceBuildAutoUpdate = g
+                        .FirstOrDefault(o => o.Name == "sourceBuildAutoUpdate".ToLowerInvariant())
```
eerhardt: Use `OrdinalIgnoreCase` here instead.

.gitmodules
eerhardt: But all the other repos are fine?

eerhardt:
> A more practical reason is that I can't think of a clear way to "extract" a property from each submodule's .proj. (I can think of parsing or building each one with a target that writes the property to a file. Maybe I'm missing something.)

You can put a new target in the `dir.targets` file in the folder, and then invoke that target on each project.  We do this quite often for things like project-to-project references, or even that's how we do multiple `TargetFrameworks` support.

Here are a few examples:

How ProjectA => ProjectB knows the output file from ProjectB to reference:
https://github.com/Microsoft/msbuild/blob/57ae27cf0812bf262c6f8be456ed527a1542dba7/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1723-L1735

How a Project builds for multiple TargetFrameworks:
https://github.com/dotnet/sdk/blob/b5aa4a2b0db4f6203050063139e1873fa024337a/src/Tasks/Microsoft.NET.Build.Tasks/buildCrossTargeting/Microsoft.NET.Sdk.targets#L65-L69

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/ReadGitConfigFile.cs
dagood: Yeah, it was working, but no need to rely on it. Adding a clause.

.gitmodules
dagood: The only other ones that have `branch`es are `coreclr`, `corefx`, `core-setup`, and `standard`, which are all up to date right now except `corefx` (which Wes has been working on).

Do you think it would be better to make this an explicit enable per-repo? I suppose it's possible that someone might add a `branch` without intending it to be auto-upgraded.

dagood:
@eerhardt Ah, neat, I looked at `TargetOutputs`, but I'd always thought that a target's `Outputs` had to be an actual file. I just tried it out with arbitrary values and it works. I'll stick with `.gitmodules` for now for the other reason, but I'll keep this in mind for the future.

.gitignore
```diff
@@ -268,4 +268,11 @@ Session.vim
...
 **/IISApplications
 
+### source-build ###
 prebuilt/
+msbuild.binlog
+
+# Exists temporarily during build. Interferes with VersionTools'
+# UpdateDependenciesAndSubmitPullRequest: if no upgrades are performed, the
+# sanity check fails when "git status" shows this file.
+msbuild.ProjectImports.zip
```
weshaggard: Interesting I've never seen this file. Is there a way to get it redirected to bin/obj instead?

.gitmodules
weshaggard: Why not just have the repo's that are fixed and not currently updatable just point to a sha (assuming the branch property can point to a sha instead).

dir.props
weshaggard: I wouldn't put it in the root nor in its own directory I would just put it in the tools-local folder directly.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/ReadGitConfigFile.cs
```diff
@@ -0,0 +1,145 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Tasks;
+using Microsoft.Build.Utilities;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.DotNet.Build.Tasks
+{
+    public class ReadGitConfigFile : Task
```
weshaggard: Why wouldn't this be in the same task library as the VersionTools? It seems like we would need it for any other repo that does submodule auto-updates.

.gitignore
dagood: Looks like it's placed in the same folder as the `binlog` file, so I can change `/bl` to `/bl:bin/logs/msbuild.binlog` to make it show up in `"bin\logs\output.ProjectImports.zip`, or `/bl:bin/obj/msbuild.binlog`.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/ReadGitConfigFile.cs
dagood: In theory, the updaters can be configured from some other data source, but I agree--this way should be standard, and we should support it from BuildTools.

.gitmodules
dagood: That would lose info about what branch it is "sort of tracking" or "manually tracking", which seems useful to me to keep around so we don't start tracking the wrong branch at some point if we have to disable a branch and enable it again.

Implementation-wise, as-is, this is used with `git ls-remote refs/heads/{branch}`, and commits aren't heads on GitHub so VersionTools gets back a blank response. This takes extra time to look up, and in the end throws an exception. I would need to add a special case at some point in the chain to handle this being set to a commit. (Fortunately GitHub doesn't allow branches to be named in a way that looks like a commit.)

I don't know if anyone plans to use this, but that also wouldn't work with `git submodule update --remote`.

dir.props
dagood: Will do.

.gitmodules
weshaggard: Good points I'm fine with adding extra property assuming it doesn't mess up git up itself or git doesn't throw them out when this file gets updated for other means.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/ReadGitConfigFile.cs
weshaggard: Right now the implementation is also very specific to what you are after but the name is pretty general. It would probably be a good idea to make this a little more general purpose at some point.

tools-local/update-submodules.targets
```diff
@@ -0,0 +1,55 @@
...
+
+    <ItemGroup>
+      <!-- By default, upgrade each submodule with a branch configured in '.gitmodules'. -->
+      <SubmoduleConfiguration>
+        <SourceBuildAutoUpdate Condition="'%(SubmoduleConfiguration.SourceBuildAutoUpdate)' == ''">true</SourceBuildAutoUpdate>
+      </SubmoduleConfiguration>
+
+      <_UpgradableSubmodule Include="@(SubmoduleConfiguration)"
+                            Condition="'%(SubmoduleConfiguration.Branch)' != '' AND
+                                       '%(SubmoduleConfiguration.SourceBuildAutoUpdate)' == 'true'" />
```
weshaggard: Why not just change the condition to `'%(SubmoduleConfiguration.SourceBuildAutoUpdate)' != 'false'`?

build.ps1
```diff
@@ -34,6 +34,12 @@ Exec-Block { & "$SCRIPT_ROOT\init-tools.cmd" } | Out-Host
 $CLIPATH = "$SCRIPT_ROOT\Tools\dotnetcli"
 $SDKPATH = "$CLIPATH\sdk\$SdkVersion"
 
+$BINLOG_ROOT = "$SCRIPT_ROOT\bin\obj\"
```
weshaggard: I think this is probably worse then the original problem. I think I like doing the gitignore over having to initialize/create this directory directly in the scripts.

dagood: Alright, reverting. Didn't realize this would be needed until I saw it in CI.

tools-local/update-submodules.targets
dagood: Depends on what we want to be truthy... I don't have a strong opinion here. As long as nobody tries to use `0`, `no`, or `null` or something, `!= 'false'` is fine and removes a few lines of code. I'll go ahead with changing it for now.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/ReadGitConfigFile.cs
dagood: I started off with some more general output, but it was pretty unfriendly to try to get the info I wanted out of a more general config dump in MSBuild. I figured adding more outputs as they are needed might be a reasonable way to expand if necessary.

.gitmodules
dagood: Cool, yeah, existing config is preserved and doesn't interfere with anything.

I'm looking at bringing the target that creates submodule updaters into buildtools, and that makes me think the `sourceBuildAutoUpdate` config should be opt-in, not opt-out, to make the experience more intentional by default.

dagood:
Now that BuildTools is up to date, should be able to finally merge this.

In my last push I updated CoreCLR, CoreFX, and Core-Setup to point to their release/2.0.0 branches because dev/release/2.0.0 was merged in.

Issue #169, opened 2017-09-25
Remove patches from aspnet/common


eerhardt:
Remove all from https://github.com/dotnet/source-build/tree/dev/release/2.0/patches/common

natemcmaster:
You can assign this one to me. I'll be doing the dev work on our side to make this repo build without patches.

PR #261, opened 2017-10-25
Remove patches for aspnet/common 

natemcmaster:
Resolves https://github.com/dotnet/source-build/issues/169.

Depends on https://github.com/aspnet/Common/pull/275

The servicing/1.0.x branch of aspnet/Common implements the repo-API for producing Microsoft.Extensions.CommandLineUtils 1.0.1 (required by nuget-client)


natemcmaster:
The build fails on Windows.

> MSBUILD : error MSB1001: Unknown switch. [D:\j\workspace\Windows_NT_Re---a3f4c8f6\repos\common.proj]
13:01:29   D:\j\workspace\Windows_NT_Re---a3f4c8f6\repos\dir.targets(131,5): error MSB3073: The command "D:\j\workspace\Windows_NT_Re---a3f4c8f6\src/common//build.cmd -NoTest /p:Configuration=Release  /p:DotNetPackageVersionPropsPath=D:\j\workspace\Windows_NT_Re---a3f4c8f6\bin/obj/x64/Release/PackageVersions.props /p:DotNetRestoreSourcePropsPath=D:\j\workspace\Windows_NT_Re---a3f4c8f6\bin/obj/x64/Release/RestoreSources.props /p:DotNetOutputBlobFeedDir=D:\j\workspace\Windows_NT_Re---a3f4c8f6\bin/obj/x64/Release/blob-feed/ " exited with code 1. [D:\j\workspace\Windows_NT_Re---a3f4c8f6\repos\common.proj]

I can't figure out which switch is not recognized. I tried adding more logging to aspnetcore's build.ps1 script, but its output doesn't appear in the build log. I suspect it might be the `--force` flag on  `dotnet restore` which is new in 2.0.0.

.gitmodules
```diff
@@ -61,6 +61,7 @@
 [submodule "src/common"]
 	path = src/common
 	url = https://github.com/aspnet/common
+	branch = servicing/1.0.x
```
weshaggard: Why the servicing/1.0.x branch?

repos/common.proj
```diff
@@ -1,18 +1,16 @@
...
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+
   <PropertyGroup>
-    <PackagesOutput>$(ProjectDirectory)src/Microsoft.Extensions.CommandLineUtils/bin/$(Configuration)/</PackagesOutput>
-    <RepoApiImplemented>false</RepoApiImplemented>
+    <RepositoryOrganization>aspnet</RepositoryOrganization>
+    <RepositoryBranch>servicing/1.0.x</RepositoryBranch>
+    <PackagesOutput>$(ProjectDirectory)artifacts/build/</PackagesOutput>
+    <BuildArguments>/p:Configuration=$(Configuration)</BuildArguments>
+    <BuildArguments Condition=" '$(OS)' != 'Windows_NT' ">$(BuildArguments) /p:SkipNETFramework=true</BuildArguments>
```
weshaggard: You should consider setting SkipNETFramework to true in your repo based on DotNetBuildFromSource being set to true https://github.com/dotnet/source-build/blob/dev/release/2.0/repos/dir.props#L27.

natemcmaster: I originally did that, but wasn't sure if there might be upstream consumers also compiling .NET Framework assets in source-build. Is it safe to assume DotNetBuildFromSource=true means .NET Framework can always be skipped, even on windows?

.gitmodules
natemcmaster: NuGet depends on Microsoft.Extensions.CommandLineUtils 1.0.1. This branch was created from our v1.0.1 tag.

repos/common.proj
weshaggard: If you need it only on non-windows then you can still do your OS check in the common repo as well. It just moves the responsibility to the repo instead of source-build needing to know its custom logic. This is going to be the model as we get closer to just using the repo API.

natemcmaster: Ok, I updated to move this into aspnet/common instead.

PR #266, opened 2017-10-30
Always pass CoreCLR msbuildonunsupportedplatform

dagood:
This makes CoreCLR try to produce packages where it would normally skip doing so, e.g. Fedora 26. This allows the CoreCLR dependency to flow.


dagood:
Thanks for verifying!

I saw in CI that this arg is only supported in CoreCLR's build.sh, not build.cmd, so I added a fix for that.

PR #274, opened 2017-11-03
Add new OS bootstrap script

janvorli:
This change adds a script that automates creating bootstrap CLI.
While there is still space for improvements in the script, I think
it is time to publish it. I have used it to bring up RHEL 6 and
Alpine and tested it on FreeBSD for x64. I have also tested
it on arm64 Linux.

### DIFF
--- a/scripts/bootstrap/buildbootstrapcli.sh
+++ b/scripts/bootstrap/buildbootstrapcli.sh
@@ -1,0 +1,296 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+
+usage()
+{
+    echo "Builds a bootstrap CLI from sources"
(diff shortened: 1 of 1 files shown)


janvorli:
@chcosta this is just a separated script from the PR that you've already approved. As I have said, I want to spend more time on the doc, but get the script in now.
cc: @sunandabalu

chcosta:
Getting this in sooner than later is great.  Do you know if the plan is to replace https://github.com/dotnet/core/tree/master/tools/dotnet-bootstrap with a pointer to this and/or the doc?

LGTM

janvorli:
@chcosta I think this should be the replacement for the other tool. I've actually got some ideas from it when creating this one.

PR #279, opened 2017-11-06
Add new OS bootstrap doc

janvorli:
This change adds a document on creating a bootstrap cli for a new
unsupported OS.

### DIFF
--- a/Documentation/boostrap-new-os.md
+++ b/Documentation/boostrap-new-os.md
@@ -1,0 +1,70 @@
+# Adding support for a new OS
+## Bootstrap CLI
+When adding support for a new OS, both the native (C++) and managed (C#) code components in the coreclr, corefx and core-setup repos need to be compiled. Making the native components build on a new OS is a quite straightforward process and the tools like CMake, Clang C/C++ compiler, python and awk that are required for the build are available for almost all platforms. 
+But for the managed code compilation, we have a chicken and egg problem. A .NET CLI toolchain is needed to build managed components in the coreclr, corefx and core-setup repos, but it is not available for the target platform yet. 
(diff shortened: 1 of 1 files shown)


janvorli:
This was actually reviewed already in another PR, but I have removed the parts describing the usage of the seed cli for now and will create another PR for them once I fix it based on recent changes in the various repos that changed parts of the process.
cc: @sunandabalu

Documentation/boostrap-new-os.md
```diff
@@ -0,0 +1,70 @@
...
+```bash
+tar -xf dotnet-sdk-2.0.0-linux-x64.tar.gz
+```
+### Choosing RID for the new OS
+The new RID represents your target OS. It is used by developers to target that OS. The RID format is `<Name>.<Version>`. The name is lower case and should match the target OS. The version is optional and should be used if it is expected that in future versions the libraries that .NET Core depends on will not be binary compatible with the first supported version. So for example for FreeBSD 11, the RID would likely be `freebsd.11`.
+For OS that has `/etc/os-release` file, the RID needs to match the `$ID.$VERSION_ID` extracted from that file. 
+### Prerequisites
+The following libraries and tools need to be installed in order to build the bootstrap CLI. The precise names of the packages are OS specific.
+#### Tools
+* Clang version 3.5 or higher, 3.9 recommended
```
chcosta: Figuring out the dependencies and package names is one of the more annoying parts of trying to bootstrap.  Coreclr documentation also has a pretty good list of dependencies in their doc.  Ultimately, sadly, I think it just boils down to try to build and then install whatever coreclr is complaining about.  The list you have here is pretty good, it's too bad you can't just link to a central shared list though.  If anything changes, this is likely going to get out of sync pretty quickly.

janvorli: This list tends to really be os and distro specific with various small variations. So it is really not possible to put down a definitive os proof and distro proof list. 
The doc in coreclr is valid only for debian / red hat related distros that we already support.

PR #280, opened 2017-11-07
Proposal: Unify the format of all source build data passed to build.cmd/sh

tmat:
Having 3 formats for various pieces of data passed by source build to the build script (action, source build parameters and response file) makes parsing of the arguments unnecessarily complicated. Let's just use simple -name value format for all this data.

Related issue: dotnet#194

### DIFF
--- a/Documentation/RepoApi.md
+++ b/Documentation/RepoApi.md
@@ -15,26 +15,11 @@ The `.cmd` file will be called when running on Windows, while the `.sh` file wil
 
 These scripts will contain the logic necessary in order to build, test, publish the repo's assets.
 
-source-build will invoke these scripts on each repo passing in:
+source-build will invoke these scripts on each repo passing in a set of parameters.
 
-1. An action to perform
-1. A set of parameters
-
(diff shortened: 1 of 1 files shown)


tmat:
@eerhardt @natemcmaster @dagood @jaredpar @weshaggard

kasper3:
IMHO, this should also be unified in build scripts of corefx/coreclr. Currently we have all kinds of ways to pass the arguments https://github.com/dotnet/corefx/blob/9693747e046fc3e4da7ac5da8a1b101e549e8e01/src/Native/build-native.sh#L292

```sh
corefx/build-native.sh clang6.0
corefx/build-native.sh -clang6.0
corefx/build-native.sh --clang6.0
```

it is just complicating the scripts all over the place and nobody is getting any real benefit out of it.

maybe we can keep single hyphen options like `-clang6.0` as it works with powershell as well, but this is just a suggestion. If removal of other variants does break someone, that is not a huge pain to add or remove a `-` especially if it is done in favor of uniformity.

PR #283, opened 2017-11-09
Add scripts to make working with submodules easier.

### LINKED ISSUE #137: Working with submodules
While getting started in this repo I forget to update/initialize the submodules. I suspect that might be a common issue for folks so I was thinking we should consider adding a `git submodule update --init` to the beginning of the build scripts but given I'm pretty new to submodules I don't know if that will start to cause issue if there are changes to the modules or if there are patches applied to them. Anyone have enough experience to advise for or against always doing an update/init from the build scripts?

crummel:
Fixes #137.
- Make sure submodules are initialized and optionally do this for you.
- Check if submodules match their checked-in hash and optionally update them if they don't.
- Check for untracked changes in submodules and optionally clean them.
- Put which submodule versions we're using into the tarball.

### DIFF
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -20,7 +20,6 @@ if [ -e "$TARBALL_ROOT" ]; then
     echo "error '$TARBALL_ROOT' exists"
 fi
 
-
 SCRIPT_ROOT="$(cd -P "$( dirname "$0" )" && pwd)"
 
 if [ $SKIP_BUILD -ne 1 ]; then
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -70,3 +70,8 @@ do
     fi
 done
 
+# Record commits for the source-build repo and all submodules, to aid in reproducibility.
+echo -e "path\tchecked-in\tactual" > $TARBALL_ROOT/commits.txt
+echo -e "source-build\t$(git rev-parse HEAD)\t$(git rev-parse HEAD)" >> $TARBALL_ROOT/commits.txt
+git submodule foreach --recursive 'actual=$(git rev-parse HEAD); echo "$toplevel/$path\t$sha1\t$actual" >> $TARBALL_ROOT/commits.txt'
+
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -14,10 +14,12 @@ if [ "${2:-}" == "--skip-build" ]; then
 fi
 
 TARBALL_ROOT=$1
+FULL_TARBALL_ROOT=$(readlink -f $TARBALL_ROOT)
 
 if [ -e "$TARBALL_ROOT" ]; then
     echo "error '$TARBALL_ROOT' exists"
 fi
+
 
 SCRIPT_ROOT="$(cd -P "$( dirname "$0" )" && pwd)"
 
@@ -73,5 +75,5 @@ done
 # Record commits for the source-build repo and all submodules, to aid in reproducibility.
 echo -e "path\tchecked-in\tactual" > $TARBALL_ROOT/commits.txt
 echo -e "source-build\t$(git rev-parse HEAD)\t$(git rev-parse HEAD)" >> $TARBALL_ROOT/commits.txt
-git submodule foreach --recursive 'actual=$(git rev-parse HEAD); echo "$toplevel/$path\t$sha1\t$actual" >> $TARBALL_ROOT/commits.txt'
+git submodule foreach --quiet --recursive "actual=\$(git rev-parse HEAD); echo -e \"\$toplevel/\$path\t\$sha1\t\$actual\" >> $FULL_TARBALL_ROOT/commits.txt"
 
--- a/build.sh
+++ b/build.sh
@@ -10,7 +10,11 @@ if [ -z "${HOME:-}" ]; then
     mkdir "$HOME"
 fi
 
-source "$SCRIPT_ROOT/check-submodules.sh"
+if [[ "${SOURCE_BUILD_SKIP_SUBMODULE_CHECK:-default}" == "default" || $SOURCE_BUILD_SKIP_SUBMODULE_CHECK == "0" || $SOURCE_BUILD_SKIP_SUBMODULE_CHECK == "false" ]]; then
+  source "$SCRIPT_ROOT/check-submodules.sh"
+fi
+
+exit 0
 
 export DOTNET_CLI_TELEMETRY_OPTOUT=1
 export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
--- a/build.sh
+++ b/build.sh
@@ -14,8 +14,6 @@ if [[ "${SOURCE_BUILD_SKIP_SUBMODULE_CHECK:-default}" == "default" || $SOURCE_BU
   source "$SCRIPT_ROOT/check-submodules.sh"
 fi
 
-exit 0
-
 export DOTNET_CLI_TELEMETRY_OPTOUT=1
 export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
 export DOTNET_MULTILEVEL_LOOKUP=0
--- a/build.sh
+++ b/build.sh
@@ -10,6 +10,8 @@ if [ -z "${HOME:-}" ]; then
     mkdir "$HOME"
 fi
 
+source "$SCRIPT_ROOT/check-submodules.sh"
+
 export DOTNET_CLI_TELEMETRY_OPTOUT=1
 export DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
 export DOTNET_MULTILEVEL_LOOKUP=0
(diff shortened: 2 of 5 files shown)


crummel:
Fedora build looks like a timeout, retrying.
@dotnet-bot Test Fedora24 Release

check-submodules.ps1
```diff
@@ -0,0 +1,147 @@
+$answeredAll = $false   # has the user answered "yes to all" to init'ing submodules?
+$ProjectRoot = $PSScriptRoot
+
+# ask for confirmation and initialize a submodule if approved.
+function init_submodule($Path) {
+  $done = $false
+  while (-Not $done) {
+    Write-Warning "submodule $Path does not appear to be initialized."
+    Write-Host "Should I initialize it for you [Yes/no/all/quit]"
+    $answer = [Console]::ReadKey()
```
dagood: What happens if you're running the build noninteractively and a submodule isn't initialized? Will this hang?

build.sh
```diff
@@ -10,6 +10,8 @@ if [ -z "${HOME:-}" ]; then
     mkdir "$HOME"
 fi
 
+source "$SCRIPT_ROOT/check-submodules.sh"
```
dagood: Can you include an environment variable check or something so these can be disabled? Running without all submodules and/or with local changes is a typical dev flow for me, and I don't want to have to hit "n" interactively to get my build going.

I'm also curious how long these commands take, in case that's also something to consider.

check-submodules.sh
```diff
@@ -0,0 +1,146 @@
...
+      exit 1
+    elif [[ $REPLY =~ ^[Yy]$ ]]; then
+	  # check if we have this commit locally and can skip the fetch
+      git cat-file -e $expected^{commit} 2>/dev/null || exitCode=$?
+      if [ $exitCode != 0 ]; then
+        git fetch
+      fi
+      exitCode=0
+	  # double-check, we should have the commit now unless something
+	  # weird is going on.
```
dagood: Tab chars before the comments in this function are causing alignment issues.

check-submodules.sh
```diff
@@ -0,0 +1,146 @@
...
+    elif [[ $REPLY == "" || $REPLY == " " || $REPLY =~ ^[Nn]$ ]]; then
+      done=1
+    else
+      echo "didn't understand that ($REPLY)"
+    fi
+  done
+}
+
+# clean a submodule if the user approves.
+clean_submodule() {
```
dseefeld: I think it would make sense to have an "All" option on clean as well.

check-submodules.sh
```diff
@@ -0,0 +1,146 @@
...
+      echo "didn't understand that ($REPLY)"
+    fi
+  done
+}
+
+# We use the same script for checking the super-repo and the submodules.
+# Having the first argument be "in-submodule" triggers this submodule behavior.
+if [ ${1:-default} == "in-submodule" ]; then
+  expected_sha=$2
+  path=$3
```
dseefeld: Nit: In the functions, path is the first parameter followed by sha if needed.  Maybe this should be consistent.

build.sh
crummel: Added `SOURCE_BUILD_SKIP_SUBMODULE_CHECK`, any value other than `0` or `false` will turn it off.  This is taking about 17 seconds on my Windows machine and 2 seconds on Linux.

check-submodules.ps1
crummel: Yes, this would hang.  Is anything currently building non-interactive without a complete clean repo?  I can check for interactivity but calling the script from build.sh/ps1 counts as non-interactive, so it'd become a manual script only.

build-source-tarball.sh
```diff
@@ -70,3 +70,8 @@ do
     fi
 done
 
+# Record commits for the source-build repo and all submodules, to aid in reproducibility.
+echo -e "path\tchecked-in\tactual" > $TARBALL_ROOT/commits.txt
+echo -e "source-build\t$(git rev-parse HEAD)\t$(git rev-parse HEAD)" >> $TARBALL_ROOT/commits.txt
+git submodule foreach --recursive 'actual=$(git rev-parse HEAD); echo "$toplevel/$path\t$sha1\t$actual" >> $TARBALL_ROOT/commits.txt'
```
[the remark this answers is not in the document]
crummel: Yes to both, thanks.

build.sh
crummel: That's without any fixes necessary - submodules initialized, up-to-date, and clean.

check-submodules.sh
crummel: Fixed.

check-submodules.sh
```diff
@@ -0,0 +1,146 @@
...
+    elif [[ $REPLY =~ ^[Yy]$ ]]; then
+	  # check if we have this commit locally and can skip the fetch
+      git cat-file -e $expected^{commit} 2>/dev/null || exitCode=$?
+      if [ $exitCode != 0 ]; then
+        git fetch
+      fi
+      exitCode=0
+	  # double-check, we should have the commit now unless something
+	  # weird is going on.
+      git cat-file -e $expected^{commit} 2>/dev/null || exitCode=$?
```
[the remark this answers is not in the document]
crummel: Added a message for this.

check-submodules.sh
crummel: Flipped these.

check-submodules.ps1
dagood: I'm not aware of any places, but if it does happen even unintentionally, hangs are a pretty bad CI/etc. experience.

check-submodules.sh
crummel: Since the `submodule foreach`s are independent this is a little ugly, but I agree that this would be useful.  Let me know if this seems too hacky or if you have a better solution.

check-submodules.sh
```diff
@@ -0,0 +1,160 @@
...
+else
+  # submodule foreach doesn't work until submodules are init'd, read the modules manually
+  for module in `git config --file .gitmodules --get-regexp path | awk '{ print $2 }'`
+  do
+    if [ ! -e "$SCRIPT_ROOT/$module/.git" ]; then
+      if [ $answered_all == 1 ]; then
+        git submodule update --init --recursive $module
+      else
+        init_submodule $module
+     fi
```
dagood: Missing space

check-submodules.sh
```diff
@@ -0,0 +1,160 @@
...
+  prompt=$3
+  done=0
+  while [ $done == 0 ]; do
+    echo -e "$msg"
+    read -p "$prompt " -n 1 -r
+    echo
+    if [[ $REPLY =~ ^[Aa]$ ]]; then
+      git clean -fxd
+      git reset --hard HEAD
+      touch $CLEAN_ALL_SENTINEL
```
dagood: Values that may have spaces (especially paths) should be quoted.

check-submodules.ps1
dseefeld: Would it make sense to add a timeout on the input and then perform a default action so it wouldn't ever hang?

check-submodules.ps1
```diff
@@ -0,0 +1,163 @@
+$answeredAll = $false   # has the user answered "yes to all" to init'ing submodules?
+$ProjectRoot = $PSScriptRoot
+$CleanAllSentinel = "$ProjectRoot\.cleansourcebuildsubmodules"
+
+# ask for confirmation and initialize a submodule if approved.
+function init_submodule($Path) {
+  $done = $false
+  while (-Not $done) {
+    Write-Warning "submodule $Path does not appear to be initialized."
+    Write-Host "Should I initialize it for you [Yes/no/all/quit]"
```
dagood: Seems like the user might try to type the whole word and answer more prompts than they intended to. `[y]es / [n]o / [a]ll / [q]uit`?

check-submodules.ps1
dagood: That could work. `q` would be my default choice.

crummel: Since initializing a submodule is "safe" (won't ever lose work), I'm thinking "yes to all" makes sense for initialization.  I was thinking `no` for update and delete (in case you start the build and walk away) but I'm fine either way.

check-submodules.ps1
crummel: Changed as suggested.

check-submodules.ps1
dagood: Good point, `a` works for me.

check-submodules.sh
crummel: Fixed.

check-submodules.sh
crummel: Fixed.

crummel:
@dotnet-bot Test Fedora24 Release

crummel:
Fedora timeouts seem unrelated to these changes, merging now and I will keep an eye on the build and log an issue if it persists.

PR #291, opened 2017-11-16
Port F# patch removal to dev/release/2.0.

crummel:
Porting Brett's patch removal changes to dev/release/2.0.


crummel:
Fedora is the same timeout we've been seeing, merging.

PR #326, opened 2018-02-09
Update PlatformApis.cs - Update based on CR comments

dseefeld:
Missed @eerhardt review comments.  Adding them here.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -36,8 +36,7 @@
   <PropertyGroup>
     <BaseOutputPath>$(ProjectDir)bin/</BaseOutputPath>
     <ToolsDir>$(ProjectDir)Tools/</ToolsDir>
-    <TaskDirectory>$(ProjectDir)tools-local/tasks/</TaskDirectory>
-    <TasksBinDir>$(TaskDirectory)Microsoft.DotNet.SourceBuild.Tasks/bin/Debug/netstandard1.5/</TasksBinDir>
+    <TasksBinDir>$(ProjectDir)bin/AnyCPU/Debug/netstandard1.5/</TasksBinDir>
     <BaseIntermediatePath>$(BaseOutputPath)obj/</BaseIntermediatePath>
     <OutputPath>$(BaseOutputPath)$(Platform)/$(Configuration)/</OutputPath>
     <IntermediatePath>$(BaseIntermediatePath)$(Platform)/$(Configuration)/</IntermediatePath>
--- a/dir.props
+++ b/dir.props
@@ -36,7 +36,8 @@
   <PropertyGroup>
     <BaseOutputPath>$(ProjectDir)bin/</BaseOutputPath>
     <ToolsDir>$(ProjectDir)Tools/</ToolsDir>
-    <TasksBinDir>$(ProjectDir)bin/AnyCPU/Debug/netstandard1.5/</TasksBinDir>
+    <TaskDirectory>$(ProjectDir)tools-local/tasks/</TaskDirectory>
+    <TasksBinDir>$(TaskDirectory)Microsoft.DotNet.SourceBuild.Tasks/bin/Debug/netstandard1.5/</TasksBinDir>
     <BaseIntermediatePath>$(BaseOutputPath)obj/</BaseIntermediatePath>
     <OutputPath>$(BaseOutputPath)$(Platform)/$(Configuration)/</OutputPath>
     <IntermediatePath>$(BaseIntermediatePath)$(Platform)/$(Configuration)/</IntermediatePath>
--- a/tools-local/init-build.proj
+++ b/tools-local/init-build.proj
@@ -26,7 +26,7 @@
   </Target>
 
   <Target Name="BuildTasks">
-    <Exec Command="$(DotNetCliToolDir)dotnet build tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj" />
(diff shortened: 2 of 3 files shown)


tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj
```diff
@@ -2,15 +2,16 @@
   <PropertyGroup>
     <TargetFramework>netstandard1.5</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <CoreSetupPlatformAbstractionsPath>../../../src/core-setup/src/managed/Microsoft.DotNet.PlatformAbstractions/</CoreSetupPlatformAbstractionsPath>
```
weshaggard: Please use `$(ProjectDir)src/core-setup/src/managed/Microsoft.DotNet.PlatformAbstractions/` That way we don't easily break the relative paths by moving this project.

tools-local/init-build.proj
```diff
@@ -26,7 +26,7 @@
   </Target>
 
   <Target Name="BuildTasks">
-    <Exec Command="$(DotNetCliToolDir)dotnet build tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj" />
+    <Exec Command="$(DotNetCliToolDir)dotnet build tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj /p:SubmoduleDirectory=$(SubmoduleDirectory)" />
```
weshaggard: Why did you need to pass it in explicitly?

dseefeld: I was getting a weird failure after I updated it to include dir.props and so I thought maybe I was getting more than I wanted by including all of dir.props, so I switched it to just include the directory directly.  I'm going to switch it back and see if I can figure out the errors.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj
```diff
@@ -1,16 +1,22 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <!-- We use dotnet-cli to build BuildTasks, so we need to disable BuildTools Roslyn properties. -->
+    <DisableBuildToolsRoslynVersion>true</DisableBuildToolsRoslynVersion>
```
weshaggard: It would seem as though we may need to split dir.props up a little bit if we cannot easily import it in our projects.

eerhardt: We do the same thing in `core-setup`.

With the move to using the .NET SDK's MSBuild work that Jose is doing, this should get better with time.  Eventually this can just go away.

dir.props
```diff
@@ -36,8 +36,7 @@
   <PropertyGroup>
     <BaseOutputPath>$(ProjectDir)bin/</BaseOutputPath>
     <ToolsDir>$(ProjectDir)Tools/</ToolsDir>
-    <TaskDirectory>$(ProjectDir)tools-local/tasks/</TaskDirectory>
-    <TasksBinDir>$(TaskDirectory)Microsoft.DotNet.SourceBuild.Tasks/bin/Debug/netstandard1.5/</TasksBinDir>
+    <TasksBinDir>$(ProjectDir)bin/AnyCPU/Debug/netstandard1.5/</TasksBinDir>
```
weshaggard: Isn't that going to be the wrong output for all other projects?

dseefeld: No.  This should be correct for all projects.

weshaggard: We are now putting the task dll directly in that directory? It seems a little odd that we no longer put it under a directory with it's name in the path.

dseefeld: I'm going to update OutputPath to put it back under the tasks directory.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj
```diff
@@ -1,16 +1,23 @@
...
+  </PropertyGroup>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
   <PropertyGroup>
     <TargetFramework>netstandard1.5</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <CoreSetupPlatformAbstractionsPath>$(SubmoduleDirectory)core-setup/src/managed/Microsoft.DotNet.PlatformAbstractions/</CoreSetupPlatformAbstractionsPath>
+    <OutputPath>$(MSBuildThisFileDirectory)bin/$(Configuration)</OutputPath>
```
eerhardt: I thought we wanted all build output to go into the root `bin` directory, and not into `bin` directories at each directory level?

weshaggard: good point, we should scope it to under the root bin folder.

dseefeld: This is where it was located before I made the change.  Including dir.props caused it to move, which caused other issues with usage of tasks and building the tarball.  I'm putting it back where it was before.

PR #340, opened 2018-03-08
Fix roslyn-tools offline build

crummel:
- Add source-build sources to the environment variables roslyn-tools uses for extra feeds.
- Use RepoBuild instead of BuildCommand to propagate environment variables.
- Add roslyn-tools back to known-good.


crummel:
The OSX failure is unrelated, so this is ready for review @dseefeld

Issue #349, opened 2018-03-09
Ensure source-build tarball build can build offline


dseefeld:
Building the produced tarball should be possible when machine is offline.
CI for the Offline build should actually build offline.

Currently, building the tarball offline causes the following error:
`  /src/tarball-output/Tools/dotnetcli/sdk/2.1.1-preview-007094/NuGet.targets(103,5): error : Failed to download package 'NuGet.Packaging.4.3.0-preview3-4168' from 'https://api.nuget.org/v3-flatcontainer/nuget.packaging/4.3.0-preview3-4168/nuget.packaging.4.3.0-preview3-4168.nupkg'. [/src/tarball-output/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj] [/src/tarball-output/tools-local/init-build.proj]
  /src/tarball-output/Tools/dotnetcli/sdk/2.1.1-preview-007094/NuGet.targets(103,5): error : An error occurred while sending the request. [/src/tarball-output/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj] [/src/tarball-output/tools-local/init-build.proj]
  /src/tarball-output/Tools/dotnetcli/sdk/2.1.1-preview-007094/NuGet.targets(103,5): error : Couldn't resolve host name [/src/tarball-output/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj] [/src/tarball-output/tools-local/init-build.proj]
  /src/tarball-output/Tools/dotnetcli/sdk/2.1.1-preview-007094/NuGet.targets(103,5): error : Failed to retrieve information about 'NuGet.Packaging' from remote source 'https://api.nuget.org/v3-flatcontainer/nuget.packaging/index.json'. [/src/tarball-output/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj] [/src/tarball-output/tools-local/init-build.proj]
[...]

PR #353, opened 2018-03-16
Repair offline build

dseefeld:
Updates to known-good projects to make sure that they don't use network resources.  I removed roslyn-tools and coreclr from known-good because they had some larger issues and I want to get this going in CI now.

### DIFF
--- a/repos/websdk.proj
+++ b/repos/websdk.proj
@@ -15,7 +15,7 @@
     <RepositoryReference Include="newtonsoft-json" />
   </ItemGroup>
 
-  <Target Name="MakeWritable" AfterTargets="ApplyPatches">
+  <Target Name="MakeExecutable" AfterTargets="ApplyPatches">
     <Exec Command="chmod +x $(ProjectDirectory)/build.sh"
           WorkingDirectory="$(ProjectDirectory)"
           EnvironmentVariables="@(EnvironmentVariables)"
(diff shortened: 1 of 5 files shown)


dseefeld:
Fixes #349

tools-local/init-build.proj
```diff
@@ -26,7 +26,10 @@
   </Target>
 
   <Target Name="BuildTasks">
-    <Exec Command="$(DotNetCliToolDir)dotnet build tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj" />
+    <PropertyGroup Condition="'$(OfflineBuild)' == 'true'">
+      <OfflineSources>$(PrebuiltPackagesPath)</OfflineSources>
+    </PropertyGroup>
+    <Exec Command="$(DotNetCliToolDir)dotnet build tasks\Microsoft.DotNet.SourceBuild.Tasks\Microsoft.DotNet.SourceBuild.Tasks.csproj /p:RestoreSources=$(OfflineSources)" />
```
crummel: Does this work online?  I'd expect blank RestoreSources to cause issues but if that works I'm fine with it.

repos/websdk.proj
```diff
@@ -15,5 +15,12 @@
     <RepositoryReference Include="newtonsoft-json" />
   </ItemGroup>
 
+  <Target Name="MakeWritable" AfterTargets="ApplyPatches">
```
crummel: This is actually `MakeExecutable`, right?

tools-local/init-build.proj
dseefeld: This does work online.  blank causes it to use the default.

Issue #370, opened 2018-03-20
When source-build updates to a ProdCon build's commits, commit hash and depth should update


dagood:
https://github.com/dotnet/source-build/issues/342 adds submodule update, but there is some hard-coded info about the submodules in the project files that isn't updated yet. (Commit hash, commit depth/count.)

This tracks auto-updating that metadata when the submodule is updated.

A few obstacles:

 * GitHub doesn't seem to have an API for commit depth of a commit. The only API solutions I could find online are traversing the entire list of commits (using a large chunk of the account's rate limit) and counting pages (which assumes that the number of commits each page has is constant). What I have in mind is pointing the updater to the submodule, then it can `git fetch ...` and use Git commands to get the depth.
 * There doesn't seem to be a standard name for the commit hash and depth variables yet. I think standardization is the way to go to keep the updater config (source-build's `dependencies.props/targets`) straightforward.

Issue #342, opened 2018-03-08
Prodcon build output SHAs are updated to source-build automatically


dleeapho:
Every prodcon build presents a coherent set of repo states that produces a product build. We should take the SHAs in the manifest and update the submodule SHAs in source-build automatically.

dagood:
PR https://github.com/dotnet/source-build/pull/371 to use the updater is created. Next steps are to create the Maestro subscription and a source-build general executor VSTS def to run the PR submission targets.

dagood:
Here's the test submodule update to the ProdCon build: https://github.com/dotnet/source-build/pull/383. Didn't work, as expected, since our fixes aren't in the ProdCon branches yet.

I think I'll leave it open to soak up new ProdCon submodule updates. (It'll also get the BuildTools updates in it if someone makes a BuildTools release/2.1 change.)

/cc @dleeapho @dseefeld @crummel @weshaggard

Issue #270, opened 2017-10-31
Add BuildToolsVersion.txt auto-updater


dagood:
This will help us see BuildTools breaking changes earlier and avoid falling behind.

dagood:
Completed by #371.

PR #371, opened 2018-03-20
Auto-update submodules to the commits used in latest ProdCon build

dagood:
To try it out, use `build.cmd /t:UpdateDependencies` to update based on the `ProdConCurrentRef` source of truth (e.g. if you want to pin to a specific ProdCon build) and `build.cmd /t:UpdateToRemoteDependencies` to update to the latest manifest on dotnet/versions master.

Uses new updater and dependency info from https://github.com/dotnet/buildtools/pull/1968.

There's some metadata checked into the project files that this updater won't change. I have a followup issue for that: https://github.com/dotnet/source-build/issues/370.

https://github.com/dotnet/source-build/issues/342

Also adds a BuildTools updater: https://github.com/dotnet/source-build/issues/270

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -26,6 +26,8 @@
     <RemoveDir Directories="$(BaseOutputPath)" />
   </Target>
 
+  <Import Project="$(ProjectDir)dependencies.targets" />
+
   <Import Project="$(ToolsDir)VersionTools.targets" />
 
 </Project>
--- a/dependencies.props
+++ b/dependencies.props
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!--
     A CurrentRef is a commit on the dotnet/versions master branch. It is the single source of truth
     for which dotnet/versions commit was last used to update the dependency.
--- a/dependencies.props
+++ b/dependencies.props
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--
+    A CurrentRef is a commit on the dotnet/versions master branch. It is the single source of truth
+    for which dotnet/versions commit was last used to update the dependency.
+  -->
+  <PropertyGroup>
+    <ProdConCurrentRef>7daa2e55de668c4c8ee832059aab04917153881e</ProdConCurrentRef>
+    <BuildToolsCurrentRef>af030b25653d53483b0200ff81d89209bd30549a</BuildToolsCurrentRef>
+  </PropertyGroup>
+
+  <!-- Package dependency verification/auto-upgrade configuration. -->
+  <PropertyGroup>
+    <BaseDotNetBuildInfo>build-info/dotnet/</BaseDotNetBuildInfo>
+    <DependencyBranch>release/2.1</DependencyBranch>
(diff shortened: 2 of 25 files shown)


dependencies.props
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--
+    A CurrentRef is a commit on the dotnet/versions master branch. It is the single source of truth
+    for which dotnet/versions commit was last used to update the dependency.
+  -->
+  <PropertyGroup>
+    <ProdConCurrentRef>7daa2e55de668c4c8ee832059aab04917153881e</ProdConCurrentRef>
```
dagood: For context, this points to https://github.com/dotnet/versions/blob/7daa2e55de668c4c8ee832059aab04917153881e/build-info/dotnet/product/cli/release/2.1-preview1/build.xml#L2938-L2953.

dagood:
I forgot to actually update BuildTools here, and that this branch pulls BuildTools release/2.1. Submitted https://github.com/dotnet/buildtools/pull/1975 to port the updater to 2.1 BuildTools, then I'll update this PR.

build.proj
```diff
@@ -26,7 +26,7 @@
     <RemoveDir Directories="$(BaseOutputPath)" />
   </Target>
 
-  <Import Project="$(ProjectDir)dependencies.targets" />
+  <Import Project="$(ProjectDir)dependencies.targets" Condition="Exists('$(ProjectDir)dependencies.targets')" />
```
eerhardt: I'm not sure I understand why you added the condition.  You are adding the files, so they will always exists, right?

dagood: Kinda described in the commit message:

> There's no need for the BfS tarball to include these files: it doesn't need to run repo-global auto-updates like submodule update.

The project files that get into the tarball are manually listed here:

https://github.com/dotnet/source-build/blob/e2ecb4c403f4d6e30137cf181ce2b43f129618b6/build-source-tarball.sh#L41-L49

I figure things that aren't used in the tarball build aren't included, and there's no need to add these two files to that list.

@crummel is that the intention? Does it need to be so explicit?

dagood:
Look like the buildtools upgrade broke the build in general. Lots of:

```
error MSB4064: The "OverrideToolHost" parameter is not supported by the "Csc" task. Verify the parameter exists on the task, and it is a settable public instance property.
error MSB4063: The "Csc" task could not be initialized with its input parameters.
```

Makes sense: BuildTools merged `master` into `release/2.1` for preview2. Similar merges have also happened in the submodule repos, but we have the old preview1 commits checked out.

I think the cleanest way to get this feature in is to update to preview2 commits first. Getting it working should involve updating BuildTools, so this PR wouldn't need to handle upgrading BuildTools over the preview1-preview2 border.

dagood:
I talked with @eerhardt and @weshaggard about the CI errors in dotnet/standard. The plan is to update standard's release/2.0.0 branch to use a compatible BuildTools. @eerhardt pointed out that the change to use the CLI's copy of MSBuild rather than the one from BuildTools is probably what's breaking--I'm working on fixing and making a dotnet/standard PR.

What I was talking about above about 2.1-preview1 vs. 2.1-preview2 doesn't really apply in this case, since we're still building dotnet/standard 2.0.

build.proj
crummel: I think we want to keep the tarball as minimal as possible so if the workaround isn't causing too much extra work I think we'd prefer to keep it out.  If it's causing issues we can include the new files.

dagood: The only concern I have is that someone could add something to dependencies.props/targets and expect it to be set during all source-build builds. Not sure how easy that would be to debug depending on what's added.

Keeping it out of the tarball is a bit of complexity that I'm not sure we need. But maintaining that list of files is also annoying... would `*.proj`, `*.props`, `*.targets` at the root be ok?

crummel: That sounds fine to me.

dagood:
@dseefeld @crummel PTAL

PR #376, opened 2018-03-21
Update BuildTools to 2.1.0-preview2-02621-01

dagood:
Update dotnet/standard commit to bring in fixes for the new version (https://github.com/dotnet/standard/pull/690), and fix init-tools.sh (same fix as standard).

Unblocks https://github.com/dotnet/source-build/pull/371


dagood:
Figured out the weirdness with tarball, I'm pretty sure. Repro steps from fresh dotnet/standard@2022919b29bc37746f760bc6391171459f613ea5 standalone clone:

 1. `./build.sh`
 2. `rm -rf bin/ packages/`
 3. `./build.sh`

Tarball build does keep some packages, but I guess not the ones that are needed. I believe BuildTools handles the package not being there by building with the old Roslyn, which causes the failures like these:

```
/home/dagood/git/standard/Tools/Microsoft.CSharp.Core.targets(106,11): error MSB4064: The "OverrideToolHost" parameter is not supported by the "Csc" task. Verify the parameter exists on the task, and it is a settable public instance property. [/home/dagood/git/standard/netstandard/ref/netstandard.csproj]
/home/dagood/git/standard/Tools/Microsoft.CSharp.Core.targets(67,5): error MSB4063: The "Csc" task could not be initialized with its input parameters.  [/home/dagood/git/standard/netstandard/ref/netstandard.csproj]
```

It seems like `init-tools` itself brings down some packages that aren't being saved? @crummel does this sound reasonable?

dagood:
I added a step to the tarball creation script that runs `Tools/init-tools.sh` to populate the `packages/` dir with the package BuildTools now expects to be there.

Removed CoreCLR from known-good temporarily to see if this makes it past CI. This works on my machine with I limit the build to `standard`. (My docker install doesn't seem to be working, so I can't test the offline tarball build, but the online one seems to work.)

dagood:
Green CI, but with CoreCLR taken out. Locally, I'm trying a merge of `release/2.1` into @weshaggard's fix branch to see if that works. The merge is clean, BuildTools version is `2.1.0-preview2-02612-03` and CLI `2.1.2`, which means it should be past at least the known break.

init-tools.msbuild
```diff
@@ -8,5 +8,6 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.DotNet.BuildTools" Version="$(BuildToolsPackageVersion)" />
     <PackageReference Include="Microsoft.DotNet.BuildTools.Coreclr" Version="1.0.4-prerelease" />
+    <PackageReference Include="ILLink.Tasks" Version="0.1.5-preview-1461378" />
```
dagood: I'll add this to `dependencies.props` in my other PR (where I introduce `dependencies.props`) to match https://github.com/dotnet/coreclr/blob/2e5f43401b458315dc33c137f0ee0c5a02d810c2/init-tools.msbuild#L20-L22.

dagood:
CI's green! Once https://github.com/dotnet/coreclr/pull/17136 is in CoreCLR release/2.1 we can move off my fork.

dagood:
@dseefeld @crummel PTAL

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "src/coreclr"]
 	path = src/coreclr
-	url = https://github.com/dotnet/coreclr
+	url = https://github.com/dagood/coreclr
```
weshaggard: Note a huge deal but I generally perfer to not switch to forks but instead just use a branch in the main repo.

weshaggard:
@dagood did you move coreclr to the tip of the release/2.1 branch + my fix? If so and that works then I can update the release/2.1 branch in coreclr to have the fix. It should be noted that my fix isn't enough and needs adjustments see (https://github.com/dotnet/coreclr/pull/17226) to ensure other scenarios still work.

dagood:
Yep, my commit is just your fix branch with release/2.1 tip merged into it. I'll push it to `dev/dagood/source-build-fixes-preview2` in dotnet/coreclr to unblock for now, then we can switch back when you have the adjustments in.

support/tarball/build.sh
```diff
@@ -12,5 +12,25 @@ export NUGET_PACKAGES="$SCRIPT_ROOT/packages/"
 
 MSBUILD_ARGUMENTS=("/p:OfflineBuild=true" "/flp:v=detailed" "/clp:v=detailed")
 
+echo "Expanding BuildTools dependencies into packages directory..."
+# init-tools tries to copy from its script directory to Tools, which in this case is a copy to
+# itself. This is an error. To avoid the error, use a temp dir that we immediately delete.
+TEMP_TOOLS_DIR="$SCRIPT_ROOT/ToolsTemp"
+PREBUILT_PACKAGE_SOURCE="$SCRIPT_ROOT/prebuilt/nuget-packages"
+(
+    set -x
```
dseefeld: Did you mean to leave this here?

dagood: Yeah, this makes the commands show up in the logs without the maintenance cost and risk of duplicating the commands in `echo ...` statements.

support/tarball/build.sh
```diff
@@ -12,5 +12,25 @@ export NUGET_PACKAGES="$SCRIPT_ROOT/packages/"
...
+
+    "$CLI_ROOT/dotnet" restore "$SCRIPT_ROOT/init-tools.msbuild" --no-cache --packages "$SCRIPT_ROOT/packages" --source "$PREBUILT_PACKAGE_SOURCE" || exit $?
```
dagood: The `|| exit $?` pattern is what I came up with to make sure it exits on the first error. `set -e` would normally do this, but it doesn't apply when the command uses `||`. (I used `||` on the subshell to notify the user that it failed rather than just having the script exit without any info.) Better ideas welcome.

PR #377, opened 2018-03-22
Mark dependencies that aren't required since 2.1

kasper3:
These dependencies are removed from CoreCLR for v2.1.


kasper3:
cc @janvorli, @omajid

janvorli:
I would prefer keeping them in, but state that they are not needed for 2.1. It is possible that someone will need to create a new bootstrap cli for 2.0 in the future. The CLI used to pull build tools in coreclr, corefx and core-setup repos is not 2.1 yet and it is not clear when it would get to 2.1.

kasper3:
@janvorli, agree and updated.

PR #388, opened 2018-03-30
Make ProdCon build auto-updates mergeable by skipping some repos

dagood:
* Stop updating MSBuild because it's on a custom branch
 * Stop updating CoreCLR and CoreFX because the update currently removes some commits that haven't been in a ProdCon build yet.
 * Call `/t:UpdateToRemoteDependencies` to simulate an auto-upgrade and see if it's green.
 * ~~Automatically stop updating repos that have patches~~

I didn't try filtering by known-good or anything like that, just the minimum to try to make the auto-PRs mergeable.

The BuildTools upgrade will conflict with @weshaggard's CoreFX PR, waiting to resolve it in this PR.

I filed a few submodule updater bugs that I hit: https://github.com/dotnet/buildtools/issues/1990, https://github.com/dotnet/buildtools/issues/1991. (They only affect local upgrade attempts.)


dagood:
@dotnet-bot test RHEL7.2 Unshared Release
Looks like nuget.org flakiness. https://ci.dot.net/job/dotnet_source-build/job/dev_release_2.1/job/RHEL7.2_Unshared_Release_prtest/47/

repos/dir.targets
```diff
@@ -219,5 +217,14 @@
   <Target Name="GetProjectDirectory" Outputs="$(ProjectDirectory)" />
   <Target Name="GetOrchestratedManifestBuildName" Outputs="$(OrchestratedManifestBuildName)" />
 
+  <Target Name="DisablePatchedRepoSubmoduleAutoUpdate"
```
dseefeld: I think we're going to have repos that we're going to want to update, even though they have patches.  Is there a different way to trigger the disable of auto update?  Or should we have an override for those repos?

dagood: I was thinking patches are brittle enough that they would break upgrade PR CI often, so we'd never want to upgrade them. The `<OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>` can certainly be inlined into the specific repo.proj files though.

dagood: We talked offline--I'm going to let these update. The idea is that we'll be able to see when the patches break and fix them at that point, so we won't get far behind.

(I'll make this change once the CoreFX PR goes in and I resolve the conflicts.)

repos/coreclr.proj
```diff
@@ -11,6 +11,8 @@
     <LatestCommit>36f70fa4be4bd37d4d76d06dd2cc433ff46351bd</LatestCommit>
     <VersionSeedDate>2017-07-19</VersionSeedDate>
     <OfficialBuildId>20170719-03</OfficialBuildId>
+    <!-- CoreCLR is part of ProdCon, but source-build fixes haven't been in a build yet. -->
```
weshaggard: Will be fixed with https://github.com/dotnet/source-build/pull/384

repos/dir.targets
```diff
@@ -4,8 +4,6 @@
     <EnvironmentVariables Include="LatestCommit=$(LatestCommit)" />
     <EnvironmentVariables Include="OfficialBuildId=$(OfficialBuildId)" />
     <EnvironmentVariables Include="VersionSeedDate=$(OfficialBuildId)" />
-    <!-- Work around a dependency on .git that doesn't have a good fallback yet: https://github.com/dotnet/buildtools/issues/1980 -->
```
weshaggard: I removed this with https://github.com/dotnet/source-build/pull/384 as part of my buildtools update.

repos/coreclr.proj
dagood: Doesn't look like it, but I'm rewording the comment because I think it was unclear: `ProdCon is behind source-build: temporarily disable auto-update.`

weshaggard: Source-build currently points to the latest commit in coreclr release/2.1 branch. What is missing?

weshaggard: Oh.. I think I understand your comment now. Your correct there hasn't been a published prodcon build with this sha yet. That should be true for corefx as well.

weshaggard: I see you disabled corefx as well.

dagood: Yep, you got it. Sorry for the confusion.

PR #390, opened 2018-03-30
Add symreader and symreader-portable to known-good

dagood:
These repos have repo toolset and build script fixes and should work now. Local tarball and offline tarball builds (with just these two repos included) seem good.


dagood:
@dotnet-bot test this please

PR #394, opened 2018-04-04
Add fsharp to known-good

dagood:
VisualFSharp changes are based on https://github.com/Microsoft/visualfsharp/pull/4674, which adds the `DotNetPackageVersionPropsPath` repo API.

### DIFF
--- a/repos/dir.targets
+++ b/repos/dir.targets
@@ -78,16 +78,21 @@
 
   <Target Name="UpdateNuGetConfig"
           BeforeTargets="Build"
-          Condition="'$(NuGetConfigFile)' != ''">
-    <RemoveInternetSourcesFromNuGetConfig NuGetConfigFile="$(NuGetConfigFile)"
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -49,7 +49,7 @@
 	branch = release/2.0.0
 [submodule "src/fsharp"]
 	path = src/fsharp
-	url = https://github.com/Microsoft/VisualFSharp
+	url = https://github.com/dagood/VisualFSharp
```
crummel: Looks like all the F# changes are in the upstream repo, I don't think we should need to change this?

dagood: They aren't quite yet. These are the new ones, one from https://github.com/Microsoft/visualfsharp/pull/4674 and a few fixup attempts by me: https://github.com/dagood/VisualFSharp/compare/6417ab1...c5f5c2c2e5a6178e9c260e83605b699b08080996

PR #395, opened 2018-04-04
Add CentOS7.1 online tarball CI leg

dagood:
This will hopefully let us get repro machines easier for this leg.

The *offline* tarball works fine as-is, since it actually uses Ubuntu with RHEL Docker containers. But it's possible for something to fail in the online leg that doesn't fail offline.

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -60,7 +60,7 @@ def addPushJob(String project, String branch, String os, String configuration)
 
 // Tarball builds that are not enforced to be offline
 [true, false].each { isPR ->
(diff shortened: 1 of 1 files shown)


crummel:
Kicked off https://ci.dot.net/job/dotnet_source-build/job/dev_release_2.1/job/GenPRTest/job/CentOS7.1_Tarball_Release/1/ from the new job, if that works this looks good to me.

dagood:
It passed! Thanks @crummel.

PR #397, opened 2018-04-04
Add roslyn-tools to known good, flow built tooling

dagood:
* Add `roslyn-tools` repo.
 * Remove patches, and clean up one I don't think I saw when I added `symreader` at first.
 * Flow the source-built roslyn-tools into all users, including the offline roslyn-tools build. Update symreader and symreader-portable to get functionality that can uptake the tooling.

/cc @tmat


repos/dir.targets
```diff
@@ -185,6 +186,24 @@
...
 
+  <Target Name="ExtractToolPackage"
+          DependsOnTargets="GatherBuiltPackages"
+          AfterTargets="Build"
+          Condition="'$(BuiltToolPackageId)' != ''">
+    <ItemGroup>
+      <_AllSourceBuiltPackages Include="$(SourceBuiltPackagesPath)*.nupkg" />
+      <_ToolPackage Include="@(_AllSourceBuiltPackages)"
+                    Condition="$([System.String]::new('%(Filename)').StartsWith('$(BuiltToolPackageId)'))" />
+    </ItemGroup>
```
dagood: I'm not too happy with this target. I think ideally we'd have something that would `dotnet restore` the toolset package based on `PackageVersions.props` to avoid the `StartsWith` fuzziness and manual zip file extraction.

repos/roslyn-tools.proj
```diff
@@ -2,27 +2,20 @@
...
     <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
+    <LatestCommit>9bb9aec2fdfd165eef3776fd3ff2f1bb9b63e46d</LatestCommit>
+    <UsesRepoToolset>true</UsesRepoToolset>
+    <BuiltToolPackageId>$(RepoToolsetPackageId)</BuiltToolPackageId>
   </PropertyGroup>
 
   <ItemGroup>
     <EnvironmentVariables Include="UsingToolMicrosoftNetCompilers=false" />
     <!-- Match the version symreader wants - this is the highest among our client repos -->
     <EnvironmentVariables Include="BUILD_BUILDNUMBER=20180215.01" />
```
dseefeld: Is there a plan for keeping the BuildNumber updated?

dagood: Hmm, I don't think it needs to be set to anything in particular anymore, since now it's auto-updated rather than matched up forcefully. I'm going to try removing this line to produce dev-versioned builds.

dir.props
```diff
@@ -17,6 +17,8 @@
     <DotNetCliToolDir>$(ProjectDir)Tools/dotnetcli/</DotNetCliToolDir>
     <PatchesDir>$(ProjectDir)patches/</PatchesDir>
     <PackagesDir>$(ProjectDir)packages/</PackagesDir>
+    <DotNetSdkDir>$(DotNetCliToolDir)sdk/$(SDK_VERSION)/</DotNetSdkDir>
```
weshaggard: I'm not too key on depending on the env variable here it does mean that it will only work when you call build.sh it won't even work for build.cmd.

dagood: Hmm, yeah, I should have just added a `/p`. Maybe there's even a property that MSBuild sets itself with the location of the SDK that I could have used, didn't look for that either.

PR #401, opened 2018-04-06
Add sdk, CliCommandLineParser, cli-migrate to known-good

dagood:
Uses patches against a dev branch in the main repos.

/cc @dleeapho @weshaggard

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -39,10 +39,10 @@
 	url = https://github.com/Microsoft/ApplicationInsights-dotnet
 [submodule "src/clicommandlineparser"]
 	path = src/clicommandlineparser
-	url = https://github.com/dotnet/clicommandlineparser
+	url = https://github.com/dagood/clicommandlineparser
 [submodule "src/cli-migrate"]
 	path = src/cli-migrate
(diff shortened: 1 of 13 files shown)


dagood:
> Change to 2.1.101 SDK

Looking back, I don't think I hit a concrete error due to this. My understanding is that it's required to build offline for 2.1-preview2, though.

> Remove --log option from defaults

This was removed as an option (https://github.com/dotnet/sdk/compare/release/2.1.3xx...b8a95329978628caeae3c89634d7235ed726ee69#diff-fec83b7e693a8e104fa9713533e1c054L69), but it still was in the root `build.sh`.

dagood:
The SDK fails to build tests in the non-offline tarball build when it tries to use xunit: https://ci.dot.net/job/dotnet_source-build/job/dev_release_2.1/job/RHEL7.2_Tarball_Release_prtest/101/

@tmat Can you try to disable the SDK repo's test projects when `DotNetBuildFromSource` is set?

dagood:
Filed https://github.com/dotnet/source-build/issues/405 to remove these patches.

PR #404, opened 2018-04-09
Add roslyn repo to known-good

dseefeld:
* Roslyn repo has been updated to incorporate previous patches into source.  Changes are on a private branch for now. 
* Added updaters for global.json and Tools.props to update CLI version.
* Added a patch to include solution to build, since solution included in roslyn branch is currently incorrect.
* Added patch for SDK PR1997 which is required until source-build moves to a newer CLI version than 2.1.101

cc @333fred, @jaredpar


patches/0001-Patch-Tools-for-sdk-pr-1997.patch
```diff
@@ -0,0 +1,25 @@
+From 735b7c82a16ddf16f39643830acb2ee62d2f3150 Mon Sep 17 00:00:00 2001
```
dagood: I think this patch should live in `tools-local/patches` or something like that--to clarify what kind of patch this is and keep repo patches separate.

init-tools.sh
```diff
@@ -186,6 +186,9 @@ ls "$__scriptpath/Tools/scripts/docker/"*.sh | xargs chmod +x
 
 "$__scriptpath/Tools/crossgen.sh" "$__scriptpath/Tools"
 
+echo "Patching Sdk to workaround issue https://github.com/dotnet/sdk/pull/1997"
+git apply --ignore-whitespace --whitespace=nowarn --directory="$__scriptpath" "$__scriptpath/patches/0001-Patch-Tools-for-sdk-pr-1997.patch"
```
dagood: Is it reasonable to do this during `tools-local/init-build.proj` to avoid customizing this script? (With some semaphore in the Tools dir to avoid attempting it more than once?)

repos/roslyn.proj
```diff
@@ -27,15 +30,26 @@
     <NuSpecFiles Include="$(ProjectDirectory)/src/NuGet/Microsoft.CodeAnalysis.Compilers.nuspec" />
     <NuSpecFiles Include="$(ProjectDirectory)/src/NuGet/Microsoft.CodeAnalysis.VisualBasic.nuspec" />
     <NuSpecFiles Include="$(ProjectDirectory)/src/NuGet/Microsoft.Net.Compilers.netcore.nuspec" />
-    <NuSpecFiles Include="$(ProjectDirectory)/src/NuGet/Microsoft.CodeAnalysis.Build.Tasks.nuspec" />
   </ItemGroup>
 
+  <Target Name="UpdateCLIVersion"
```
dagood: Could this be in `dir.targets` with the json and props paths parameterized? (But we can put the move off until we use it to remove patches in other repos too.)

dseefeld: I'd rather leave this here for now and move it when we need it for others.

repos/roslyn.proj
```diff
@@ -27,15 +30,26 @@
...
+          BeforeTargets="Restore">
+
+    <UpdateJson JsonFilePath="$(ProjectDirectory)/global.json"
+                PathToAttribute="sdk.version"
+                NewAttributeValue="$(SDK_VERSION)" />
+
+    <UpdateXml  XmlFilePath="$(ProjectDirectory)/build/Targets/Tools.props"
+                PathToElement="Project/PropertyGroup/dotnetSdkVersion"
+                NewElementValue="$(SDK_VERSION)" />
```
jaredpar: This needs to be removed once we merge the private branch as we jus condition away this check altogether.

init-tools.sh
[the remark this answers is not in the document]
dagood: In theory the working dir might not be `$__scriptpath` at this point though, and `--directory="$__scriptpath"` adds support for that, right?

patches/roslyn/0002-Add-solution-for-source-build.patch
```diff
@@ -0,0 +1,147 @@
+From c7dbba01272248423989a1816625dccb3e129b74 Mon Sep 17 00:00:00 2001
```
333fred: This patch is unnecessary, this commit of roslyn has Sourcebuild.sln.

repos/roslyn.proj
333fred: This can just be removed in this PR as it's based on the dev/sourcebuild branch.

---
In reply to: [180162495](https://github.com/dotnet/source-build/pull/404#discussion_r180162495) [](ancestors = 180162495)

333fred:
We will need to update the PR to pull in roslyn from dev15.7.x instead of dev/sourcebuild as we get that branch merged in.

patches/0001-Patch-Tools-for-sdk-pr-1997.patch
dseefeld: Yes.  I moved it.

init-tools.sh
dseefeld: @dagood and I discussed offline.  I will leave it in the init-tools.sh script since 1) it is very temporary and will go away on the next update of CLI version and 2) it can more easily participate in the Tools setup semaphore.

repos/roslyn.proj
dseefeld: Removed.  Thanks.

init-tools.sh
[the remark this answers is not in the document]
dseefeld: With help from @dagood , found that this was the issue with the git apply in CI.  Removed the --directory option.

patches/roslyn/0002-Add-solution-for-source-build.patch
dseefeld: Updated the sha to the commit that fixes SourceBuild.sln.

PR #423, opened 2018-04-13
Updated roslyn to consume the new builtin support for dotnet pack.

333fred:
cc @jaredpar. This gets us almost all of the way off of the patch files, with the only one remaining being the global.json patch.


333fred:
I don't have a linux box to test locally, so I was going to let CI test this. It may be blocked by https://github.com/NuGet/Home/issues/6829.

333fred:
@dseefeld looks like this should be good to go.

Issue #427, opened 2018-04-13
Enable cli crossgen step


dagood:
https://github.com/dotnet/source-build/pull/418 adds this to unblock the CLI build:

https://github.com/dotnet/source-build/blob/a184e0b053beeaa64ed5321f06f7eded3ad1c8ee/repos/cli.proj#L65-L71

I'm wondering if this is important to fix in 2.1-preview2 to improve our 2.1 RTM build performance. (I'm not familiar with the kinds of speedup we see, or if there's any direct requirement for crossgenning.)

/cc @dleeapho @dseefeld @peterhuene

PR #440, opened 2018-04-16
Enable CLI crossgen: fix dependency on Roslyn

dagood:
Fixes https://github.com/dotnet/source-build/issues/427

The package that CLI was depending on didn't match what we build. I found that https://github.com/dotnet/cli/commit/ddae0875cf1adaeb0575bffc34630a078d6b8cca#diff-acadc45977478b307ad6d53f23f6e17c works. I didn't look into timelines or reasoning--getting this PR in as soon as I have something that seems to work so that it can be checked in CI.


dagood:
CI failed--not surprising since my local verification was a few tweaks on an existing tarball build. Doing a fresh build locally to repro and work on.

dagood:
Closing: I think this is the wrong way to do it. The change I reverted to isn't recent, and I see now that we are skipping the Roslyn package that it's currently trying to consume. I'm able to source-build that Roslyn package with a few changes locally, and that seems much better. (Especially for the future, because we won't need a weird CLI reversion patch.)

Issue #448, opened 2018-04-17
Reduce CI turnaround time: look for low-hanging fruit


dagood:
CI turnaround time for the tarball legs in particular is more than 2 hours. A lot of this is because we've been adding a lot of product to the build, but we should try (briefly) finding some low-hanging fruit.

One thing that comes to mind is that CLI is printing out diag MSBuild output, which helps make our jenkins console output > 200 MB. I don't know if or how much this affects performance. (Doesn't seem to help, tried in https://github.com/dotnet/source-build/issues/448.)

/cc @dseefeld @crummel @dleeapho

kasper3:
or the more recent one: `tmpfs`  https://askubuntu.com/a/173294

https://www.jamescoyle.net/knowledge/951-the-difference-between-a-tmpfs-and-ramfs-ram-disk

dagood:
Interesting ideas, they don't seem to me like low-hanging right now, though.

For `ramfs`, our build expands quite a bit, and I'm not sure what the machine specs are. `tmpfs` looks like it at least wouldn't kill the machine if it doesn't fit. We could try building in `/tmp`, normally set up as a `tmpfs`. Not sure what the repercussions are for building outside the workspace we're given, or if there's some "real" way to do this. I heard a while back that RAM filesystems are less worthwhile recently with fast SSDs, which I'm pretty sure the CI agents use.

Parallelizing sounds nice, but we're pretty limited. At the repo level, we only have the amount of parallel work each repo can support with its toolset. Above that, building entire repos in parallel to each other seems likely to have weird races. (Even if we don't hit any at first, changing the build order or even just slightly different timing could cause one.) Maybe we could be more sure if we broke the build down into a bunch of isolated Docker container builds. Even then, our long pole is CoreCLR => CoreFX => Core-Setup => FSharp => CLI, and these are our longest individual builds anyway.

PR #449, opened 2018-04-17
(WIP) Reduce CLI build verbosity to minimal

dagood:
I'm curious how much this will reduce CI time (if significant at all). https://github.com/dotnet/source-build/issues/448.

It looks like `/v:detailed` makes CLI output `/v:diagnostic` output, not sure why. Maybe the nested MSBuild calls are all `/v:diagnostic`, and `/v:detailed` is enough to let their output through.


dleeapho:
Definitely worth a try.

dagood:
The new times look within what I'd expect from randomness, no noticeable change. Oh well.

Here are the "normal" tarball builds (hidden by the generated CI legs finishing and reporting status on this PR):
 * https://ci.dot.net/job/dotnet_source-build/job/dev_release_2.1/job/RHEL7.2_Tarball_Release_prtest/155/ Still timed out.
 * https://ci.dot.net/job/dotnet_source-build/job/dev_release_2.1/job/CentOS7.1_Tarball_Release_prtest/75/ Passed, but borderline at 1h 53m.

PR #457, opened 2018-04-19
Add smoke-test.sh to test built cli

dseefeld:


smoke-test.sh
```diff
@@ -0,0 +1,176 @@
...
+        --projectoutput)
+            projectOutput=true
+            ;;
+        --keepprojects)
+            keepProjects=true
+            ;;
+        --includeweb)
+            includeWeb=true
+            ;;
+        --clean)
```
crummel: I think this could be the default, I've tried to run smoke tests and ran into the existing directory way more times than I ran it without wanting wanting to delete the previous version.

smoke-test.sh
```diff
@@ -0,0 +1,176 @@
...
+    while :; do
+        if [ $# -le 0 ]; then
+            break
+        fi
+
+        echo "    running $1" | tee -a ../smoke-test.log
+
+        if [ "$1" == "new" ]; then
+            if [ "$projectOutput" == "true" ]; then
+                ${dotnetCmd} new $proj -lang $lang
```
crummel: I think this could `tee` to `smoke-test.log` just in case you want to go back and look at the output.  Same below.

smoke-test.sh
```diff
@@ -0,0 +1,176 @@
...
+            if [ "$projectOutput" == "true" ]; then
+                ${dotnetCmd} $1
+            else
+                ${dotnetCmd} $1 >> ../smoke-test.log 2>&1
+            fi
+        fi
+        if [ $? -eq 0 ]; then
+            echo "    $1 succeeded" >> ../smoke-test.log
+        else
+            echo "    $1 failed with exit code $?" >> tee -a ../smoke-test.log
```
crummel: Should be `| tee` instead of `>> tee`.

smoke-test.sh
dseefeld: Added an option to prompt for removal rather than an arg.

smoke-test.sh
dseefeld: fixed.

PR #458, opened 2018-04-19
Add smoke-test.sh to test built cli

dseefeld:


smoke-test.sh
```diff
@@ -0,0 +1,176 @@
...
+DOTNET_TARBALL_REL=$(ls ${OUTPUT_DIR}dotnet-sdk-${VERSION_PREFIX}*)
+DOTNET_TARBALL=$(readlink -e ${DOTNET_TARBALL_REL})
+
+projectOutput=false
+keepProjects=false
+dotnetDir=""
+workingDir="$(pwd)"
+includeWeb=false
+
+function usage {
```
dagood: Doesn't matter in bash I suppose, but in other places we have `()` always, and a few Bash Hackers Wiki pages suggest even using the portable POSIX `usage () { ... }`.

smoke-test.sh
```diff
@@ -0,0 +1,176 @@
...
+    while :; do
+        if [ $# -le 0 ]; then
+            break
+        fi
+
+        echo "    running $1" | tee -a ../smoke-test.log
+
+        if [ "$1" == "new" ]; then
+            if [ "$projectOutput" == "true" ]; then
+                ${dotnetCmd} new $proj -lang $lang
```
dagood: The `${dotnetCmd}`s here should have quotes, in case of spaces.

smoke-test.sh
```diff
@@ -0,0 +1,176 @@
...
+mkdir testing-smoke
+cd testing-smoke
+
+# Unzip dotnet if the dotnetDir is not specified
+if [ "$dotnetDir" == "" ]; then
+    mkdir builtCli
+    cd builtCli
+    tar xzf ${DOTNET_TARBALL}
+    cd ..
+    dotnetDir="../builtCli/"
```
dagood: The `..` is confusing in this context... convert to a full path instead?

smoke-test.sh
```diff
@@ -0,0 +1,176 @@
...
+
+mkdir testing-smoke
+cd testing-smoke
+
+# Unzip dotnet if the dotnetDir is not specified
+if [ "$dotnetDir" == "" ]; then
+    mkdir builtCli
+    cd builtCli
+    tar xzf ${DOTNET_TARBALL}
+    cd ..
```
dagood: I like subshells for environmental things you have to "reverse" like `cd`--makes sure you don't end up in an unexpected place if you abort the script.

smoke-test.sh
dseefeld: I changed this to define paths up front and extract to the defined path.

smoke-test.sh
```diff
@@ -0,0 +1,177 @@
...
+    doCommand C# mvc new restore run
+    doCommand C# webapi new restore run
+    doCommand C# razor new restore run
+
+    doCommand F# web new restore run
+    doCommand F# mvc new restore run
+    doCommand F# webapi new restore run
+fi
+
+rm -rf builtCli
```
dagood: Should this be `"$cliDir"` now?

smoke-test.sh
```diff
@@ -0,0 +1,177 @@
...
+VERSION_PREFIX=2.1
+OUTPUT_DIR=bin/x64/Release/
+DOTNET_TARBALL_REL=$(ls ${OUTPUT_DIR}dotnet-sdk-${VERSION_PREFIX}*)
+DOTNET_TARBALL=$(readlink -e ${DOTNET_TARBALL_REL})
+
+projectOutput=false
+keepProjects=false
+dotnetDir=""
+includeWeb=false
+SCRIPT_ROOT="$(pwd)"
```
dagood: Calling this `SCRIPT_ROOT` seems a bit misleading since it's just the working dir. It looks like `DOTNET_TARBALL_REL` assumes that this script is run with the working dir pointing at the root of a built repo, too--move `SCRIPT_ROOT` to the top, calculate actual script dir, and use it to find the tarball outputs?

PR #462, opened 2018-04-20
Add a 2.0.0 SDK to build FSharp for 2.1.0-preview2

dagood:
This 2.0.0 SDK is used only for F#, to fix the smoke test.

FYI @brettfo


dagood:
I hit a problem building F# in my local offline build:

```
/work/src/fsharp/src/buildfromsource.sh: line 6: dotnet: command not found
```

Probably something simple like I'm passing the directory in wrong or not copying something to the tarball correctly--looking into it.

brettfo:
buildfromsource.sh assumes the dotnet tool is on the path.

dagood:
Yeah, my problem was that the way I set up the tarball copies put the tools in the wrong folder. (One dir upwards from where they should be.) Created a new tarball and building it now.

crummel:
Online build worked for me, starting a tarball build now.

dagood:
My offline tarball build gave me an SDK that smoke tested successfully. 🎉

crummel:
Cool, if CI beats me this looks good to me.

build.sh
```diff
@@ -20,6 +20,12 @@ export DOTNET_MULTILEVEL_LOOKUP=0
 export NUGET_PACKAGES="$SCRIPT_ROOT/packages/"
 
 source "$SCRIPT_ROOT/init-tools.sh"
+(
+  echo "Also initializing 2.0.0 toolset to work around a build issue"
+  cd "$SCRIPT_ROOT/tools-2.0.0"
+  # Don't use 'source': the earlier init-tools created some vars that interfere with this one.
+  ./init-tools.sh
```
dagood: Not sourcing `init-tools.sh` means it can't exit when it sees the buildtools semaphore. (It uses `return`, not `exit`.) I'm not going to respin CI for this--it's only a perf issue, and only when repeating a build--but something I have in mind.

PR #463, opened 2018-04-20
Adding more repos that build in RC

dseefeld:
There are a handful of repos that build in RC.  Un-commenting them from known-good.proj.

PR #465, opened 2018-04-23
Fix clicommandlineparser build.

crummel:
- Remove obsolete NuGet.Config entry.
- Add clicommandlineparser back to known good.


repos/known-good.proj
```diff
@@ -20,7 +20,7 @@
     <RepositoryReference Include="xliff-tasks" />
 
     <!--RepositoryReference Include="msbuild" /-->
-    <!--RepositoryReference Include="clicommandlineparser" /-->
+    <RepositoryReference Include="clicommandlineparser" />
```
dagood: This should be uncommented in cli.proj too--I think your branch forked off before @dseefeld commented out the cli.proj refs in https://github.com/dotnet/source-build/pull/463.

crummel: Thanks, fixed.

PR #469, opened 2018-04-23
Add updater to update CLI version in global.json

dseefeld:
Add an updater to update sdk version in global.json and update projects that require it.


repos/roslyn-tools.proj
```diff
@@ -6,6 +6,7 @@
     <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
     <UsesRepoToolset>true</UsesRepoToolset>
     <BuiltToolPackageId>$(RepoToolsetPackageId)</BuiltToolPackageId>
+    <GlobalJsonFile>$(ProjectDirectory)/global.json</GlobalJsonFile>
```
dagood: For NuGet.Config we auto-detect the basic `$(ProjectDirectory)/NuGet.Config`, thought about doing the same here to minimize the proj files? Are there some repos with global.json that we don't need to update?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UpdateJson.cs
```diff
@@ -0,0 +1,53 @@
...
+        public override bool Execute()
+        {
+            JObject jsonObj = JObject.Parse(File.ReadAllText(JsonFilePath));
+
+            UpdateAttribute(jsonObj, PathToAttribute.Split('.'), NewAttributeValue);
+
+            File.WriteAllText(JsonFilePath, jsonObj.ToString());
+            return true;
+        }
+        private void UpdateAttribute(JToken jsonObj, string[] path, string newValue)
```
dagood: Missing newline above the method, and there's an unnecessary newline at the end of the method.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UpdateJson.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            string pathItem = path[0];
+            if (jsonObj[pathItem] == null) throw new ArgumentException(string.Format("Path item [{0}] not found in json file.", pathItem), "PathToAttribute");
+            if (path.Length == 1) 
+            {
+                jsonObj[pathItem] = newValue;
+                return;
+            }
+            UpdateAttribute(jsonObj[pathItem], path.Skip(1).ToArray(), newValue);
```
dagood: This seems fine, but I suggest checking out [`SelectToken`](https://www.newtonsoft.com/json/help/html/SelectToken.htm)--might be able to go down to even less code. (Wondered if there's anything built in for this (like xpath) and found that API today.)

repos/roslyn-tools.proj
dseefeld: There are other repos with global.json that don't need to be updated.  They don't have sdk.version, though, but I thought it better to be explicit in this case.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UpdateJson.cs
dseefeld: fixed.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UpdateJson.cs
dseefeld: I had started with SelectToken, but in the end, to update the target attribute, you need to select up to that attribute, and it didn't really save that much code.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UpdateJson.cs
```diff
@@ -0,0 +1,53 @@
...
+            UpdateAttribute(jsonObj, PathToAttribute.Split('.'), NewAttributeValue);
+
+            File.WriteAllText(JsonFilePath, jsonObj.ToString());
+            return true;
+        }
+
+        private void UpdateAttribute(JToken jsonObj, string[] path, string newValue)
+        {
+            string pathItem = path[0];
+            if (jsonObj[pathItem] == null) throw new ArgumentException(string.Format("Path item [{0}] not found in json file.", pathItem), "PathToAttribute");
```
dagood: A few style things: I usually expect `{ }` always (although some styles allow it for one-liners like this), `string.Format` should be `$""`, `"PathToAttribute"` should be `nameof(PathToAttribute)`.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UpdateJson.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using System.IO;
+using System.Linq;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using NuGet.Packaging;
+using NuGet.Packaging.Core;
```
dagood: Are the NuGet usings necessary?

repos/roslyn.proj
```diff
@@ -8,6 +8,7 @@
     <BuildCommand>$(DotnetToolCommand) build $(ProjectDirectory)/SourceBuild.sln /p:Configuration=$(Configuration) /p:OfficialBuild=true /p:BuildNumber=$(BuildNumber)</BuildCommand>
     <PackagesOutput>$(ProjectDirectory)/Binaries/packages</PackagesOutput>
     <RepoApiImplemented>false</RepoApiImplemented>
+    <GlobalJsonFile>$(ProjectDirectory)/global.json</GlobalJsonFile>
```
dagood: `ProjectDirectory` already has the `/`.

dagood:
I'll look into the `System.IO.FileNotFoundException: Could not load file or assembly 'Newtonsoft.Json, Version=10.0.0.0,` in the tarball build to move this along.

crummel:
I can repro locally.  I think this is the old issue where we need the version of Newtonsoft.Json we build before we build it, so it's removed from prebuilt.  I'm trying the build using Newtonsoft.Json 10.0.3, which is included in prebuilt, to see if that works.

dagood:
Yeah, I got one too when I added newtonsoft-json back in:

 1. The source-build tasks project depends on Newtonsoft.Json 9.0.1.
 2. In the first stage, the package is acquired, which sets the version it tries to load. I'm pretty sure it's actually loaded from the tooling CLI's sdk directory, sibling to to MSBuild.dll. It's also 9.0.1. (I still haven't been able to figure out how to get .NET Core to give me DLL loading diagnostics.)
 3. During the first stage, we make a source-built Newtonsoft.Json 9.0.1.
 4. The tarball build deletes the prebuilt Newtonsoft.Json 9.0.1 because we source-built the same version.
 5. Inside the tarball build, the task project can't find 9.0.1, and uses the prebuilt 10.0.3 instead. This sets the assembly version it depends on to 10.0.0.0.
 6. When running the task, it tries to find 10.0.0.0 in the SDK directory, but can't.

If that's right, it explains why my earlier repro attempt based on a partial build failed: it didn't include the newtonsoft-json build which causes the 9.0.3 prebuilt to be excluded. Forcing 9.0.3 to be included as a prebuilt sounds bad, since it would end up in the packages/ dir and probably leak into the CLI we assemble. Not sure if upgrading Newtonsoft.Json will work if it still tries to find the new version in the SDK dir.

`Reference`ing the Newtonsoft.Json.dll from the SDK dir seems to work. (I also had to reference the NuGet DLLs, since that PackageReference pulls in a Newtonsoft.Json package.) I haven't seen a straight DLL `Reference` in a while, but it does seem to express what we're trying to do better. No package versions to keep synced with the toolset.

```msbuild
<PropertyGroup>
  <SdkDllDir>$(ToolsDir)/dotnetcli/sdk/$(SDK_VERSION)/</SdkDllDir>
</PropertyGroup>
<ItemGroup>
  <NuGetDll Include="$(SdkDllDir)NuGet.*.dll" />
  <Reference Include="Newtonsoft.Json">
    <HintPath>$(SdkDllDir)Newtonsoft.Json.dll</HintPath>
  </Reference>
  <Reference Include="@(NuGetDll -> '%(FileName)')">
    <HintPath>$(SdkDllDir)%(Identity).dll</HintPath>
  </Reference>
</ItemGroup>
```

crummel:
That seems like a good fix to me.  Do you think we should do the same for the `Microsoft.Build` and `Microsoft.Build.Utilities.Core` references in the tasks project or are those not likely to ever match up?

dagood:
I hit problems changing over to include everything from the SDK, generally errors like `error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.`. It seems the package reference brings in some needed assemblies, and I'm not sure which. I'm leaving the `Microsoft.Build` references as `PackageReference`, clearing up the conflict, then I'll push to @dseefeld's branch.

dagood:
java.nio.channels.ClosedChannelException ([build](https://ci.dot.net/job/dotnet_source-build/job/dev_release_2.1/job/OSX10.12_Release_prtest/225/))

PR #478, opened 2018-04-26
Add cli-migrate, sdk to known-good

dagood:
Removes patches for sdk.

I suspect we'll have these cli-migrate patches back in, unfortunately: https://github.com/dotnet/source-build/tree/d2daebd13d8c40dacc35c4165b4b174ae6a1dde3/patches/cli-migrate. They still apply as they were. Submitting in a separate commit to get CI runs with+without. @tmat, could you look at adding them to the main branch?


tmat:
This should do: https://github.com/dotnet/cli-migrate/pull/73

dagood:
I don't see how either--I'll send you the binlog from a repro offline.

dagood:
Look ok to merge if the last CI leg passes? This gets sdk in patchless and cli-migrate in with patches for now.

PR #481, opened 2018-04-26
Update NuGet to include patch commit and remove patches.

crummel:


crummel:
@dagood Is https://github.com/dotnet/source-build/blob/275e6615655bc36401aadcc67320db6231f8d3ca/repos/nuget-client.proj#L11 already correct to enable nuget-client auto-update?

PR #490, opened 2018-04-30
Force corefx to build identity and runtime packages

weshaggard:
@dseefeld this should fix the issue you are seeing with corefx not publishing packages.

Looks like when we tried to fix the corefx/coreclr package builds we based it off the wrong property.

https://github.com/dotnet/corefx/pull/28607
https://github.com/dotnet/coreclr/pull/17323

In those PR's we based it on DotNetBuildOffline but that is only set in the tarball not the source-build. We should instead update corefx and coreclr to use DotNetBuildFromSource for these conditions. I will submit a PR to change them. 

Note that I'm only workaround the corefx in this PR because we don't actually use the identity packages from coreclr in our other builds.

### DIFF
--- a/repos/corefx.proj
+++ b/repos/corefx.proj
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
   <PropertyGroup>
     <BuildArguments>-$(Configuration) -buildArch=$(Platform) -portable=$(PortableBuild) -BuildTests=false</BuildArguments>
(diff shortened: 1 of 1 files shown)


weshaggard:
I've submitted PR's to fix the release/2.1 branches:
https://github.com/dotnet/coreclr/pull/17846
https://github.com/dotnet/corefx/pull/29418

dagood:
@weshaggard @dseefeld Can you give some more detail about what this fixes? Where are the CoreFX identity packages used?

weshaggard:
test Debian8.4 Release
test RHEL7.2 Tarball Release

weshaggard:
@dotnet-bot test Debian8.4 Release
@dotnet-bot test RHEL7.2 Tarball Release

PR #498, opened 2018-05-02
Remove CLI patch #5 to use CoreSetupRid and set the global CLI RID instead.

crummel:
The `--runtime-id` property in CLI sets the MSBuild `Rid` property so this patch seems to be a roundabout way of doing the same thing.  This does change the output directory for CLI so I've also fixed that.  I am currently running Debian and offline RHEL tests on this change.


crummel:
cc @weshaggard @livarcocc

dir.props
```diff
@@ -77,11 +77,6 @@
 
   <Import Project="$(TargetInfoProps)" Condition="$(GeneratingStaticPropertiesFile) != 'true' AND Exists('$(TargetInfoProps)')" />
 
-  <PropertyGroup>
-    <!-- Always build portable runtime on OSX to match the CLI repo's filename expectations. https://github.com/dotnet/source-build/issues/438 -->
-    <PortableBuild Condition="'$(TargetOS)' == 'OSX'">true</PortableBuild>
-  </PropertyGroup>
```
dagood: I agree, it makes sense to be able to remove this now.

repos/cli.proj
```diff
@@ -2,7 +2,8 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
   <PropertyGroup>
-    <BuildCommandArgs>--stage0 $(DotNetCliToolDir)</BuildCommandArgs>
+    <RuntimeArg Condition="'$(TargetOS)' != 'OSX'">--runtime-id $(TargetRid)</RuntimeArg>
```
weshaggard: Why is OSX special? 

I suspect windows maybe special in the same way.

crummel: I believe you are correct.  The relevant code sets the `HostMonikerRid` [here](https://github.com/dotnet/cli/blob/4db4d9b568d575b379e2b9174d1aeba093695c07/build/Branding.props#L20).  Then `CoreSetupRid` gets set to `HostMonikerRid` [here](https://github.com/dotnet/cli/blob/4db4d9b568d575b379e2b9174d1aeba093695c07/build/BundledRuntimes.props#L4).  Then `SharedFrameworkRid` gets set to `CoreSetupRid` [here](https://github.com/dotnet/cli/blob/4db4d9b568d575b379e2b9174d1aeba093695c07/build/BundledRuntimes.props#L27).  Then `SharedFrameworkRid` gets used in filenames like [this](https://github.com/dotnet/cli/blob/4db4d9b568d575b379e2b9174d1aeba093695c07/build/BundledRuntimes.props#L30).  I think restoring the forced Portable=true and special casing our [CLI path](https://github.com/crummel/dotnet_source-build/blob/b97922f0560dff72145295c1466ee9fa15632f3b/repos/cli.proj#L32) for Windows and Mac vs Linux is probably the best move for now.

cc @livarcocc

Issue #10, opened 2016-08-23
Do not use NI/R2R images in prebuilt tools


ellismg:
Prebuilt tools (cli, roslyn's toolchain, etc) shouldn't use NI/R2R images.  Instead, they should be IL only and after boot strapping coreclr, we can crossgen them as needed.

This is needed because R2R images are not portable across either architectures or OSes (e.g. Windows, Linux, OSX or other *NIXes).

Obviously, NI images are even more fragile.

PR #499, opened 2018-05-02
Remove CLI patch #10 - Remove conflicting Microsoft.CodeAnalysis.CSharp pkg from ref-assemblies

dseefeld:
Microsoft.CodeAnalysis.CSharp.2.1.0 is built as a reference assembly in the offline build.  CLI uses it and attempts to copy the asssembly from the package, which doesn't exist.  This causes crossgen to fail with the error:
`Could not load file or assembly 'Microsoft.CodeAnalysis.CSharp, Version=2.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.`

The fix is to remove the package from reference assemblies for the offline build.


dagood:
I suspect this will produce a good CLI, but it will mean we're shipping Roslyn assemblies from prebuilt.

PR #509, opened 2018-05-04
Update msbuild version info to use property expected by repotoolset

dseefeld:
msbuild was updated to use a newer version of repotoolset.  It now expects VersionPrefix instead of VersionBase.


dseefeld:
Built source-build and smoke-tests passed.  Building the tarball now.

PR #513, opened 2018-05-04
Disable PGO optimization for now as it requires a higher version of clang.

crummel:


repos/coreclr.proj
```diff
@@ -2,7 +2,7 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
   <PropertyGroup>
-    <BuildArguments>$(Platform) $(Configuration) skiptests -PortableBuild=$(PortableBuild)</BuildArguments>
+    <BuildArguments>$(Platform) $(Configuration) skiptests -PortableBuild=$(PortableBuild) -nopgooptimize</BuildArguments>
```
dagood: Does it make sense to make this toggleable with a property so that if a distro uses a good enough `clang` they can easily turn it back on?

crummel: Makes sense, added.

PR #521, opened 2018-05-08
Fix smoke test for OSX

dseefeld:
Update sed command, since OSX expects an extension on the -i command.
Remove usage of readlink, since path is already absolute.

### DIFF
--- a/smoke-test.sh
+++ b/smoke-test.sh
@@ -4,10 +4,7 @@ set -e
 SCRIPT_ROOT="$(cd -P "$( dirname "$0" )" && pwd)"
 TARBALL_PREFIX=dotnet-sdk-
 VERSION_PREFIX=2.1
-OUTPUT_DIR="$SCRIPT_ROOT/bin/x64/Release/"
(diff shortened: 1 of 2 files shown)


smoke-test.sh
```diff
@@ -182,6 +179,9 @@ cd "$testingDir"
 
 # Unzip dotnet if the dotnetDir is not specified
 if [ "$dotnetDir" == "" ]; then
+    OUTPUT_DIR="$SCRIPT_ROOT/bin/x64/Release/"
+    DOTNET_TARBALL=$(ls ${OUTPUT_DIR}dotnet-sdk-${VERSION_PREFIX}*)
```
dagood: While you're moving this, might as well add quotes around the `ls` arg.

crummel:
@dotnet-bot test OSX10.12 Release please (timeout)

smoke-test.sh
dseefeld: Fixed.

PR #524, opened 2018-05-09
Remove cli patches

dagood:
Move to latest ProdCon 2.1-rtm build that includes fixes necessary to remove patches.

Switch to ProdCon API to pass package versions props file, rather than source-build API.

Use existing environment variable rather than source-build API to override package sources.


repos/cli.proj
```diff
@@ -50,7 +54,7 @@
         Microsoft.AspNetCore.All;
         Microsoft.AspNetCore.App;
         Microsoft.NET.Sdk.Razor"
-      Version="2.1.0-rc1-30682" />
+      Version="2.1.0-rc1-final" />
```
dagood: I'm trying changing this to `2.1.0-rtm-30762` locally and adding the dotnet-core MyGet feed to the smoke test to see if we can get the aspnetcore experience going.

I think it's worth merging this PR now as an incremental step whether or not that works first try, since CI is green now.

PR #526, opened 2018-05-09
Remove CoreFX BuildingAnOfficialBuildLeg=false

dagood:
We now have the fix that lets us remove it in the CoreFX submodule. Added in #490, fixed in https://github.com/dotnet/corefx/pull/28607.


weshaggard:
Thanks for cleaning this up.

PR #529, opened 2018-05-09
Remove msbuild patch and update sha to latest v15.7_source-build branch

dseefeld:
Patch for msbuild has been merged into msbuild's v15.7_sourcebuild branch.  Remove patch and move to new commit.

cc @cdmihai


dagood:
The source-build branch was merged into `master`, too, FWIW: https://github.com/Microsoft/msbuild/pull/3288.

I was going to suggest turning on ProdCon auto-update for MSBuild because of that, but it turns out MSBuild is also no longer being built in ProdCon, like Roslyn.

PR #539, opened 2018-05-11
Add web to smoke-test, with and without dev certs

dagood:
This gets the aspnetcore smoke-tests working, both with and without dev certs. I changed the 20 second delay to check every second up to 20 to see if the project's ready--this speeds it up a bit.

The challenge was figuring out why SIGINT wasn't working as expected. It turns out to be new behavior in 2.1 because https://github.com/dotnet/coreclr/pull/15393 makes .NET Core preserve `SIG_IGN` (signal ignore) when it's set on SIGINT. ~~I added `set -m` to avoid `SIG_IGN` being set when our `dotnet run` is backgrounded. (https://unix.stackexchange.com/a/356480)~~ I changed it to SIGTERM per @tmds's suggestion.

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -13,8 +13,13 @@ def addBuildStepsAndSetMachineAffinity(def job, String os, String configuration)
       }
       else {
         shell("git submodule update --init --recursive");
-        shell("./build.sh /p:Configuration=${configuration} ${loggingOptions}")
-        shell("./smoke-test.sh --minimal --projectOutput --configuration ${configuration}")
+        shell("./build.sh /p:Configuration=${configuration} ${loggingOptions}");
+        smokeTestExcludes = "";
+        if (os == "Fedora24" || os == "OSX10.12") {
+          // Dev certs doesn't seem to work in these platforms. https://github.com/dotnet/source-build/issues/560
+          smokeTestExcludes += " --excludeWebHttpsTests";
+        }
+        shell("./smoke-test.sh --minimal --projectOutput --configuration ${configuration} ${smokeTestExcludes}");
       }
     };
   };
--- a/smoke-test.sh
+++ b/smoke-test.sh
@@ -4,30 +4,40 @@ set -euo pipefail
 SCRIPT_ROOT="$(cd -P "$( dirname "$0" )" && pwd)"
 TARBALL_PREFIX=dotnet-sdk-
 VERSION_PREFIX=2.1
+DEV_CERTS_VERSION_DEFAULT=2.1.0
 
 projectOutput=false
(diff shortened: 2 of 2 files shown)


dagood:
`java.nio.channels.ClosedChannelException`

dagood:
@dotnet-bot test OSX10.12 Release (disconnected)

@dotnet-bot test Fedora24 Release
[Fedora24 error](https://ci.dot.net/job/dotnet_source-build/job/dev_release_2.1/job/Fedora24_Release_prtest/302/) might be legitimate... but without the smoke-test logs, can't tell what happened. Retrying in case it helps, and I'll grab a machine to repro on.

tmds:
@dagood fyi, changing to SIGTERM (which is the default for kill-commands) should also make things work (without requiring `set -m`).

dagood:
In a repro Fedora24 machine, I get this error message setting up the dev cert:

```
+ /mnt/resource/workspace/Fedora24_Release_prtest/testing-smoke/home/.dotnet/tools/dotnet-dev-certs https
There was an error saving the HTTPS developer certificate to the current user personal certificate store.
```

That error's written by this code: https://github.com/aspnet/Common/blob/96eefd7eb1de752b8c569f820a796969f1f60ec4/shared/Microsoft.AspNetCore.Certificates.Generation.Sources/CertificateManager.cs#L669. Adding `-v` for verbose doesn't give any more detail. Would be nice to get the exception output.

The command works fine using a Microsoft build of the RTM CLI (which bundles the tool). Smoke test also works on my Fedora 28 source-builds (build-source-tarball and building inside the tarball).

For this PR, I'm going to make Fedora 24 skip this test category and file an issue to follow up. I'm also going to write out the logs when dev certs generation fails so I can see if the same thing's going on for OSX. (I don't have my OSX machine at the moment to try to repro.)

dagood:
@tmds Interesting, I was worried that would change the escape code or make the aspnet server respond differently, but it doesn't seem to. I must have been trying a different one. I'll switch to that, thanks!

dagood:
@dotnet-bot test ci please

dagood:
@dotnet-bot test Ubuntu16.04 Release
@dotnet-bot test Fedora24 Release

dagood:
Ubuntu16.04 still works with https smoke testing with the CI changes: https://ci.dot.net/job/dotnet_source-build/job/dev_release_2.1/job/GenPRTest/job/Ubuntu16.04_Release_prtest/1/

Still waiting for Fedora24 and OSX legs to complete, but the configurations do have `--excludeWebHttpsTests` in the Jenkins UI.

PR #545, opened 2018-05-15
Add VersionStamp to CoreFX and core-setup builds.

crummel:
This will make produced versions only use "-rc1" suffixes instead of "-rc1-buildnumber", which matches up with the official RC1 release.

### DIFF
--- a/repos/dir.props
+++ b/repos/dir.props
@@ -46,6 +46,11 @@
     <EnvironmentVariables Include="DotNetBuildOffline=true" Condition="'$(OfflineBuild)' == 'true'" />
     <EnvironmentVariables Include="DotNetCoreSdkDir=$(DotNetCliToolDir)" />
     <EnvironmentVariables Include="DotNetBuildToolsDir=$(ProjectDir)Tools/" />
+    <!-- Always build source-build as stabilized.  We don't publish packages so this
+         won't cause conflicts on package feeds, and it will make us match up for releases.  -->
+    <EnvironmentVariables Include="StabilizePackageVersion=true" />
+    <!-- Update the line below when branch quality changes -->
+    <EnvironmentVariables Include="PackageVersionStamp=rc1" />
   </ItemGroup>
 
 </Project>
--- a/repos/dir.props
+++ b/repos/dir.props
@@ -46,11 +46,8 @@
     <EnvironmentVariables Include="DotNetBuildOffline=true" Condition="'$(OfflineBuild)' == 'true'" />
     <EnvironmentVariables Include="DotNetCoreSdkDir=$(DotNetCliToolDir)" />
     <EnvironmentVariables Include="DotNetBuildToolsDir=$(ProjectDir)Tools/" />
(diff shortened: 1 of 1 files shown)


crummel:
@dotnet-bot test this please

repos/core-setup.proj
```diff
@@ -21,6 +21,7 @@
 
   <ItemGroup>
     <EnvironmentVariables Include="OfficialBuildId=$(OfficialBuildId)" />
+    <EnvironmentVariables Include="PackageVersionStamp=rc1" />
```
dseefeld: Is this enough?  Do you have to do something with StabilizePackageVersion as well?

crummel: Yes, I think this is closer to what prodcon does so probably better to do it this way.

repos/dir.props
```diff
@@ -46,6 +46,11 @@
     <EnvironmentVariables Include="DotNetBuildOffline=true" Condition="'$(OfflineBuild)' == 'true'" />
     <EnvironmentVariables Include="DotNetCoreSdkDir=$(DotNetCliToolDir)" />
     <EnvironmentVariables Include="DotNetBuildToolsDir=$(ProjectDir)Tools/" />
+    <!-- Always build source-build as stabilized.  We don't publish packages so this
+         won't cause conflicts on package feeds, and it will make us match up for releases.  -->
```
weshaggard: Do we really want to hard-code these values as true/rc1? I think we want to have them be parameterized somehow. Also I think I would prefer they get passed as part of the command line instead of the env variables but I'd say the same for most of these properties so I can live with it for now.

repos/core-setup.proj
weshaggard: Prodcon actually sets the PB_IsStable and PB_VersionStamp as env variables/build parameters and then the repo's map them to what they use. See https://github.com/dotnet/corefx/blob/master/buildpipeline/pipeline.json#L23 for example. While the core-repos are pretty consistent the others likely don't use those same properties.

repos/dir.props
crummel: Added parameters for these.  I looked at doing this as part of the build.sh script but we'd need to have some way to pass settings from the non-tarball build to the tarball build, and later when we fix incremental build we'd have to do something similar to keep these the same between rebuilds.

dagood:
Does this mean you have to run this sequence of commands to get a build?

```
./build.sh /p:ArchiveDownloadedPackages=true /p:VersionStamp=rc1
./build-source-tarball.sh ../tarball --skip-build
cd ../tarball
./build.sh /p:VersionStamp=rc1
```

(Calling build separately from build-source-tarball because there's no way to pass in params to the inner build command.)

I'm not a fan of how complicated this got without defaults.

crummel:
Yes, that's correct.  I can add a default for this.

weshaggard:
Yeah I just commented in the other PR as well I'm fine with having defaults as properties that could be overridden.

PR #556, opened 2018-05-16
Use stable RTM Microsoft.AspNetCore packages

dagood:
Move from a dev build version to the version that will be released on the nuget gallery. Add ProdCon feed for smoke testing.

Adding the ProdCon feed might actually let us do online tests early, too, since a stable-versioned Microsoft.NETCore.App is also available there.


dagood:
`Unable to find package Microsoft.NET.Sdk.Razor with version (>= 2.1.0)`

I forgot this is restored during the build itself rather than during smoke-testing. The stable version isn't available on dotnet-core MyGet/Central feeds. The obvious alternatives are:
 1. Leave `Microsoft.NET.Sdk.Razor` on an `rtm-...` version that's similar enough (hopefully identical) to the stable-versioned package. Only update the other two aspnetcore packages to stable.
 2. Point the CLI build at the ProdCon blob feed like I did in the smoke-test nuget.config. This way, we'll pull down the exact ProdCon package.

dagood:
@crummel also mentioned another option:

 3. Create a dummy package `Microsoft.NET.Sdk.Razor 2.1.0`.

This depends on if we need the functionality--but we probably do want it.

dagood:
Went with 1. for a first attempt. I compared package contents vs. 2.1.0 and the only differences are nuget metadata and the version in the nuspec.

dagood:
Option 1 fails in CI trying to restore 2.1.0 `Microsoft.DotNet.Web.ItemTemplates`, 
`Microsoft.DotNet.Web.ProjectTemplates.2.1`, and
`Microsoft.DotNet.Web.Spa.ProjectTemplates`. CLI uses the aspnetcore App package version to determine the versions of these it grabs, so we can't override without a patch. So I'm going to move on to option 2.

repos/cli.proj
```diff
@@ -41,7 +41,13 @@
     <RepoApiImplemented>false</RepoApiImplemented>
     <SourceOverrideRepoApiImplemented>true</SourceOverrideRepoApiImplemented>
 
+    <!-- Temporarily disable auto-update: patch consistently conflicts with dependency changes. -->
+    <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
+
+    <ProdConBlobFeedUrl>$([System.IO.File]::ReadAllText('$(ProdConFeedPath)').Trim())</ProdConBlobFeedUrl>
+
     <EnvironmentExternalRestoreSources>$(SourceBuiltPackagesPath)</EnvironmentExternalRestoreSources>
+    <EnvironmentExternalRestoreSources>$(EnvironmentExternalRestoreSources)%3B$(ProdConBlobFeedUrl)</EnvironmentExternalRestoreSources>
```
dagood: I skipped ahead to a CLI-only build to see if this works: it doesn't. MSBuild changes the `https://` to `https:/`. I'm trying the usual escaping with `%2F` and `%252F` which don't seem to change the result, I think because the string leaves the MSBuild context and comes back into CLI as a property through the environment variable.

dagood: We'll need a fix/patch in CLI to do this properly, I'm pretty sure. `%25252F` gets nice escaped `/`s in there, but CLI doesn't have an Unescape call to convert them into `/`s. There's also as far as I can tell no way to manually hook into the build to modify the nuget.config after CLI generates it.

In the meantime, I think I can do a "soft patch"/updater for CLI's `NugetConfigFile.targets` to add a line to the source code. I'm going to try that.

repos/cli.proj
```diff
@@ -39,6 +39,9 @@
     <RepoApiImplemented>false</RepoApiImplemented>
     <SourceOverrideRepoApiImplemented>true</SourceOverrideRepoApiImplemented>
 
+    <!-- Temporarily disable auto-update: patch consistently conflicts with dependency changes. -->
```
crummel: Do we need to re-disable this?

dagood:
Newest commit rebases on latest `dev/release/2.1` and injects the prodcon url by modifying `NugetConfigFile.targets` directly to add the item. I also tried out the smoke-test change on the tooling CLI and it made the expected nuget.config change to add the prodcon feed.

repos/cli.proj
dagood: Oh, this was unintentional, must have messed up my rebase flow. Fixing.

dagood:
This uses the 20180515-06 ProdCon feed, but 20180515-07 is the new ProdCon build. I'm submitting a PR to update, but this would only affect the result if the prebuilt SDK packages changed between these builds, which seems unlikely.

crummel:
Smoke test passed for me as well on the clean tarball build, RHEL7.4.

PR #557, opened 2018-05-16
Update CLI for new prodcon build.

crummel:
New prodcon build hasn't come out yet but this is the CLI hash that it will be (https://github.com/dotnet/cli/commit/32f29b6eb9fd4088625b1dea76000f0211c32305).


crummel:
Looks like CLI never got auto-update re-enabled after we removed patches so we were behind.  Added a commit to re-enable it.

PR #562, opened 2018-05-17
Update ProdCon feed to 20180515-07

dagood:
Updates ProdCon feed to latest from `2.1-rtm`. This would only affect the result if the prebuilt SDK packages changed between these builds, which seems unlikely.

### DIFF
--- a/ProdConFeed.txt
+++ b/ProdConFeed.txt
@@ -1,1 +1,1 @@
-https://dotnetfeed.blob.core.windows.net/orchestrated-release-2-1/20180515-06/final/index.json
+https://dotnetfeed.blob.core.windows.net/orchestrated-release-2-1/20180515-07/final/index.json


dagood:
We only need this for RTM if there's a difference in the prebuilts we pull from the ProdCon feed. I believe that's only the prebuilt SDK packages from aspnetcore.

I've started a F28 run on this PR in case we want to take it.

dagood:
java.nio.channels.ClosedChannelException

dagood:
Before/after this PR, all prebuilt packages are identical except:

```
microsoft.dotnet.web.itemtemplates.2.1.0
microsoft.dotnet.web.projecttemplates.2.1.0
microsoft.dotnet.web.spa.projecttemplates.2.1.0
microsoft.net.sdk.razor.2.1.0
```

Both sets of these are identical except for package metadata and the changes from https://github.com/aspnet/templating/pull/521: fixing jQuery script tag `integrity` attributes throughout the project templates. It seems like we should wait for this change.

PR #571, opened 2018-05-21
Merge dev/release/2.1 into master

safern:
cc: @weshaggard @dagood @crummel @dleeapho


dagood:
Comparing dev/release/2.1 to the merge commit, my git client tells me:

```diff
diff --git a/clean.sh b/clean.sh
old mode 100755
new mode 100644
```

Can you make sure the script keeps the right permissions?

safern:
Fixed :)

dagood:
I just merged a few PRs into dev/release/2.1 that were waiting for RTM. I think this PR is mergeable (make sure not to squash), then I'll submit more for the new commits.

crummel:
Could you please update the ReadMe to match netci.groovy?  We're not building ARM right now and we've added the tarball legs, I think you'll get those but the ReadMe doesn't look like it changed.

repos/dir.props
```diff
@@ -18,6 +31,16 @@
...
   </PropertyGroup>
 
+  <PropertyGroup>
+    <OrchestratedManifestBuildName Condition="'$(OrchestratedManifestBuildName)' == ''">$(RepositoryName)</OrchestratedManifestBuildName>
+  </PropertyGroup>
+
+  <!-- for RTM only, default to stabilized builds because that's what matches the official builds -->
+  <!-- once https://github.com/dotnet/source-build/issues/550 is fixed we will be able to determine this automatically -->
+  <PropertyGroup>
+    <UseStableVersions Condition="'$(UseStableVersions)' == ''">true</UseStableVersions>
```
crummel: Looks like you didn't get the `DropSuffix` change in CLI but did get this change - I think we want to be consistent in stable vs. unstable.  I think master makes sense to be unstable by default.

safern: I can update the PR to get the DropSuffix change in. 

>  I think master makes sense to be unstable by default.

So that would mean setting UseStableVersions to false by default and not getting the DropSuffix change in cli.proj right?

crummel: Yes, that's right - @dagood does master unstable make sense to you?

dagood: It sounds fine, I think the important thing is matching ProdCon though, or even being more stable than ProdCon to make sure we won't be broken when ProdCon stabilizes.

safern: So I updated UseStableVersions to false according to this.

weshaggard: Yeah I think it makes sense to produce unstable builds especially as we start to take additional changes.

safern:
@dotnet-bot test CentOS7.1 Release please

PR #573, opened 2018-05-23
Fix typo 2.0 -> 2.1

kasper3:
Something that i missed in https://github.com/dotnet/source-build/pull/377.


kasper3:
@janvorli, can you please take a look?

PR #583, opened 2018-05-31
Move msbuild repo from private dev branch to master

dseefeld:
This gets source-build onto the most current branch for msbuild.


crummel:
I think you'll want to update https://github.com/dotnet/source-build/blob/ac7159ec5f61d1ebadb454471d5f6190a4ebbc3b/repos/msbuild.proj#L18 as well.

Issue #534, opened 2018-05-10
Add updater to update version info from prodcon build manifest


dseefeld:

dagood:
To clarify a bit, this tracks bringing this kind of info from the manifest:

```
<Build Name="core-setup" BuildId="20180515-03" ...
```

to

https://github.com/dotnet/source-build/blob/c627c32ee8818a88d84713a2cff5c14272ea257b/repos/core-setup.proj#L9

dagood:
The functionality chunk I'm adding for the build ID also lets us update the ProdCon feed URL and update package versions from the manifest.

dagood:
Moving this back to backlog for now until I'm available to work on it. This updater won't be useful until we start getting ProdCon builds for the next release (whether from `master` or `release/2.1`) and it's just automation of a task that can be done manually so I think it can wait.

dagood:
Completed: new updaters added to `master` #602 and `release/2.1` #605.

Issue #592, opened 2018-06-07
Enable net46 builds in source-build


crummel:
End-to-end, this enables VS Code usage with source-build .NET Core (https://github.com/dotnet/source-build/issues/125 -> https://github.com/dotnet/sdk/issues/2249)

crummel:
Related: https://github.com/dotnet/sdk/pull/2309, https://github.com/dotnet/source-build/pull/588.

dleeapho:
@crummel Please test the E2E scenario with VS Code once PRs are merged.

dagood:
Merged in https://github.com/dotnet/source-build/pull/598. I'm going to try validating VS Code/Omnisharp with a local build I'm building of https://github.com/dotnet/source-build/pull/610.

crummel:
Confirmed, I also get a working VS Code.  `dotnet --info` output:
```
.NET Core SDK (reflecting any global.json):
 Version:   2.1.301
 Commit:    59524873d6

Runtime Environment:
 OS Name:     rhel
 OS Version:  7
 OS Platform: Linux
 RID:         rhel.7-x64
 Base Path:   /home/chris/dotnet/sdk/2.1.301/

Host (useful for support):
  Version: 2.1.1
  Commit:  N/A

.NET Core SDKs installed:
  2.1.301 [/home/chris/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.NETCore.App 2.1.1 [/home/chris/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET Core runtimes or SDKs:
  https://aka.ms/dotnet-download
```

PR #602, opened 2018-06-08
Add additional ProdCon updaters pointed at recent 2.1.1 build

dagood:
Upgrades buildtools to get new updaters implemented in https://github.com/dotnet/buildtools/pull/2047 and configures them, pointing at the recent 2.1.1 build.

I'm planning on porting the buildtools change from master to 2.1 once this gets merged, then porting this PR to source-build 2.1.

Ideally I'd use a master manifest rather than 2.1.1, but we don't have one. The source-build `master` appears to still be on 2.1.0 for now so this is still an upgrade.

#534


repos/dir.targets
```diff
@@ -371,6 +371,7 @@
 
   <Target Name="GetProjectDirectory" Outputs="$(ProjectDirectory)" />
   <Target Name="GetOrchestratedManifestBuildName" Outputs="$(OrchestratedManifestBuildName)" />
+  <Target Name="GetOfficialBuildId" Outputs="OfficialBuildId" />
```
eerhardt: Does this need to be `$(OfficialBuildId)` ?

dagood:
In this PR I updated buildtools to latest `master` to get the new updaters, but this is incompatible with the 2.1.1 CoreFX commit. I reverted the BuildTools update to see if the submodule updates go through CI other than that.

Once https://github.com/dotnet/buildtools/pull/2063 is through, we'll have a 2.1 BuildTools that includes the updaters.

We're also at an awkward place where https://github.com/dotnet/source-build/pull/589 will update the Core repos to `master` soon. There aren't any ProdCon `master` builds to point at, so it will become impossible to meaningfully test auto-update config changes in source-build `master`.

dagood:
@crummel looks like I ran into why we need net46 builds? (https://github.com/dotnet/source-build/issues/592)

```
/mnt/j/workspace/dotnet_source-build/master/Ubuntu16.04_Release_prtest/src/sdk/src/Tasks/Microsoft.NET.Build.Extensions.Tasks/Microsoft.NET.Build.Extensions.Tasks.csproj :
error NU1202: Package Microsoft.Build.Framework 15.7.179 is not compatible with net46 (.NETFramework,Version=v4.6).
Package Microsoft.Build.Framework 15.7.179 supports: netstandard2.0 (.NETStandard,Version=v2.0) 
[/mnt/j/workspace/dotnet_source-build/master/Ubuntu16.04_Release_prtest/src/sdk/sdk.sln] [/mnt/j/workspace/dotnet_source-build/master/Ubuntu16.04_Release_prtest/repos/sdk.proj]
```

I'll revert the upgrade entirely for now, and just get the configuration in.

PR #622, opened 2018-06-29
Bin place produced runtime by core-setup in bin folder

safern:
cc: @dagood @weshaggard

### DIFF
--- a/Documentation/README.md
+++ b/Documentation/README.md
@@ -129,7 +129,7 @@ Once the build is done, we will have a .NET Runtime containing the `Hashtable.PN
 
 ### How to use the produced Runtime
 
-The produced runtime will be under `bin/obj/{architecture}/{configuration}/blob-feed/packages`.
+The produced runtime will be under `bin/{architecture}/{configuration}/runtime`.
 
 Now create a new .NET Core project:
 ```console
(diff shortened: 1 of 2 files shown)


Documentation/README.md
```diff
@@ -145,7 +145,7 @@ In your new project add a new NuGet.config to point to the Microsoft.NETCore.App
   <!-- The command-line doesn't need it, but the IDE does.                    -->
   <packageSources>
     <clear/>
-    <add key="local feed" value="C:\repos\source-build\bin\obj\x64\Release\blob-feed\packages" />
+    <add key="local feed" value="E:\repos\source-build\bin\x64\Release\runtime" />
```
dagood: Doesn't the core-setup.proj change only copy the installers/tarballs (assets) to this location, not the nupkgs? I'd think the packages dir would still be the right nuget source. The earlier changed line in this file might also be better if it described both places.

repos/core-setup.proj
```diff
@@ -37,4 +37,14 @@
                               DestinationFolderTemplate="$(LocalBlobStorageRoot)Runtime/{version}/" />
   </Target>
 
+  <Target Name="CopyBinariesToBinFolder" AfterTargets="PublishBinaries">
+    <ItemGroup>
+      <BinariesToCopy Include="$(SourceBuiltAssetsDir)*.*" />
```
weshaggard: What is currently under this directory?

safern:
I've updated this to copy the nupkgs for the runtime so that the packages can be consumed by using the bin folder in `NuGet.config` and forget about the obj folder.

PR #631, opened 2018-07-08
disable crossgen on FreeBSD for now

wfurt:
cross-genning is failing now and needs to be investigated further.
 
This may be temporary or permanent change.

### DIFF
--- a/repos/core-setup.proj
+++ b/repos/core-setup.proj
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <BuildArguments>-ConfigurationGroup=$(Configuration) -PortableBuild=$(PortableBuild) -SkipTests=true </BuildArguments>
     <BuildArguments Condition="$(Platform.Contains('arm'))">$(BuildArguments) -TargetArchitecture=$(Platform) -DisableCrossgen=true -CrossBuild=true</BuildArguments>
+    <BuildArguments Condition="'$(TargetOS)' == 'FreeBSD'">$(BuildArguments) -DisableCrossgen=true</BuildArguments>
     <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) $(BuildArguments) -- /p:BuildDebPackage=false /p:BuildAllPackages=true</BuildCommand>
     <BuildCommand Condition="$(Platform.Contains('arm'))">$(ArmEnvironmentVariables) $(BuildCommand)</BuildCommand>
     <OfficialBuildId>20180509-04</OfficialBuildId>


repos/core-setup.proj
```diff
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <BuildArguments>-ConfigurationGroup=$(Configuration) -PortableBuild=$(PortableBuild) -SkipTests=true </BuildArguments>
     <BuildArguments Condition="$(Platform.Contains('arm'))">$(BuildArguments) -TargetArchitecture=$(Platform) -DisableCrossgen=true -CrossBuild=true</BuildArguments>
+    <BuildArguments Condition="'$(TargetOS)' == 'FreeBSD'">$(BuildArguments) -DisableCrossgen=true</BuildArguments>
```
dagood: On first glance I think `-DisableCrossgen=true` should be factored out so it only appears on one place, but that means repeating the `arm` condition again (unless I'm not thinking of a better factoring). I'm fine either way.

eerhardt:
Do we have a bug logged to track investigating why crossgen is failing for FreeBSD?

repos/core-setup.proj
eerhardt: I think this is the wrong place for these kinds of settings. These kinds of things should go into the repo itself.

A concrete example why is the line directly above this:

`<BuildArguments Condition="$(Platform.Contains('arm'))">$(BuildArguments) -TargetArchitecture=$(Platform) -DisableCrossgen=true -CrossBuild=true</BuildArguments>`

In 2.1, we enabled crossgen on `linux-arm` builds.  However, if someone was using source-build to build for `linux-arm`, now crossgen is turned off, just because this outdated setting is here.

Another side of the same reason: if I was building core-setup directly on FreeBSD (outside of source-build), my build would try to crossgen and it would fail.

Instead, I think we should be defaulting `DisableCrossgen` inside of core-setup's msbuild files correctly - https://github.com/dotnet/core-setup/blob/57692e1e8e52ad0e373682f3b1c0c33272269502/dir.props#L275

weshaggard: I agree these are better served for everyone if they are in the repo itself instead of in source-build.

wfurt: Should I (re)move the arm section as well? 
Note, that source-build is right now only one way how to build on FreeBSD as we do not have any packages published.

eerhardt: If you want to do the work in core-setup to default `DisableCrossgen` on arm appropriately, I won't stop you. It will be very close to what you will need to do for `FreeBSD`.

Issue #561, opened 2018-05-17
Remove source code modification workaround that injects ProdCon NuGet feed into CLI


dagood:
For 2.1 RTM, we started changing a CLI MSBuild file to inject the current ProdCon feed: https://github.com/dotnet/source-build/pull/556. This was done because the `//` was turned into `/` by MSBuild when we tried to pass it in through the usual environment variable, and we couldn't find a way to escape it.

The source-build repo API approach is to pass a file that CLI would import to read a property that contains all the sources.

To remove this code modification workaround, we need to decide if we want a targeted fix (maybe adding unescaping to the CLI file), have CLI implement this one part of the source-build repo API, or think forward to what the Arcade/ProdCon v2 approach will be.

/cc @dleeapho @crummel @dseefeld

dseefeld:
@dagood to create an issue in CLI for this.

dagood:
I'm not sure anymore that we **need** a CLI fix for this. On Fedora28, in CLI `9ab03e695`, I can run:

```
export ExternalRestoreSources='https://example.com/foo/bar;/home/abc'
./build.sh
(cancel the build when it starts restoring to save time)
cat NuGet.Config
```

And I see these lines, which are fine:

```
<add key="PrivateBlobFeedhttps://example.com/foo/bar" value="https://example.com/foo/bar" />
<add key="PrivateBlobFeed/home/abc" value="/home/abc" />
```

I think the issue is that we're storing the environment variables in Items in source-build, and/or `Exec` is interpreting things.

That said, it would still definitely be easiest for us if the CLI implemented `DotNetRestoreSourcePropsPath`, so I'm submitting that as an issue. Escaping (especially the `;`) is notoriously difficult to get right everywhere.

MichaelSimons:
[Triage] The injection of the ProdCon feed is handled by generic Arcade infra.  The aforementioned mechanism has been removed.  Closing as obsolete.

PR #639, opened 2018-07-11
Update cli project to build cli after repo refactoring

dseefeld:
* Update RootRepo to build through cli for everything but Windows
* Remove build args and flags that no longer apply to cli
* Update RepositoryReferences to only include ones used by cli

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -92,7 +92,8 @@
     <RootRepo>cli</RootRepo>
     <!-- Temporarily update RootRepo for Windows to core-setup repo project
     until all repos build clean out of master -->
-    <RootRepo Condition="'$(OS))' == 'Windows_NT'">core-setup</RootRepo>
+    <RootRepo Condition="'$(OS)' == 'Windows_NT'">core-setup</RootRepo>
+    <!-- Update RootRepo for arm to only build through core-setup -->
     <RootRepo Condition="$(Platform.Contains('arm'))">core-setup</RootRepo>
   </PropertyGroup>
 
--- a/dir.props
+++ b/dir.props
@@ -89,17 +89,10 @@
   <Import Project="$(ProjectDir)dependencies.props" />
 
   <PropertyGroup>
-    <!--
-        Temporarily update RootRepo to known-good
-        repo project until all repos are building clean.
-        When all repos build, this can be changed back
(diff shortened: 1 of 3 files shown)


dir.props
```diff
@@ -89,17 +89,10 @@
...
-        When all repos build, this can be changed back
-        to cli.
     <RootRepo>cli</RootRepo>
-    -->
-
-    <!-- Temporarily update RootRepo to core-setup repo project
+    <!-- Temporarily update RootRepo for Windows to core-setup repo project
     until all repos build clean out of master -->
-    <RootRepo>core-setup</RootRepo>
+    <RootRepo Condition="'$(OS))' == 'Windows_NT'">core-setup</RootRepo>
```
dagood: Typo: extra `)` in `$(OS))`

dir.props
```diff
@@ -89,17 +89,10 @@
...
-    <!--
-        Temporarily update RootRepo to known-good
-        repo project until all repos are building clean.
-        When all repos build, this can be changed back
-        to cli.
     <RootRepo>cli</RootRepo>
-    -->
-
-    <!-- Temporarily update RootRepo to core-setup repo project
+    <!-- Temporarily update RootRepo for Windows to core-setup repo project
```
dagood: Maybe the comment should be split into one each for Windows vs. arm? (Also justifies having two props setting the same value IMO.)

repos/cli.proj
```diff
@@ -118,37 +91,5 @@
                       Overwrite="true" />
   </Target>
```
dagood: Do you mind adding a link to https://github.com/dotnet/cli/issues/9457 tracking adding a feature that lets us remove this target while you're cleaning up?

dagood: Or https://github.com/dotnet/source-build/issues/561 actually.

dir.props
dseefeld: Added an additional comment for  arm.

dir.props
dseefeld: Fixed.  Thanks.

repos/cli.proj
dseefeld: I added a comment with this link.

Issue #643, opened 2018-07-12
Fix release/2.1 auto-update: failing in VSTS GeneralExecutor definition


dagood:
The release/2.1 auto-update builds are failing due to `error MSB4062: The "JoinItems" task could not be loaded from the assembly ...`.

I worked around for private auto-update by calling InitBuild explicitly, but this won't work as easily as a Maestro subscription where only command gets called:

https://github.com/dotnet/source-build/blob/4ca4d03aaa959683451c570fd25fa656e102b4fb/.vsts.pipelines/builds/vsts_prodcon_update.yml#L64-L66

We should figure out how to make sure InitBuild runs properly before trying to auto-update.

I'm not sure off the top of my head how DLL loading works for tasks built within the MSBuild context being run, so I don't know if a simple target dependency will work. We may need to do something like the nested MSBuild call(s) that we do for the normal source-build.

dagood:
I overlooked the simple solution to `JoinItems`: just add a step to the GeneralExecutor def. There's another problem behind that, so I'm making this issue generic to track getting auto-updates back up.

It'd be more time consuming to do these manually for 2.1 *and* 2.2, so it would be nice to get the automation running again.

dagood:
It turns out source-build didn't have a web hook pointing at Maestro yet, so commits weren't being mirrored to VSTS. This caused auto-update to try to make an  update based on a really old commit that didn't support the more recent manifests.

After fixing the mirror, rerunning the auto-PR def generated https://github.com/dotnet/source-build/pull/670. 🎉

Issue #611, opened 2018-06-13
Auto-update AspNetCore package versions from ProdCon builds


dagood:
These versions should be updated from the ProdCon build to remove a manual servicing build update step:

https://github.com/dotnet/source-build/blob/6ec0829d9b40171f512e4fad7920977c8570c8da/repos/cli.proj#L47-L52

dagood:
We have to track these separately: ASP.NET Core intentionally versions each package independently. 2.1.3 built each of these:

```xml
<ExtraPackageVersionPropsPackageInfo Include="Microsoft.AspNetCore.All" Version="2.1.3" />
<ExtraPackageVersionPropsPackageInfo Include="Microsoft.AspNetCore.App" Version="2.1.3" />
<ExtraPackageVersionPropsPackageInfo Include="Microsoft.NET.Sdk.Razor" Version="2.1.2" />
```

FYI @dleeapho @dseefeld @crummel

dseefeld:
Is this complete?

dagood:
`release/2.1` and `release/2.2` do:

https://github.com/dotnet/source-build/blob/b5a02e93099cbef7ad5e6fce3800903322859f7c/dependencies.targets#L99-L103

https://github.com/dotnet/source-build/blob/f1c73c8babc0303b57ddf5f2355dacc38f229cd4/dependencies.targets#L99-L103

`master` doesn't have it, but it doesn't need it. It's only useful for uptaking ProdCon v1 builds. For 3.0, ProdCon v2 obviates this work and we'll have aspnet building from source anyway.

Closing as complete.

Issue #606, opened 2018-06-12
Assert source-built package-versions weren't restored by an earlier project


dagood:
Catch this scenario and exit early:

 1. A dev attempts to upgrade CLI to the next version, e.g. 2.1.300-rc1 to 2.1.300.
 1. init-tools detects the runtime the new SDK is using, `2.1.0`.
    * Before the upgrade, it would have detected, say, `2.1.0-rc1-12345`.
 1. init-tools restores `Microsoft.NETCore.App 2.1.0` from the NuGet Gallery.
 1. Core-Setup builds `Microsoft.NETCore.App 2.1.0`.
 1. CLI restores `Microsoft.NETCore.App 2.1.0`. It finds the NuGet Gallery one because the id + version already exists in `packages/`.
    * CLI is unable to get the source-built package. The important difference in this case is that the source-built package has the RID being source-built, e.g. centos.
 1. An error occurs in CLI because the NuGet Gallery package doesn't have the runtime being source-built.

The CLI error is difficult to investigate, since the cause is so far from the real problem. We can catch it after the Core-Setup build by looking at all the new packages in the blob feed and checking to see if any exist in `packages/`.

This kind of error isn't limited to tooling restore vs. a repo build--it could also happen between two source-built repos.

dagood:
We hit this again when trying to move from 2.1.300 => 2.1.301. The workaround is to make sure the tooling SDK contains a different runtime version than the one that Core-Setup is producing. (E.g. it's fine to upgrade to the N-1 version, but upgrading to N causes this issue.)

/cc @dseefeld

dagood:
Making this assert as written would be good for diagnosing, but it's not compatible with the idea of the bootstrap flow where we expect some "circular" dependencies. E.g. Microsoft.NETCore.Platforms being required to build Platforms.

If the bootstrap flow is far enough away, this assert still seems worthwhile to me because it catches a pretty common issue and it wouldn't take much to implement. A basic idea is to verify after every build that all new source-built nupkgs aren't in the `packages/` dir. (Or they match the nupkg in `packages/` exactly, if there's a dependency within a single repo that gets resolved through packages.)

However, it might not be difficult to set up the bootstrap flow, either, and maybe we should just do that.

PR #644, opened 2018-07-12
Update CoreClr, ProdCon to servicing-26708-02, 20180708-01-1845879, respectively

dagood:
(Note: CoreCLR in the title refers to the ILAsm tool version in tools-local/ILAsmVersion.txt, not related to prodcon.)

I noticed that auto-update builds are failing, filed https://github.com/dotnet/source-build/issues/643 to fix.


repos/cli.proj
```diff
@@ -49,11 +49,10 @@
       Include="
         Microsoft.AspNetCore.All;
         Microsoft.AspNetCore.App"
-      Version="2.1.2" />
-    <!-- This package wasn't built in latest 2.1.2 build: stick with what we had from 2.1.1. -->
+      Version="2.1.3" />
     <ExtraPackageVersionPropsPackageInfo
       Include="Microsoft.NET.Sdk.Razor"
-      Version="2.1.1" />
+      Version="2.1.2" />
```
dagood: See https://github.com/dotnet/source-build/issues/611 for plans to auto-update this, and why 2.1.2 was built in the 2.1.3 build train.

dagood:
Looks like our NuGet version is behind (`4.7.0` vs. `4.8.0`), and we're not building the `Microsoft.Build.NuGetSdkResolver` package. This causes a conflict between the direct dependency on `NuGet.ProjectModel` (which we do build) vs. the prebuilt `Microsoft.Build.NuGetSdkResolver`'s transitive dependency on `NuGet.ProjectModel`.

```
cli/src/redist/redist.csproj : error NU1605: Detected package downgrade: NuGet.ProjectModel from 4.8.0-preview3.5278 to 4.7.0-rtm.5156. Reference the package directly from the project to select a different version.
cli/src/redist/redist.csproj : error NU1605: redist -> Microsoft.Build.NuGetSdkResolver 4.8.0-preview3.5278 -> NuGet.Commands 4.8.0-preview3.5278 -> NuGet.ProjectModel (>= 4.8.0-preview3.5278)
cli/src/redist/redist.csproj : error NU1605: redist -> NuGet.ProjectModel (>= 4.7.0-rtm.5156)
```

(`NU1605` is normally a warning, but since CLI has `TreatWarningsAsErrors` and doesn't include it in `NoWarn` it is an error in this case.)

ProdCon for 2.1.3 is not building NuGet. I'll see what updating that repo looks like.

patches/nuget-client/0001-Add-TargetingPack-and-enable-net46-builds-for-source.patch
```diff
@@ -1,28 +1,28 @@
-From a0c69929154c148ccbbf5e833c830d3be512df68 Mon Sep 17 00:00:00 2001
+From 18a85b215bacb7ca8fe0d4cdae3cd591b8899cbc Mon Sep 17 00:00:00 2001
 From: Chris Rummel <crummel@microsoft.com>
 Date: Tue, 12 Jun 2018 18:14:46 -0500
 Subject: [PATCH] Add TargetingPack and enable net46 builds for source-build.
```
dagood: @crummel is there anything tracking removing this patch I could add a link to?

crummel: This is https://github.com/NuGet/NuGet.Client/pull/2287.

dagood:

Hoping this is just some infra failure:

```
00:04:03.465 + ./build.sh /p:Configuration=Release /clp:v=detailed /p:MinimalConsoleLogOutput=false
00:04:16.784 Installing dotnet cli...
00:04:21.716 Restoring BuildTools version 2.1.0-rc1-03006-01...
00:04:24.926 
Build step 'Execute shell' marked build as failure
```

dagood:
Hitting errors building roslyn-tools inside tarball. I think the roslyn-tools commit isn't able to build itself properly: when I copy the "online" roslyn-tools `tools/*` (mainly MSBuild files) to `Tools/source-built/RoslynTools.RepoToolset` in the tarball, overwriting, the roslyn-tools build works.

The build then fails at the msbuild repo saying it can't find some Roslyn packages. I imagine this is because roslyn didn't build yet and we filter out prebuilt nupkgs that we build, which could include this package. Still looking into it.

dagood:
> I imagine this is because roslyn didn't build yet and we filter out prebuilt nupkgs that we build, which could include this package. Still looking into it.

This is the case. (https://github.com/dotnet/source-build/issues/606 tracks catching this early and descriptively.)

I think this means `msbuild` should actually be Tier 2, since we now know depends on the Tier 1 repo `roslyn`. This shouldn't affect others that depend on `msbuild`: `sdk` is Tier 5, `cli` is Tier 6. At least... there no dependencies *that we know of* affected.

dagood:
@dseefeld @crummel could you take a look at the workaround? (More info about it at https://github.com/dotnet/source-build/issues/654.) Local build on Fedora28 with the current commit works and smoke-tests from tarball, so I'm hopeful this is it.

crummel:
@dotnet/dnceng Looks like the Mac needs cleaning:
`/Library/Developer/CommandLineTools/usr/bin/ranlib: can't write to output file (No space left on device)`

PR #645, opened 2018-07-13
Update core repos tip to latest master

safern:
The repos where moved to produce netcoreapp3.0 packages. If someone wants to test local changes with a produced runtime and point to a updated coreclr/corefx but leave the core-setup submodule as is, they will find issues since we're pointing to a core-setup commit that is still restoring packages for netcoreapp2.2

cc: @weshaggard @dagood 

FYI: @GrabYourPitchforks


DotnetCLIVersion.txt
```diff
@@ -1 +1 @@
-2.1.301
+2.1.401-preview-009081
```
dagood: This should be `2.1.302` (N-1) at maximum right now. I don't know if there's a plan for how to build targeting netcoreapp3.0 without it yet.

/cc @dleeapho @dseefeld @crummel

dagood: I see some discussion here about this version: https://github.com/dotnet/corefx/pull/30949. /cc @eerhardt.

eerhardt: corefx needs this version because of https://github.com/dotnet/core-setup/issues/4291.

I'm not too concerned here in the 'master' branch on staying on N-1.  `2.1.401` SDK will ship LONG before `3.0` will ship.

dagood: Good point, we're already getting the `2.1.401` SDK ready in the `release/2.1` branch. This seems fine to be practical.

weshaggard:
@safern I suspect you also need to update the version of BuildTools in order to fix the other CI issues.

DotnetCLIVersion.txt
dleeapho: Agree, `2.1.401` is good for `3.0`

safern:
test Windows_NT Release please
test CentOS7.1 Release

weshaggard:
To fix the coreclr build you need to move coreclr up to include https://github.com/dotnet/coreclr/pull/18922

PR #577, opened 2018-05-25
Merge dev/release/2.1 into master

eerhardt:
There were a few changes that went into 2.1 that should be ported to master.

PR #598, opened 2018-06-08
Upgrade to latest prodcon build and add targeting packs for net46 support

crummel:
Testing build with net46 changes

PR #646, opened 2018-07-13
Merge release/2.1 infrastructure changes into master

dagood:
The 2.1.2 release saw some changes in `release/2.1` infra to support VSTS builds. This merge takes those changes into `master` for the future. I made some edits to the merge commit to fix conflicts and remove any submodule upgrades related to the 2.1.2 release.

**Don't squash this PR.** Squashing https://github.com/dotnet/source-build/pull/577 (an earlier 2.1 => master merge) made some commits it included to show up in this PR's commit list, too.

### DIFF
--- a/repos/dir.props
+++ b/repos/dir.props
@@ -53,6 +53,7 @@
     <EnvironmentVariables Include="DotNetCoreSdkDir=$(DotNetCliToolDir)" />
     <EnvironmentVariables Include="DotNetBuildToolsDir=$(ProjectDir)Tools/" />
     <EnvironmentVariables Include="StabilizePackageVersion=$(UseStableVersions)" Condition="'$(UseStableVersions)' != ''" />
+    <EnvironmentVariables Include="PB_IsStable=$(UseStableVersions)" Condition="'$(UseStableVersions)' != ''" />
     <EnvironmentVariables Include="PackageVersionStamp=$(VersionStamp)" Condition="'$(VersionStamp)' != ''" />
 
     <!-- Turn off node reuse for source build because repos use conflicting versions 
--- a/repos/nuget-client.proj
+++ b/repos/nuget-client.proj
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <EnvironmentVariables Include="MS_PFX_PATH=$(NuGetKeyFilePath)" />
+    <EnvironmentVariables Include="NUGET_PFX_PATH=$(NuGetKeyFilePath)" />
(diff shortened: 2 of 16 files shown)


repos/nuget-client.proj
```diff
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <EnvironmentVariables Include="MS_PFX_PATH=$(NuGetKeyFilePath)" />
+    <EnvironmentVariables Include="NUGET_PFX_PATH=$(NuGetKeyFilePath)" />
```
dagood: I wasn't sure about this when I first saw it. I see now it comes from https://github.com/dotnet/source-build/pull/598, so we probably do need it. @crummel

repos/dir.props
```diff
@@ -53,6 +53,7 @@
     <EnvironmentVariables Include="DotNetCoreSdkDir=$(DotNetCliToolDir)" />
     <EnvironmentVariables Include="DotNetBuildToolsDir=$(ProjectDir)Tools/" />
     <EnvironmentVariables Include="StabilizePackageVersion=$(UseStableVersions)" Condition="'$(UseStableVersions)' != ''" />
+    <EnvironmentVariables Include="PB_IsStable=$(UseStableVersions)" Condition="'$(UseStableVersions)' != ''" />
```
dagood: This is from https://github.com/dotnet/source-build/commit/cdf7249ccb14bf1b1da499f0c283b78129556036, we do need it for proper versions.

netci.groovy
```diff
@@ -44,7 +44,7 @@ def addBuildStepsAndSetMachineAffinity(def job, String os, String configuration)
         shell("git submodule update --init --recursive");
         shell("./build.sh /p:Configuration=${configuration} ${loggingOptions}");
         smokeTestExcludes = "";
-        if (os == "OSX10.12") {
+        if (os == "Fedora24" || os == "OSX10.12") {
```
dseefeld: We probably don't need this change in master.  Master is updated to build on Fedora28, now.

dagood: Good call, removed.

repos/nuget-client.proj
crummel: Yeah, there's now two different variables used for the PFX path, we will attempt to delay-sign and fail without this.  I can look at whether we can disable signing entirely, that seems like the better option.

dagood: Cool, filed https://github.com/dotnet/source-build/issues/647.

PR #649, opened 2018-07-16
Detect the base commit in extract-patches.sh

dagood:
Small improvement to `extract-patches.sh` to simplify patch workflow:

 1. `git am ../../patches/<repo>/*`
 2. Make changes, fixups, etc.
 3. `../../extract-patches.sh`
 4. Reset inadvertent changes to existing patch files due to client formatting or the new "From" commit hash.

### DIFF
--- a/extract-patches.sh
+++ b/extract-patches.sh
@@ -2,6 +2,13 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+getIndexedSubmoduleSha() {
+    (
+        cd "$DIR"
+        git ls-tree HEAD src/cli | awk '{print $3}'
+    )
+}
+
 SOURCE="${BASH_SOURCE[0]}"
 while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
   DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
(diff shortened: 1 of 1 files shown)


extract-patches.sh
```diff
@@ -2,6 +2,13 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+getIndexedSubmoduleSha() {
+    (
+        cd "$DIR"
+        git ls-tree HEAD src/cli | awk '{print $3}'
```
dseefeld: I don't understand why this is specific to cli?

dagood: My bad, it's not. Fixing

extract-patches.sh
```diff
@@ -2,6 +2,13 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+getIndexedSubmoduleSha() {
+    (
+        cd "$DIR"
+        git ls-tree HEAD "src/$REPO_NAME" | awk '{print $3}'
```
crummel: I think this will fail for NuGet's nested submodules - `REPO_NAME` will just be `Common`, etc.

dagood: Yep, fixed. I don't think we've done nested submodule patches in the build, but no need to prevent that here.

dagood:
[Assuming just machine or nuget flakiness.](https://ci.dot.net/job/dotnet_source-build/job/release_2.1/job/RHEL7.2_Unshared_Release_prtest/37/)

Issue #635, opened 2018-07-09
Get CI working through smoke-test


dseefeld:
Once core-sdk is building, re-enable smoke-test in CI.

crummel:
My PR https://github.com/dotnet/source-build/pull/658 introduced two hacks in smoke-test that have to be removed after getting the associated scenarios passing:
- Disabling tests with remote sources and web tests in general (packages are not available yet).
- Replacing `netcoreapp2.1` with `netcoreapp3.0` in projects after `new` and before `restore`.

dagood:
We should be good to go with 3.0.0-preview1 submodules set up and 3.0 preview1 Microsoft build released. Moving to backlog.

dleeapho:
Closing since we have this working in 3.0 Preview 2.

Issue #554, opened 2018-05-16
Get core-setup versions from authoritative location instead of regex


crummel:
See https://github.com/dotnet/source-build/pull/551#issuecomment-389597137.

We do not have a regex that works for all cases:
* 2.1.0
* 2.1.0-rc1
* 2.1.0-rc1-12345-67

And does not capture
* 2.1.0-ubuntu

A better approach is to get the version out of NuGet packages or some other authoritative location.

dleeapho:
Is this a dupe of #550 ?

crummel:
I don't think so.  #550 is for getting stabilization/version stamp status from ProdCon so we can match the official build's version numbers.  This is an issue in our `PublishCoreSetupBinaries` task where we don't have a good way to get the version numbers for assets that core-setup produced.  We could generate this version ourselves based on the stabilization/version stamp info from prodcon but I'm not sure we want to duplicate that code.

dagood:
Something to consider is having core-setup itself publish to the location we expect, and we just tell it the base dir. This is similar to how it works for ProdCon (although the code probably doesn't work right now with a file system).

crummel:
When this is fixed we should also remove `tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/PublishToolsetBinaries.cs` as it is mostly duplicated code with a changed regex.

MichaelSimons:
[Triage] This issue is obsolete.  Source-build no longer uses this regex to retrieve the version number.

PR #658, opened 2018-07-19
Update source-build master to core repo masters

crummel:
Most of the work here was in adding the new toolset and core-sdk repos and adjusting our current process to the way they work.  I also had the usual build number/hash updates.  This includes a few workarounds and patches that I believe are all tagged with an associated issue or PR:
- `PublishToolsetBinaries` is a near-duplicate of `PublishCoreSetupBinaries` with a changed regex: https://github.com/dotnet/source-build/issues/554.
- `smoke-test.sh` always skips some tests and replaces the TFM in projects: https://github.com/dotnet/source-build/issues/635.
- Several patches among core-sdk, toolset, and CLI.  These have the issues in their descriptions.

### DIFF
--- a/repos/core-sdk.proj
+++ b/repos/core-sdk.proj
@@ -3,6 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
 
   <PropertyGroup>
+    <BuildCommandArgs Condition="'$(TargetOS)' == 'Linux'">--runtime-id $(TargetRid)</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) -c $(Configuration)</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) --stage0 $(DotNetCliToolDir)</BuildCommandArgs>
 
--- a/repos/core-sdk.proj
+++ b/repos/core-sdk.proj
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
 
   <PropertyGroup>
-    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -c $(Configuration) --stage0 $(DotNetCliToolDir) --</BuildCommand>
+    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -c $(Configuration) --stage0 $(DotNetCliToolDir)</BuildCommand>
     <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
   </PropertyGroup>
 
--- a/repos/core-sdk.proj
+++ b/repos/core-sdk.proj
@@ -36,6 +36,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <RepositoryReference Include="javascriptservices" />
+    <RepositoryReference Include="newtonsoft-json" />
     <RepositoryReference Include="toolset" />
   </ItemGroup>
 
--- a/repos/core-sdk.proj
+++ b/repos/core-sdk.proj
@@ -3,13 +3,112 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
 
   <PropertyGroup>
-    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -c $(Configuration) --stage0 $(DotNetCliToolDir)</BuildCommand>
+    <BuildCommandArgs>$(BuildCommandArgs) -c $(Configuration)</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) --stage0 $(DotNetCliToolDir)</BuildCommandArgs>
+
+    <!-- Consume the source-built Core-Setup. This line must be removed to source-build CLI without source-building Core-Setup first. -->
+    <BuildCommandArgs>$(BuildCommandArgs) /p:CoreSetupBlobRootUrl=file:%2F%2F$(LocalBlobStorageRoot)</BuildCommandArgs>
+
+    <!-- Pass in package version props using the Product Construction (ProdCon) API. -->
+    <BuildCommandArgs>$(BuildCommandArgs) /p:PB_PackageVersionPropsUrl=file:%2F%2F$(PackageVersionPropsPath)</BuildCommandArgs>
+
+    <BuildCommandArgs>$(BuildCommandArgs) /p:SkipBuildingInstallers=true</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:IncludeNuGetPackageArchive=false</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:IncludeAdditionalSharedFrameworks=false</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:UsePortableLinuxSharedFramework=false</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:IncludeSharedFrameworksForBackwardsCompatibilityTests=false</BuildCommandArgs>
+
+    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) $(BuildCommandArgs)</BuildCommand>
+
+    <PackageOutputRid Condition="'$(TargetOS)' == 'Windows_NT'">win-x64</PackageOutputRid>
+    <PackageOutputRid Condition="'$(TargetOS)' == 'OSX'">osx-x64</PackageOutputRid>
+    <PackageOutputRid Condition="'$(PackageOutputRid)' == ''">$(TargetRid)</PackageOutputRid>
+    <PackagesOutput>$(ProjectDirectory)bin/2/$(PackageOutputRid)/packages/</PackagesOutput>
+    <TarBallOutput>$(PackagesOutput)</TarBallOutput>
+    <RepoApiImplemented>false</RepoApiImplemented>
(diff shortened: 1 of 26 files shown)


crummel:
cc @nguerrera @dsplaisted @weshaggard if any of you would like to take a look.

crummel:

repos/core-sdk.proj
```diff
@@ -0,0 +1,115 @@
...
+    <EnvironmentExternalRestoreSources>$(SourceBuiltPackagesPath)</EnvironmentExternalRestoreSources>
+    <EnvironmentExternalRestoreSources Condition="'$(OfflineBuild)' == 'true'">$(EnvironmentExternalRestoreSources)%3B$(PrebuiltPackagesPath)</EnvironmentExternalRestoreSources>
+
+    <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <RepositoryReference Include="toolset" />
+  </ItemGroup>
+
```
dseefeld: Do we not still need the ExtraPackageVersionPropsPackageInfo stuff for AspNetCore.All, AspNetCore.App and NET.Sdk.Razor like we had in CLI?

repos/toolset.proj
```diff
@@ -0,0 +1,68 @@
...
+    <!-- Produce stable versions for RTM only.  When https://github.com/dotnet/source-build/issues/550
+                  is fixed we will be able to determine this automatically.  -->
+    <EnvironmentVariables Include="DropSuffix=true"
+                          Condition="'$(UseStableVersions)' == 'true'" />
+
+    <EnvironmentVariables Include="CLIBUILD_SKIP_TESTS=true" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <RepositoryReference Include="cli" />
```
dseefeld: In https://github.com/dotnet/source-build/pull/639, I removed the references to app-insights, fsharp, javascriptservices, netcorecli-fsc, newtonsoft-json, vstest and websdk from repos/cli.proj because it wasn't dependent on these.  These should probably be added to either toolset.proj or core-sdk.proj.

repos/core-sdk.proj
dagood: I don't think we need those anymore. They're for ProdCon builds, because the CLI Git repo state is not reliably auto-updated. For `master` we shouldn't need to worry about that at any point in the future, since we won't use ProdCon for 3.0. With ProdCon v2, it won't ever be out of date because dependency changes are checked in.

repos/corefx.proj
```diff
@@ -3,11 +3,11 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
   <PropertyGroup>
     <BuildArguments>-$(Configuration) -buildArch=$(Platform) -portable=$(PortableBuild) -BuildTests=false</BuildArguments>
-    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) $(BuildArguments) -- /p:ILLinkTrimAssembly=false</BuildCommand>
+    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) $(BuildArguments) -- /p:ILLinkTrimAssembly=false /p:DotNetAssetRootUrl=file:%2F%2F$(SourceBuiltAssetsDir)</BuildCommand>
```
dagood: What's `/p:DotNetAssetRootUrl` for in corefx?

repos/toolset.proj
```diff
@@ -0,0 +1,68 @@
...
+  <UsingTask AssemblyFile="$(TasksBinDir)Microsoft.DotNet.SourceBuild.Tasks.dll" TaskName="PublishToolsetBinaries" />
+
+  <!-- Need to "publish" the tarballs for cli, until cli respects the assets blob -->
+  <Target Name="PublishBinaries" AfterTargets="Build">
+    <ItemGroup>
+      <ToolsetZips Include="$(ToolsetZipOutput)dotnet-*.zip" />
+    </ItemGroup>
+    <Message Importance="High" Text="Publishing @(ToolsetZips) from $(ToolsetZipOutput) to $(LocalBlobStorageRoot)Toolset/{version}/" />
+    <PublishToolsetBinaries Binaries="@(ToolsetZips)"
+                              DestinationFolderTemplate="$(LocalBlobStorageRoot)Toolset/{version}/" />
```
dagood: (Nit) Misalignment

repos/toolset.proj
```diff
@@ -0,0 +1,68 @@
...
+    <BuildCommandArgs>$(BuildCommandArgs) /p:IncludeNuGetPackageArchive=false</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:UsePortableLinuxSharedFramework=false</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:IncludeSharedFrameworksForBackwardsCompatibilityTests=false</BuildCommandArgs>
+
+    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) $(BuildCommandArgs)</BuildCommand>
+
+    <PackageOutputRid Condition="'$(TargetOS)' == 'Windows_NT'">win-x64</PackageOutputRid>
+    <PackageOutputRid Condition="'$(TargetOS)' == 'OSX'">osx-x64</PackageOutputRid>
+    <PackageOutputRid Condition="'$(PackageOutputRid)' == ''">$(TargetRid)</PackageOutputRid>
+    <ToolsetZipOutput>$(ProjectDirectory)bin/2/$(PackageOutputRid)/packages/</ToolsetZipOutput>
```
dagood: Could you call this `ToolsetZipOutputDir` for convention?

smoke-test.sh
```diff
@@ -161,6 +161,12 @@ function doCommand() {
             else
                 "${dotnetCmd}" $newArgs >> "$logFile" 2>&1
             fi
+            # XXX crummel XXX
```
dagood: (Nit) IMO the username comments aren't needed with the issue link + git blame.

smoke-test.sh
```diff
@@ -349,6 +355,14 @@ fi
 export NUGET_PACKAGES="$restoredPackagesDir"
 SOURCE_BUILT_PKGS_PATH="$SCRIPT_ROOT/bin/obj/x64/$configuration/blob-feed/packages/"
 
+# XXX crummel XXX
+# This is a temporary workaround to disable tests that will fail until ASP.NET packages are updated.
+# See https://github.com/dotnet/source-build/issues/635 for more details.
+echo "WARNING: Ignoring parameters to disable known-failing tests." | tee -a "$logFile"
```
dagood: (Nit) `Ignoring parameters` sounds kinda alarmingly general, I'd just list `Setting excludeWebTests and excludeOnlineTests to disable known-failing tests.` even though it's a little bit of extra maintenance in theory.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/PublishToolsetBinaries.cs
```diff
@@ -0,0 +1,62 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+// This is mostly a duplicate of PublishCoreSetupBinaries with a
+// different regex (because version numbers are formatted differently
+// between core-setup and toolset).
+// https://github.com/dotnet/source-build/issues/554 tracks figuring
+// out a better process for doing this.
```
dagood: Did you consider adding a bool to pick the regex instead of copying the file?

repos/core-sdk.proj
dseefeld: Got it.  Thanks!

repos/toolset.proj
crummel: I think this might have changed between 2.1 and master - in CLI I see [app-insights](https://github.com/dotnet/cli/blob/ddb64191922a5b5ac75232d71c8209c9f52cadbc/src/dotnet/dotnet.csproj#L58), [newtonsoft](https://github.com/dotnet/cli/blob/bdc5cede292e13f64c648b06b595d7ddb4eca6e5/src/dotnet/dotnet.csproj#L56), and [websdk](https://github.com/dotnet/cli/blob/f0b60ed02c513eb3cdf52e7a6b3353d6b8e52cf8/build/sdks/sdks.csproj#L12).  I added those back to CLI, the other removals still look good.  Also searched for and updated repo references in toolset and core-sdk.

repos/corefx.proj
crummel: I was attempting to fix a new warning, I'll back this out for now.

repos/toolset.proj
crummel: Fixed.

repos/toolset.proj
crummel: Fixed.

smoke-test.sh
crummel: Removed.

smoke-test.sh
crummel: Changed as suggested.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/PublishToolsetBinaries.cs
crummel: I thought this approach would stick out more as a bad thing where a boolean might get overlooked but I'm fine with that approach if you think it's better.

dagood: Well, to keep the task name different, could subclass and override. Or use a descriptive required string param to keep things distinguished. The thing I'm not a fan of here is that you have to do a diff to see what's different, and it just seems unnecessary in this case.

dseefeld:
@dotnet-bot test this please

crummel:
Merging ahead of CI as the last run passed and the latest commit has no functional change.

PR #704, opened 2018-08-22
[release/2.1] Split smoke test prebuilts out of main tarball

dagood:
* Create two tarballs: split the smoke test prereqs into a separate tarball. They aren't required at all for builds, and for smoke-test, they're only required when the packages can't easily be acquired through the usual endpoint.
* Reduce `prebuilt/smoke-test-packages/` from an expanded NuGet package cache to a directory of nupkg files. This eliminates the redundancy of having both the nupkgs and their expanded contents in the tarball. (290 MB => 146 MB.)
* Set the working directory so the tarball has the `tarball_<build number>` dir in it, not `tb/tarball_<build number>` (absolute path).

We did these fixes by hand a few times--it'll be nice to get them in by default.

The tarball's still missing a nice name. Right now it's like `tarball_1967206-smoke-test-prereqs.tar.gz`. I can't think of a great way to get the branding (like `2.1.3`) in.

Example build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=1967945

### DIFF
--- a/.vsts.pipelines/phases/ci-linux.yml
+++ b/.vsts.pipelines/phases/ci-linux.yml
@@ -129,9 +129,13 @@ phases:
 
   # tar the tarball directory into the drop directory.
   - script: |
-      $(docker.run) $(docker.tb.map) $(docker.drop.map) $(imageName) /bin/bash -c '
+      $(docker.run) $(docker.tb.map) $(docker.drop.map) $(docker.tb.work) $(imageName) /bin/bash -c '
         mkdir -p /drop/tarball/
-        tar -zcf "/drop/tarball/$(tarballName).tar.gz" "/tb/$(tarballName)"'
+        smokeTestPackages="$(tarballName)/prebuilt/smoke-test-packages"
+        find "$smokeTestPackages" -iname "*.nupkg" -exec mv {} "$smokeTestPackages" \;
+        find "$smokeTestPackages" -not -iname "*.nupkg" -delete
+        tar --numeric-owner "--exclude=$smokeTestPackages" -zcf "/drop/tarball/$(tarballName).tar.gz" "$(tarballName)"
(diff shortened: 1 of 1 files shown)


.vsts.pipelines/phases/ci-linux.yml
```diff
@@ -129,9 +129,13 @@ phases:
 
   # tar the tarball directory into the drop directory.
   - script: |
-      $(docker.run) $(docker.tb.map) $(docker.drop.map) $(imageName) /bin/bash -c '
+      $(docker.run) $(docker.tb.map) $(docker.drop.map) $(docker.tb.work) $(imageName) /bin/bash -c '
         mkdir -p /drop/tarball/
-        tar -zcf "/drop/tarball/$(tarballName).tar.gz" "/tb/$(tarballName)"'
+        smokeTestPackages="$(tarballName)/prebuilt/smoke-test-packages"
+        find "$smokeTestPackages" -iname "*.nupkg" -exec mv {} "$smokeTestPackages" \;
```
dseefeld: What is this line doing?  It looks like it's finding all *.nupkg files and moving them into the same directory it found them in.  Am I missing something?

dagood: Package cache structure is like this:

```
├── commandlineparser
│   └── 2.1.1-beta
│       ├── commandlineparser.2.1.1-beta.nupkg
... (contents of the nupkg)
├── coveralls.io
│   └── 1.4.0
│       ├── coveralls.io.1.4.0.nupkg
...
```

These `find`s put all the nupkgs into the root folder (like they are in `prebuilt/nuget-packages`), and cleans up the expanded files.

dseefeld: Oh, got it.  Thanks.

dagood: I'll add a comment. Just the `-delete` would probably produce a usable tarball (since NuGet does search recursively) but it seems better to keep the structure the same as the other prebuilt dir.

PR #749, opened 2018-09-06
Remove old docker files and add doc for using docker

dseefeld:
The docker files in the docker directory were old and moldy.  Removed all docker files and scripts and replaced with a document pointing out locations of docker images in microsoft/dotnet-buildtools-prereqs and how to build with them.

Skip CI Please


docker/DockerImages.md
```diff
@@ -0,0 +1,25 @@
+# Docker Images #
```
dagood: I'd recommend making this file `README.md` to make it show up without an extra click in GitHub. (Also to have an index if we want more docs here in the future.)

docker/DockerImages.md
```diff
@@ -0,0 +1,25 @@
...
+
+The following docker images have been verified to successfully build the source-build repo.
```
dagood: > have been

In the context of putting the doc in the repo, I think it's important to say how/when.

dagood: I think you should link https://github.com/dotnet/dotnet-buildtools-prereqs-docker, maybe a direct link to each Dockerfile in the table below?

docker/DockerImages.md
```diff
@@ -0,0 +1,25 @@
...
+> docker volume create source-build-volume 
+
+> ### Clone source-build repo
+> docker run -it --rm --mount source=source-build-test,target=/src -w /src <imageName> /bin/bash -c "git clone https://github.com/dotnet/source-build"
+
+> ### Build source-build to create a tarball
+> docker run -it --rm --mount source=source-build-test,target=/src -w /src/source-build <imageName> /bin/bash -c "./build-source-tarball.sh ../tarball"
+
+> ### Build the tarball
+> docker run -it --rm --mount source=source-build-test,target=/src -w /src/tarball <imageName> /bin/bash -c "./build.sh"
```
dagood: I feel like it would be more useful to give the reader something easy to understand at first glance, and they can use more Docker features if they want to. How about:

```sh
### Run a bash shell inside a docker container. All later steps run in the interactive shell.
docker run -it --rm <imageName> /bin/bash

### Clone source-build repo
git clone https://github.com/dotnet/source-build
cd source-build

### Build source-build to create a tarball
./build-source-tarball.sh ../tarball

### Build the tarball
cd ../tarball
./build.sh

### Clean up the container when you're done working with the build results
exit
```

docker/DockerImages.md
dseefeld: Done.

docker/DockerImages.md
dseefeld: I added how in the comments and a "Verification Date" column in the table.

dseefeld: Agreed.  Added.

docker/DockerImages.md
dseefeld: I added this as an "interactive docker container" section and left the other as an example with a docker volume.

docker/README.md
```diff
@@ -0,0 +1,62 @@
+# Docker Images #
+
+The following docker images have been verified to successfully build the source-build repo by:
+
+1. Building source-build
+2. Building the resulting tarball
+3. Running smoke tests
+
+| OS | Image Name | Link to Dockerfile | Verification Date |
```
dagood: Commit hash should be included IMO. Also, I think the table would look cleaner with the Dockerfile link just added to an existing column, and I think the link should be pinned to the image's commit, like:

| OS | Image Name |  Tested commit |
| --- | --- | --- |
| CentOS 7.1 | [`microsoft/dotnet-buildtools-prereqs:centos-7-b46d863-20180719033416`](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/b46d863/src/centos/7/Dockerfile) | [4f3eacd](https://github.com/dotnet/source-build/commit/4f3eacdd0551677444ab917d4e332653672e0bd5), 2018.09.05

docker/README.md
```diff
@@ -0,0 +1,62 @@
...
+./build.sh
+
+### Run smoke tests
+./smoke-test.sh
+
+### Clean up the container when you're done working with the build results
+exit
+```
+
+### Running build commands in a docker container with a docker volume for a working directory.  A docker volume allows you to persist data across docker container instances.
```
dagood: I don't think so much info belongs in a header. (A concrete reason is that this determines the anchor url, which is now https://github.com/dotnet/source-build/blob/7daf6c43b87b6cfd3129a60ee2ca4d0ee55d3969/docker/README.md#running-build-commands-in-a-docker-container-with-a-docker-volume-for-a-working-directory--a-docker-volume-allows-you-to-persist-data-across-docker-container-instances :smile:)

docker/README.md
dseefeld: The commit hash is in the image name, so I didn't think it was necessary in the link.

docker/README.md
dseefeld: Fixed

docker/README.md
dagood: I think it makes it easy to generate the link, but I think it's needed or else people won't notice they're being linked to a mismatched Dockerfile.

dseefeld: I see what you're saying.  Updated.

Issue #493, opened 2018-05-01
Create a prebuilt package baseline for the prebuilt usage report


dagood:
We should use the baseline to reduce the report to packages that we don't expect to be prebuilt. We can use that list to eliminate them over time, and alert us to new packages showing up as prebuilts to avoid regressions.

Related: https://github.com/dotnet/source-build/issues/461 (generating the reports)

PR #756, opened 2018-09-06
[release/2.1] Compare prebuilt usage against a baseline

dagood:
Check in a baseline of prebuilt usages for the online and offline build, and fail the build when more prebuilts are used than in the baseline. Updating the baseline is done by overwriting the current baseline with a generated file.

Validation also generates a diff that tells you which project.assets.json files restored the packages. The task also tells you when/which prebuilts are no longer being used.

I generated the first baselines (online and offline (tarball)) using `microsoft/dotnet-buildtools-prereqs:centos-7-b46d863-20180719033416`.

One issue with baselining prebuilts is the potential prebuilt difference between platforms because of RIDs in runtime packages. For now, I made validation ignore the RID. We'll see if that's enough when CI runs through it. The baseline I have here worked on OSX for me.

(I also fixed some silly code that was making `WritePackageUsageData` take a minute when it only needed a second. The creation time of the PVP snapshot is also now stored in the snapshots, so the prebuilt-report dir is more resilient to being moved around.)

https://github.com/dotnet/source-build/issues/493


tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/ValidateUsageAgainstBaseline.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            Directory.CreateDirectory(Path.GetDirectoryName(OutputReportFile));
+            File.WriteAllText(OutputReportFile, report.ToString());
+
+            if (tellUserToUpdateBaseline)
+            {
+                Log.LogWarning(
+                    "Prebuilt usages are different from the baseline. If detected changes are " +
+                    "acceptable, update baseline with:\n" +
+                    $"cp '{OutputBaselineFile}' '{BaselineDataFile}'");
```
dseefeld: I like the instructions to update! 👍

PR #781, opened 2018-09-18
[WIP] Build all corefx packages

crummel:
Do not merge, using for BuildAllPackages testing.


crummel:

crummel:
@dotnet-bot test this please.

crummel:
@dagood @dseefeld This is looking pretty good to me as a first step, it removes two prebuilts:
```
<PackageIdentity Id="Microsoft.NETCore.Platforms" Version="2.1.1"/>
<PackageIdentity Id="Microsoft.NETCore.Targets" Version="2.0.0"/>
```
Do we want to go ahead with this first and save the marker files for later?

dagood:
I think I'm ok with this... my uneasiness is mainly that they'll get in where we should be using source-built packages, but the `GetSourceBuiltNupkgCacheConflicts` check should actually detect that. (And it didn't see any, at least in the Centos unshared leg.

It's kind of interesting that this dir is a place where someone sufficiently motivated could replace all the content with a previous build's nupkgs and get some small part of N-1 going.

Do you know how much size this adds to the tarball?

build-source-tarball.sh
```diff
@@ -129,6 +130,12 @@ do
     fi
 done
 
+for built_package in $(find $SCRIPT_ROOT/bin/obj/x64/Release/blob-feed/packages/ -name '*.nupkg')
+do
+    echo "copying $built_package to $TARBALL_ROOT/prebuilt/source-built/"
```
dagood: I think you should add an `echo` before the `for` saying what these copies are for.

dir.props
```diff
@@ -57,6 +57,7 @@
     <SourceBuiltPackagesPath>$(SourceBuiltBlobFeedDir)packages/</SourceBuiltPackagesPath>
     <SourceBuiltAssetsDir>$(SourceBuiltBlobFeedDir)assets/</SourceBuiltAssetsDir>
     <PrebuiltPackagesPath>$(ProjectDir)prebuilt/nuget-packages/</PrebuiltPackagesPath>
+    <PreviouslySourceBuiltPackagesPath>$(ProjectDir)prebuilt/source-built/</PreviouslySourceBuiltPackagesPath>
```
dagood: There's a `_PreviouslySourceBuiltPackages` property in `dir.targets` that means something quite different--packages built already during the current build. Call this `PrebuiltSourceBuiltPackagesPath`? Or find a better name for `_PreviouslySourceBuiltPackages`?

I'm also wondering if we should settle on terms for the online build vs. tarball build and use those instead of relative terms.

build-source-tarball.sh
crummel: I'll add an echo but printing each package was just for debugging, I'll remove it.

PR #790, opened 2018-09-26
Update MSBuild, NuGet.Client to build as used in 2.1.5

dagood:
New versions of these packages were used by the 2.1.5 build, but the repos weren't built during ProdCon so we didn't receive updates. NuGet.Client commit found via tag, MSBuild found by looking at the DLL version in the package.

This case would have been caught by the "Compare the source-built version of a package to the checked-in dependency's version" check proposed in https://github.com/dotnet/source-build/issues/748.

The update had me remove the only MSBuild patch!


dagood:
For in-place context, the plan is to get https://github.com/dotnet/source-build/pull/791 merged if possible--this is the backup.

PR #800, opened 2018-10-03
Add capture and re-build of reference-only packages to eliminate prebuilts

dseefeld:
Adds detection of packages in the source-build build that either a) have no binaries or b) only have dlls that are in the /ref/ folder.  For ones with dlls, disassembles them with ildasm and provides the IL as source to the tarball.  When building the tarball, these assemblies are re-build with ilasm and the packages are re-packed to be available for the tarball build.
There are a couple more items to complete before this is done:
- [x] There are two dlls that don't roundtrip with this process: System.Runtime.dll and netstandard.dll - These have been identified as an issue with ilasm/ildasm (see: https://github.com/dotnet/coreclr/issues/20262)
- [x] The prebuilt reporting doesn't detect that these packages are no longer prebuilts.

@dagood & @crummel  PTAL

### DIFF
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -148,13 +148,13 @@ done
 
 echo 'Removing reference-only packages from tarball prebuilts...'
 
-for built_package in $(find $SCRIPT_ROOT/reference-packages/packages-to-delete/ -name '*.nupkg' | tr '[:upper:]' '[:lower:]')
+for ref_package in $(find $SCRIPT_ROOT/reference-packages/packages-to-delete/ -name '*.nupkg' | tr '[:upper:]' '[:lower:]')
 do
-    if [ -e $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $built_package) ]; then
-        rm $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $built_package)
+    if [ -e $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $ref_package) ]; then
+        rm $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $ref_package)
     fi
-    if [ -e $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $built_package) ]; then
-        rm $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $built_package)
+    if [ -e $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $ref_package) ]; then
+        rm $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $ref_package)
     fi
 done
 
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -119,11 +119,11 @@ mkdir -p $TARBALL_ROOT/prebuilt/source-built
 find $SCRIPT_ROOT/packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 find $SCRIPT_ROOT/bin/obj/x64/Release/nuget-packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 
-# Copy reference-packages from bin dir to reference-packages directory.
+# Copy refpkgs from bin dir to refpkgs directory.
 # See corresponding change in dir.props to change ReferencePackagesBasePath conditionally in offline build.
-mkdir -p $TARBALL_ROOT/reference-packages
-cp -r $SCRIPT_ROOT/bin/obj/x64/Release/reference-packages/source $TARBALL_ROOT/reference-packages/source
-cp -r $SCRIPT_ROOT/bin/obj/x64/Release/reference-packages/staging $TARBALL_ROOT/reference-packages/staging
+mkdir -p $TARBALL_ROOT/refpkgs
+cp -r $SCRIPT_ROOT/bin/obj/x64/Release/refpkgs/src $TARBALL_ROOT/refpkgs/src
+cp -r $SCRIPT_ROOT/bin/obj/x64/Release/refpkgs/staging $TARBALL_ROOT/refpkgs/staging
 
 if [ -e $SCRIPT_ROOT/testing-smoke/smoke-test-packages ]; then
     cp -rf $SCRIPT_ROOT/testing-smoke/smoke-test-packages $TARBALL_ROOT/prebuilt
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -119,7 +119,9 @@ mkdir -p $TARBALL_ROOT/prebuilt/source-built
 find $SCRIPT_ROOT/packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 find $SCRIPT_ROOT/bin/obj/x64/Release/nuget-packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 
-cp -r $SCRIPT_ROOT/reference-packages $TARBALL_ROOT/reference-packages
+mkdir -p $TARBALL_ROOT/reference-packages
+cp -r $SCRIPT_ROOT/reference-packages/source $TARBALL_ROOT/reference-packages/source
+cp -r $SCRIPT_ROOT/reference-packages/staging $TARBALL_ROOT/reference-packages/staging
 
 if [ -e $SCRIPT_ROOT/testing-smoke/smoke-test-packages ]; then
     cp -rf $SCRIPT_ROOT/testing-smoke/smoke-test-packages $TARBALL_ROOT/prebuilt
@@ -142,6 +144,18 @@ echo 'Copying source-built packages to tarball to replace packages needed before
 for built_package in $(find $SCRIPT_ROOT/bin/obj/x64/Release/blob-feed/packages/ -name '*.nupkg')
 do
     cp $built_package $TARBALL_ROOT/prebuilt/source-built/
+done
+
+echo 'Removing reference-only packages from tarball prebuilts...'
+
+for built_package in $(find $SCRIPT_ROOT/reference-packages/packages-to-delete/ -name '*.nupkg' | tr '[:upper:]' '[:lower:]')
+do
+    if [ -e $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $built_package) ]; then
+        rm $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $built_package)
+    fi
+    if [ -e $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $built_package) ]; then
+        rm $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $built_package)
+    fi
 done
 
 echo 'WORKAROUND: Overwriting the source-built roslyn-tools MSBuild files with prebuilt so that roslyn-tools can successfully build in the tarball... (https://github.com/dotnet/source-build/issues/654)'
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -119,9 +119,11 @@ mkdir -p $TARBALL_ROOT/prebuilt/source-built
 find $SCRIPT_ROOT/packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 find $SCRIPT_ROOT/bin/obj/x64/Release/nuget-packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 
+# Copy reference-packages from bin dir to reference-packages directory.
+# See corresponding change in dir.props to change ReferencePackagesBasePath conditionally in offline build.
 mkdir -p $TARBALL_ROOT/reference-packages
-cp -r $SCRIPT_ROOT/reference-packages/source $TARBALL_ROOT/reference-packages/source
-cp -r $SCRIPT_ROOT/reference-packages/staging $TARBALL_ROOT/reference-packages/staging
+cp -r $SCRIPT_ROOT/bin/obj/x64/Release/reference-packages/source $TARBALL_ROOT/reference-packages/source
+cp -r $SCRIPT_ROOT/bin/obj/x64/Release/reference-packages/staging $TARBALL_ROOT/reference-packages/staging
 
 if [ -e $SCRIPT_ROOT/testing-smoke/smoke-test-packages ]; then
     cp -rf $SCRIPT_ROOT/testing-smoke/smoke-test-packages $TARBALL_ROOT/prebuilt
@@ -148,13 +150,10 @@ done
 
 echo 'Removing reference-only packages from tarball prebuilts...'
 
-for ref_package in $(find $SCRIPT_ROOT/reference-packages/packages-to-delete/ -name '*.nupkg' | tr '[:upper:]' '[:lower:]')
+for ref_package in $(find $SCRIPT_ROOT/bin/obj/x64/Release/reference-packages/packages-to-delete/ -name '*.nupkg' | tr '[:upper:]' '[:lower:]')
 do
     if [ -e $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $ref_package) ]; then
         rm $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $ref_package)
-    fi
-    if [ -e $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $ref_package) ]; then
-        rm $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $ref_package)
     fi
 done
 
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -119,6 +119,8 @@ mkdir -p $TARBALL_ROOT/prebuilt/source-built
 find $SCRIPT_ROOT/packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 find $SCRIPT_ROOT/bin/obj/x64/Release/nuget-packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 
+cp -r $SCRIPT_ROOT/reference-packages $TARBALL_ROOT/reference-packages
+
 if [ -e $SCRIPT_ROOT/testing-smoke/smoke-test-packages ]; then
     cp -rf $SCRIPT_ROOT/testing-smoke/smoke-test-packages $TARBALL_ROOT/prebuilt
 fi
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -119,11 +119,11 @@ mkdir -p $TARBALL_ROOT/prebuilt/source-built
 find $SCRIPT_ROOT/packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 find $SCRIPT_ROOT/bin/obj/x64/Release/nuget-packages -name '*.nupkg' -exec cp {} $TARBALL_ROOT/prebuilt/nuget-packages/ \;
 
-# Copy refpkgs from bin dir to refpkgs directory.
+# Copy reference-packages from bin dir to reference-packages directory.
 # See corresponding change in dir.props to change ReferencePackagesBasePath conditionally in offline build.
-mkdir -p $TARBALL_ROOT/refpkgs
-cp -r $SCRIPT_ROOT/bin/obj/x64/Release/refpkgs/src $TARBALL_ROOT/refpkgs/src
-cp -r $SCRIPT_ROOT/bin/obj/x64/Release/refpkgs/staging $TARBALL_ROOT/refpkgs/staging
+mkdir -p $TARBALL_ROOT/reference-packages
+cp -r $SCRIPT_ROOT/bin/obj/x64/Release/reference-packages/source $TARBALL_ROOT/reference-packages/source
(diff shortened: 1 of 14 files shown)


tools-local/init-build.proj
```diff
@@ -5,6 +5,8 @@
     <GeneratingStaticPropertiesFile>true</GeneratingStaticPropertiesFile>
   </PropertyGroup>
   <Import Project="..\dir.props" />
+  <Import Project="..\Tools\Packaging.targets" />
+  <Import Project="..\Tools\IL.targets" />
```
dagood: Use `$(ToolsDir)`?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -0,0 +1,116 @@
...
+                    
+                    // Return directories that, if containing dlls, only have dlls in the
+                    // ref folder.
+                    return
+                        !nupkgFiles.Any(file => extensionsToExclude.Contains(Path.GetExtension(file)) || pathsToExclude.Any(path => file.Contains(path))) &&
+                        nupkgFiles
+                            .Where(file => String.Equals(Path.GetExtension(file), ".dll", StringComparison.OrdinalIgnoreCase))
+                            .All(dir => dir.Contains(@"/ref/"));
+                })
+                .Select(f => Path.GetDirectoryName(f));
```
dagood: (Nit) Use method group?

dagood: Add a `ToArray()`: this is being enumerated once for the `foreach`, and again for the `Count` later.

Looks like `nupkgFiles` is in the same boat.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -0,0 +1,116 @@
...
+                    string[] extensionsToExclude = { ".exe", ".dylib", ".so", ".profdata", ".pgd" };
+                    string[] pathsToExclude = { "testdata" };
+                    
+                    // Return directories that, if containing dlls, only have dlls in the
+                    // ref folder.
+                    return
+                        !nupkgFiles.Any(file => extensionsToExclude.Contains(Path.GetExtension(file)) || pathsToExclude.Any(path => file.Contains(path))) &&
+                        nupkgFiles
+                            .Where(file => String.Equals(Path.GetExtension(file), ".dll", StringComparison.OrdinalIgnoreCase))
+                            .All(dir => dir.Contains(@"/ref/"));
```
dagood: Move into early exit `if` and/or bools with nice names to clarify the logic?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// The destination directory for the reference packages.
+        /// </summary>
+        [Required]
+        public string DestinationDir { get; set; }
+
+        public override bool Execute()
+        {
+            DateTime startTime = DateTime.Now;
+
+            var referenceOnlyPackageDirectories = Directory.EnumerateFiles(PackageCacheDir, "*.nupkg", SearchOption.AllDirectories)
```
dagood: Line length--maybe add a helper method that is always `SearchOption.AllDirectories` to make this more concise throughout. (I also wouldn't object to chopping after `Directory`.)

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -0,0 +1,116 @@
...
+                })
+                .Select(f => Path.GetDirectoryName(f));
+
+            Directory.CreateDirectory(IdentifiedPackagesDir);
+            foreach (var dir in referenceOnlyPackageDirectories)
+            {
+                Directory.CreateDirectory(dir.Replace(PackageCacheDir, DestinationDir));
+                foreach (string dirPath in Directory.EnumerateDirectories(dir, "*", SearchOption.AllDirectories))
+                {
+                    Directory.CreateDirectory(dirPath.Replace(PackageCacheDir, DestinationDir));
```
dagood: Is replicating the dir structure manually necessary? I usually just `Directory.CreateDirectory(Path.GetDirectoryName(...))` before creating new files.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -0,0 +1,116 @@
...
+                    if (file.EndsWith(".dll"))
+                    {
+                        destination = file.Replace(PackageCacheDir, DllDestinationDir);
+                        Directory.CreateDirectory(Path.GetDirectoryName(destination));
+                    }
+                    File.Copy(file, destination);
+                }
+                foreach (var file in Directory.EnumerateFiles(dir, "*.*", SearchOption.AllDirectories).Where(f => f.EndsWith(".nupkg")))
+                {
+                    File.Copy(file, IdentifiedPackagesDir);
```
dagood: Merge this foreach with the previous to do less file enumerating?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -0,0 +1,116 @@
...
+            var referenceOnlyPackageDirectories = Directory.EnumerateFiles(PackageCacheDir, "*.nupkg", SearchOption.AllDirectories)
+                .Where(nupkgFilePath =>
+                {
+                    // Get all files in the nupkg path and normalize directory separators
+                    var nupkgFiles = Directory.EnumerateFiles(Path.GetDirectoryName(nupkgFilePath), "*.*", SearchOption.AllDirectories).Select(f => f.Replace("\\","/"));
+
+                    // Do not include directories that contain exes, shared object files, OSX dynamic libraries
+                    // or profiling data
+                    string[] extensionsToExclude = { ".exe", ".dylib", ".so", ".profdata", ".pgd" };
+                    string[] pathsToExclude = { "testdata" };
```
dagood: Move to static?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -0,0 +1,116 @@
...
+                            .Where(file => String.Equals(Path.GetExtension(file), ".dll", StringComparison.OrdinalIgnoreCase))
+                            .All(dir => dir.Contains(@"/ref/"));
+                })
+                .Select(f => Path.GetDirectoryName(f));
+
+            Directory.CreateDirectory(IdentifiedPackagesDir);
+            foreach (var dir in referenceOnlyPackageDirectories)
+            {
+                Directory.CreateDirectory(dir.Replace(PackageCacheDir, DestinationDir));
+                foreach (string dirPath in Directory.EnumerateDirectories(dir, "*", SearchOption.AllDirectories))
```
dagood: Could you save the enumeration results used to determine that this is a ref-only package and use it here? (Keep it in an anonymous type?)

build-source-tarball.sh
```diff
@@ -137,6 +141,18 @@ do
...
 
+echo 'Removing reference-only packages from tarball prebuilts...'
+
+for ref_package in $(find $SCRIPT_ROOT/reference-packages/packages-to-delete/ -name '*.nupkg' | tr '[:upper:]' '[:lower:]')
+do
+    if [ -e $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $ref_package) ]; then
+        rm $TARBALL_ROOT/prebuilt/nuget-packages/$(basename $ref_package)
+    fi
+    if [ -e $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $ref_package) ]; then
+        rm $TARBALL_ROOT/prebuilt/smoke-test-packages/$(basename $ref_package)
```
dagood: I don't think this is right--these aren't included in the tarball that we give Red Hat, and it's only a way to transport packages that aren't available from any online source at that point for testing purposes. They're actually not included in the tarball.

Core problem is that they're called "prebuilts", come to think of it, when they aren't really.

dir.props
```diff
@@ -19,6 +19,10 @@
     <DotNetCliToolDir>$(ProjectDir)Tools/dotnetcli/</DotNetCliToolDir>
     <PatchesDir>$(ProjectDir)patches/</PatchesDir>
     <PackagesDir>$(ProjectDir)packages/</PackagesDir>
+    <ReferencePackagesStagingDir>$(ProjectDir)reference-packages/staging/</ReferencePackagesStagingDir>
```
dagood: I think the tarball production build should put this stuff in `bin` if possible.

patches/newtonsoft-json/0001-Run-dotnet-migrate-to-generate-Newtonsoft.Json.Dotne.patch
```diff
@@ -53,6 +53,7 @@ index 0000000..797f637
 +    <Reference Include="System.Xml.Linq" />
 +    <Reference Include="System" />
 +    <Reference Include="Microsoft.CSharp" />
++    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0-alpha-003" PrivateAssets="all"/>
```
crummel: Does this change introduce building Newtonsoft for full-framework or is this fixing a separate issue?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public override bool Execute()
+        {
+            DateTime startTime = DateTime.Now;
+
+            var referenceOnlyPackageDirectories = Directory.EnumerateFiles(PackageCacheDir, "*.nupkg", SearchOption.AllDirectories)
+                .Where(nupkgFilePath =>
+                {
+                    // Get all files in the nupkg path and normalize directory separators
+                    var nupkgFiles = Directory.EnumerateFiles(Path.GetDirectoryName(nupkgFilePath), "*.*", SearchOption.AllDirectories).Select(f => f.Replace("\\","/"));
```
crummel: Is `f.Replace("\\","/")` necessary?  I would have thought `EnumerateFiles` would use a platform-appropriate path separator.  If not, should we be replacing both `/` and `\` with `Path.DirectorySeparator`?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
crummel: Just to make sure I understand: If a package includes DLLs only in a `ref` folder, but also EXEs in (say) the `content` folder, this is considered a ref package, and we only include the DLLs in the version of the package that we end up generating?

build.proj
```diff
@@ -28,6 +30,43 @@
...
+         They'll end up in the source directory as dlls for now 
+
+        Keep the following ItemGroup when workaround is removed:
+    <ItemGroup>
+        <Dlls Include="$(ReferencePackagesSourceDir)/**/*.dll" />
+    </ItemGroup>
+    -->
+
+    <ItemGroup>
+        <Dlls Include="$(ReferencePackagesSourceDir)/**/*.dll" Exclude="$(ReferencePackagesSourceDir)/**/System.Runtime.dll;$(ReferencePackagesSourceDir)/**/netstandard.dll" />
```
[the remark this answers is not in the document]
dseefeld: Instead of Exclude, I used Remove for just the two dlls in the workaround.

patches/newtonsoft-json/0001-Run-dotnet-migrate-to-generate-Newtonsoft.Json.Dotne.patch
dseefeld: It re-ordered something so that another repo is now using the version that is source-built and was complaining about not being able to build for net45.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
dseefeld: Yeah.  I was doing this because I was checking for /ref/ below in the path.  I modified this to use Path.DirectorySeparatorChar and removed the unnecessary Replace().

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
dseefeld: No.  I exclude packages that have other binary types in the extensionsToExclude list above.

dseefeld:
@dotnet-bot test this please

dseefeld:
@dotnet-bot test this please

dseefeld:
@dotnet-bot test this please

dseefeld:
@crummel , @dagood - This should be complete now, if you want to give it another look.

build.proj
```diff
@@ -30,6 +31,61 @@
...
+        DestinationDir="$(ReferencePackagesStagingDir)"
+        />
+
+    <!-- Ildasm has trouble writing to paths that are longer than 260 chars. (see https://github.com/dotnet/coreclr/issues/20397)
+         Create a temporary path to write il output which will then be copied back to the reference-packages/source dir.
+    -->
+    <PropertyGroup>
+      <TempPath>$([System.IO.Path]::GetTempPath())</TempPath>
+      <IldasmTempOutputFolderName>$([System.IO.Path]::GetRandomFileName())</IldasmTempOutputFolderName>
+      <IldasmTempOutputPath>$([System.IO.Path]::Combine($(TempPath),$(IldasmTempOutputFolderName)))/</IldasmTempOutputPath>
```
dagood: I feel like it would be good manners to try to delete this folder when we're done, but I'm not sure... what do you think?

build.proj
```diff
@@ -30,6 +31,61 @@
...
+    </PropertyGroup>
+    
+    <ItemGroup>
+      <ReferenceOnlyPackages Include="$(ReferencePackagesToDeleteDir)/**/*.nupkg" />
+      <ReferenceOnlyPackageDlls Include="$(ReferencePackagesSourceDir)/**/*.dll" />
```
dagood: `/` after the `Dir` property is redundant. (Also in next itemgroup.)

build.proj
```diff
@@ -30,6 +31,61 @@
...
+    </ItemGroup>
+
+    <!-- Temporary workaround to exclude dlls that can't be round-tripped. See: https://github.com/dotnet/coreclr/issues/20262
+         They'll end up in the source directory as dlls for now 
+    -->
+    <ItemGroup>
+      <ReferenceOnlyPackageDlls Remove="$(ReferencePackagesSourceDir)/**/System.Runtime.dll;$(ReferencePackagesSourceDir)/**/netstandard.dll" />
```
dagood: (Nit) Item elements that include multiple files are clearer to me when there's a newline after the `;` or they're just done with multiple elements.

dir.props
```diff
@@ -80,6 +81,17 @@
     <ProdConManifestFile>$(PackageReportDir)prodcon-build.xml</ProdConManifestFile>
     <PoisonedReportFile>$(PackageReportDir)poisoned.txt</PoisonedReportFile>
     <ConflictingPackageReportDir>$(BaseOutputPath)conflict-report/</ConflictingPackageReportDir>
+    <ReferencePackagesBasePath>$(IntermediatePath)</ReferencePackagesBasePath>
+    <!--
+      Change ReferencePackagesBasePath conditionally in offline build.
+      See corresponding change in build-source-tarball.sh to copy reference-packages
+      from source-build bin dir to tarball reference-packages dir.
+    -->
+    <ReferencePackagesBasePath Condition="'$(OfflineBuild)' == 'true'">$(ProjectDir)</ReferencePackagesBasePath>
```
dagood: I think this would be clearer and remove some duplication:

```xml
    <ReferencePackagesBasePath>$(IntermediatePath)reference-packages/</ReferencePackagesBasePath>
    <!--
      Change ReferencePackagesBasePath conditionally in offline build.
      See corresponding change in build-source-tarball.sh to copy reference-packages
      from source-build bin dir to tarball reference-packages dir.
    -->
    <ReferencePackagesBasePath Condition="'$(OfflineBuild)' == 'true'">$(ProjectDir)reference-packages/</ReferencePackagesBasePath>
    <ReferencePackagesStagingDir>$(ReferencePackagesBasePath)staging/</ReferencePackagesStagingDir>
```

Technically duplicating the dir name, but I think it's fine and makes it clear what the prop's pointing to in context.

(I'd also call it `ReferencePackagesBaseDir` rather than `Path`. 😄)

patches/linker/0002-Update-linker-to-use-2.0.0-runtime.patch
```diff
@@ -0,0 +1,25 @@
+From c884d2eeacddbb3a09b8b2bac3c132ff3b3b63e4 Mon Sep 17 00:00:00 2001
+From: dseefeld <dseefeld@microsoft.com>
+Date: Mon, 15 Oct 2018 19:18:18 +0000
+Subject: [PATCH] Update linker to use 2.0.0 runtime
+
```
dagood: Can you add something to the body about why this patch was added?

I am curious though if we could pass `/p:RuntimeFrameworkVersion=2.0.0` rather than using a patch for this, since IIRC properties set that way can't be changed... I don't mind the patch though.

tools-local/init-build.proj
```diff
@@ -67,6 +69,31 @@
      <Exec Condition="'$(Platform)' == 'armel'" Command="$(ArmEnvironmentVariables) $(ProjectDir)cross/armel/tizen-build-rootfs.sh" />
   </Target>
 
+  <Target Name="BuildReferenceAssemblies" Condition="'$(OfflineBuild)' == 'true'">
+    <ItemGroup>
+      <IlSource Include="$(ReferencePackagesSourceDir)/**/*.il" />
+      <ReferenceAssemblyDirectories Include ="$(ReferencePackagesStagingDir)/**/*.nuspec" />
+    </ItemGroup>
+
+    <MakeDir Directories="$(ReferencePackagesStagingDir)%(IlSource.RecursiveDir)" Condition="!Exists('$(ReferencePackagesStagingDir)%(IlSource.RecursiveDir)')" />
```
dagood: Is the condition necessary? I thought it just kept going if it exists.

tools-local/init-build.proj
```diff
@@ -67,6 +69,31 @@
...
+    <Exec Command="$(IlasmToolPath) %(IlSource.Identity) -dll -quiet -nologo -output=$(ReferencePackagesStagingDir)%(IlSource.RecursiveDir)%(IlSource.Filename).dll" />
```
dagood: Include timestamped messages in case this takes a while?

tools-local/init-build.proj
```diff
@@ -67,6 +69,31 @@
...
+
+    <!-- Temporary workaround to copy dlls that can't be round-tripped 
+         They just get copied from the source to the package staging area
+    -->
+    <ItemGroup>
+        <DllsInSource Include="$(ReferencePackagesSourceDir)/**/*.dll" />
```
dagood: (Nit) Too many spaces

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -0,0 +1,132 @@
...
+                            .Where(file => String.Equals(Path.GetExtension(file), ".dll", StringComparison.OrdinalIgnoreCase))
+                            .All(dir => dir.Contains(refPath));
+                })
+                .Select(Path.GetDirectoryName)
+                .ToArray();
+
+            Directory.CreateDirectory(IdentifiedPackagesDir);
+            foreach (var dir in referenceOnlyPackageDirectories)
+            {
+                foreach (var file in EnumerateAllFiles(dir, "*.*"))
```
dagood: In the earlier iteration I suggested keeping the enumeration results from `nupkgFiles` rather than repeating it here potentially using an anonymous type--understandable if it's not worth it and that's why it's marked resolved, but pointing out it's still here.

build.proj
dseefeld: Agreed. Fixed.

dir.props
dseefeld: Agreed. Thanks.

patches/linker/0002-Update-linker-to-use-2.0.0-runtime.patch
dseefeld: I'll update the patch.

tools-local/init-build.proj
dseefeld: I'll try without.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
dseefeld: I think I'll leave this.  I'm not sure it's worth changing things around at this point.

dseefeld:
@dotnet-bot test this please

dseefeld:
source-build-CI isn't meant to work, yet.  OK to merge.

Issue #514, opened 2018-05-04
Re-enable PGO by default


crummel:
See https://github.com/dotnet/source-build/issues/512 and https://github.com/dotnet/source-build/pull/513.  Once we have an environment set up that properly does PGO, we should remove https://github.com/dotnet/source-build/pull/513/files#diff-03e2fc79096a19a506e5d2bf1dab3d99R6 or at least make PGO on by default.

crummel:
According to [LLVM](https://llvm.org/docs/GoldPlugin.html#lto-how-to-build), PGO requires LTO which requires LLVMgold which requires `ld` plugin support.  On our CI docker image I get:
```
[root@77810b22880b /]# ld -plugin
ld: unrecognized option '-plugin'
```
which indicates no plugin support, vs my working PGO machine with
```
[chris@crummel-rh2 ~]$ ld -plugin
ld: -plugin: missing argument
```

So that seems to be the root cause of PGO not running on the CI image at all.

crummel:
This is blocked by https://github.com/dotnet/source-build/issues/247.

PR #802, opened 2018-10-08
Set the build-rootfs like the other repos

deinok:
This mirrors the build-rootfs.sh from coreclr, also added a patch


crummel:
Hi @deinok, thanks for the patch!  Can I ask how you're testing this?  It looks like our Ubuntu ARM build is having issues (#811 and possibly #514) so I was going to try it out on Alpine.

deinok:
Hi @crummel , I have just copied the coreclr version and test it with x86.
For what I expect the cross folder should be a mirror of coreclr cross folder

crummel:
Ok, I see on the CoreCLR PR that Jan said these should all be in sync, that looks good.

crummel:
@dotnet-bot test this please.

Issue #757, opened 2018-09-07
Use VSTS dnceng instance for public CI


dagood:
We have YAML definitions for our internal builds, so we should be able to reuse it for public CI as well. See https://github.com/dotnet/arcade/blob/master/Documentation/VSTS/VSTSGuidance.md.

There isn't as much variety of "real" machines in VSTS yet, so we'd be using Docker containers for everything at that point. Our YAML defs are set up for that.

dagood:
FYI @chcosta, if you'd like to track this one too.

dagood:
Blocked on MacOS pool being fixed to be capable of building CoreCLR: https://github.com/dotnet/core-eng/issues/4210.

dagood:
I was able to patch CoreCLR to build on the existing MacOS pool just fine: https://github.com/dotnet/coreclr/issues/20092.

Now, we need more disk space on the agent. An effort to enable that is tracked at https://github.com/dotnet/core-eng/issues/4286.

dagood:
Merged with https://github.com/dotnet/source-build/pull/813, spawned new issues to follow up on: #817 #814 #815

PR #805, opened 2018-10-09
Use hosted pool

chcosta:
Testing if the docker hosted pool works for source-build

FYI @riarenas


chcosta:
Doesn't look like Azure DevOps has a build def for source-build master.  I'll look into validating this another way.

dagood:
Yeah, sorry, dnceng AzDO CI isn't implemented yet for this repo. This YAML is currently used in the devdiv instance. Using dnceng AzDO is tracked by https://github.com/dotnet/source-build/issues/757.

PR #806, opened 2018-10-12
Cross for x86

deinok:
This should enable the usage of cross for x86

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -81,7 +81,7 @@
   <PropertyGroup>
      <ArmEnvironmentVariables Condition="'$(ArmEnvironmentVariables)' == ''">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/arm</ArmEnvironmentVariables>
      <ArmEnvironmentVariables Condition="'$(Platform)' == 'armel'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/armel</ArmEnvironmentVariables>
-     <CrossRootFsEnvironmentVariables Condition="'$(Platform)' == 'x86'">ROOTFS_DIR=$(BaseIntermediatePath)cross/rootfs/x86</CrossRootFsEnvironmentVariables> 
+     <CrossRootFsEnvironmentVariables Condition="'$(Platform)' == 'x86'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/x86</CrossRootFsEnvironmentVariables> 
   </PropertyGroup>
 
   <Import Project="$(TargetInfoProps)" Condition="$(GeneratingStaticPropertiesFile) != 'true' AND Exists('$(TargetInfoProps)')" />
--- a/dir.props
+++ b/dir.props
@@ -81,7 +81,6 @@
   <PropertyGroup>
      <ArmEnvironmentVariables Condition="'$(ArmEnvironmentVariables)' == ''">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/arm</ArmEnvironmentVariables>
      <ArmEnvironmentVariables Condition="'$(Platform)' == 'armel'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/armel</ArmEnvironmentVariables>
-     <CrossRootFsEnvironmentVariables Condition="'$(CrossRootFsEnvironmentVariables)' == ''">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/x86</CrossRootFsEnvironmentVariables> 
      <CrossRootFsEnvironmentVariables Condition="'$(Platform)' == 'x86'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/x86</CrossRootFsEnvironmentVariables> 
   </PropertyGroup>
 
--- a/dir.props
+++ b/dir.props
@@ -79,9 +79,9 @@
   <Import Project="$(ToolsDir)Build.Common.props" Condition="Exists('$(ToolsDir)Build.Common.props')" />
 
   <PropertyGroup>
-     <ArmEnvironmentVariables Condition="'$(ArmEnvironmentVariables)' == ''">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/arm</ArmEnvironmentVariables>
-     <ArmEnvironmentVariables Condition="'$(Platform)' == 'armel'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/armel</ArmEnvironmentVariables>
-     <CrossRootFsEnvironmentVariables Condition="'$(Platform)' == 'x86'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/x86</CrossRootFsEnvironmentVariables> 
+     <RootFsEnvironmentVariables Condition="'$(RootFsEnvironmentVariables)' == ''">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/arm</RootFsEnvironmentVariables>
+     <RootFsEnvironmentVariables Condition="'$(Platform)' == 'armel'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/armel</RootFsEnvironmentVariables>
+     <RootFsEnvironmentVariables Condition="'$(Platform)' == 'x86'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/x86</RootFsEnvironmentVariables> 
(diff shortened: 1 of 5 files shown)


dir.props
```diff
@@ -81,7 +81,8 @@
   <PropertyGroup>
      <ArmEnvironmentVariables Condition="'$(ArmEnvironmentVariables)' == ''">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/arm</ArmEnvironmentVariables>
      <ArmEnvironmentVariables Condition="'$(Platform)' == 'armel'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/armel</ArmEnvironmentVariables>
-   </PropertyGroup>
+     <CrossRootFsEnvironmentVariables Condition="'$(Platform)' == 'x86'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/x86</CrossRootFsEnvironmentVariables>
```
dagood: It seems like it would be cleaner to rename all three of these to `RootFsEnvironmentVariables` so `ROOTFS_DIR` can be set more uniformly in all the build commands.

repos/core-setup.proj
```diff
@@ -4,8 +4,10 @@
   <PropertyGroup>
     <BuildArguments>-ConfigurationGroup=$(Configuration) -PortableBuild=$(PortableBuild) -SkipTests=true </BuildArguments>
     <BuildArguments Condition="$(Platform.Contains('arm'))">$(BuildArguments) -TargetArchitecture=$(Platform) -DisableCrossgen=true -CrossBuild=true</BuildArguments>
+    <BuildArguments Condition="'$(OS)' != 'Windows_NT' AND '$(Platform)' == 'x86'">$(BuildArguments) -TargetArchitecture=$(Platform) -DisableCrossgen=true -CrossBuild=true</BuildArguments>
```
dagood: I'm not a fan of the duplication here... could the `-TargetArchitecture=$(Platform) -DisableCrossgen=true -CrossBuild=true` be moved to a property? (The conditions could alternatively be combined, but that would make a fairly long line which I'm not a fan of.)

repos/core-setup.proj
```diff
@@ -4,8 +4,10 @@
...
     <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) $(BuildArguments) -- /p:BuildDebPackage=false /p:BuildAllPackages=true /p:PreReleaseLabel="preview"</BuildCommand>
     <BuildCommand Condition="$(Platform.Contains('arm'))">$(ArmEnvironmentVariables) $(BuildCommand)</BuildCommand>
+    <BuildCommand Condition="'$(OS)' != 'Windows_NT' AND '$(Platform)' == 'x86'">$(CrossRootFsEnvironmentVariables) $(BuildCommand)</BuildCommand>
```
dagood: I imagine these two lines could be made like this with the `RootFsEnvironmentVariables` rename...

```xml
<BuildCommand Condition="'$(RootFsEnvironmentVariables)' != ''">$(RootFsEnvironmentVariables) $(BuildCommand)</BuildCommand>
```

I'm not familiar with this scenario, so maybe there's some reason that doesn't work out I'm not aware of. Using the existence of `RootFsEnvironmentVariables` rather than checking conditions again seems nice to me though.

repos/coreclr.proj
```diff
@@ -7,9 +7,11 @@
     <BuildArguments Condition="'$(OS)' != 'Windows_NT'">$(BuildArguments) msbuildonunsupportedplatform</BuildArguments>
     <BuildArguments Condition="'$(UseSystemLibraries)' == 'true'">$(BuildArguments) cmakeargs -DCLR_CMAKE_USE_SYSTEM_LIBUNWIND=TRUE</BuildArguments>
     <BuildArguments Condition="$(Platform.Contains('arm'))">$(BuildArguments) skipnuget cross -skiprestore cmakeargs -DFEATURE_GDBJIT=TRUE</BuildArguments>
+    <BuildArguments Condition="'$(OS)' != 'Windows_NT' AND '$(Platform)' == 'x86'">$(BuildArguments) skipnuget cross -skiprestore cmakeargs -DFEATURE_GDBJIT=TRUE</BuildArguments>
```
dagood: Same comments from core-setup apply here to coreclr, and corefx too.

dir.props
deinok: Yep, I have the same idea but didn't want to rename things. 
So set everything as `<RootFsEnvironmentVariables>` ?

dagood: Doing a search, I don't see any places a rename would affect other than the parts you're already making changes near, so I'd say it'd be good to make this refactor now. Without CI testing x86 or ARM it's hard to know for sure how fragile this is, though....

I'm ok leaving it like you have it now. The other comments (`BuildArguments` duplication) seem to have a bigger impact on maintainability, anyway.

deinok: I have renamed all cross related EnvironmentVariables to RootFsEnvironmentVariables
commit: #9f1a856

repos/core-setup.proj
deinok: Yes, this seems a better way

repos/coreclr.proj
deinok: Here, I have duplicated the code because imagine that following:
x86 should generate nuget while for arm we have to skipnuget.

This way we can have a more granular way to manage the BuildArguments for OS-ARCH

repos/core-setup.proj
deinok: Here, I have duplicated the code because imagine that following:
x86 should enable CrossGen while for arm we have to set Crossgen.

This way we can have a more granular way to manage the BuildArguments for OS-ARCH

dir.props
```diff
@@ -79,9 +79,10 @@
   <Import Project="$(ToolsDir)Build.Common.props" Condition="Exists('$(ToolsDir)Build.Common.props')" />
 
   <PropertyGroup>
-     <ArmEnvironmentVariables Condition="'$(ArmEnvironmentVariables)' == ''">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/arm</ArmEnvironmentVariables>
-     <ArmEnvironmentVariables Condition="'$(Platform)' == 'armel'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/armel</ArmEnvironmentVariables>
-   </PropertyGroup>
+     <RootFsEnvironmentVariables Condition="'$(RootFsEnvironmentVariables)' == ''">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/arm</RootFsEnvironmentVariables>
+     <RootFsEnvironmentVariables Condition="'$(Platform)' == 'armel'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/armel</RootFsEnvironmentVariables>
+     <RootFsEnvironmentVariables Condition="'$(Platform)' == 'x86'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/x86</RootFsEnvironmentVariables>
```
dagood: With the other change to `<BuildCommand Condition="'$(RootFsEnvironmentVariables)' != ''">$(RootFsEnvironmentVariables)`, I think you'll have to change these to make sure `RootFsEnvironmentVariables` is only defined on platforms that need rootfs. This is what comes to mind but I'm not sure:

```xml
  <PropertyGroup Condition="'$(RootFsEnvironmentVariables)' == ''">
     <RootFsEnvironmentVariables Condition="$(Platform.Contains('arm'))">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/arm</RootFsEnvironmentVariables>
     <RootFsEnvironmentVariables Condition="'$(Platform)' == 'armel'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/armel</RootFsEnvironmentVariables>
     <RootFsEnvironmentVariables Condition="'$(OS)' != 'Windows_NT' AND '$(Platform)' == 'x86'">ROOTFS_DIR=$(BaseIntermediatePath)crossrootfs/x86</RootFsEnvironmentVariables> 
  </PropertyGroup>
```

repos/core-setup.proj
dagood: I think that calls for extracting the parameters into variables, not duplication. And even then, can't it wait until differences actually exist so we know for sure it's arranged in a useful way?

dir.props
deinok: LGTM

crummel:
Closing and reopening to get new CI.

deinok:
I had a script to update the gitsubmodules every day, but seems like it have corrupted the git info.
I will redone this week

PR #807, opened 2018-10-12
Add leak detection to CentOS builds.

crummel:
Set up CI to use leak detection is the CentOS tarball and offline builds.  Getting this in before the actual leak detection stuff so we can be sure everything works all the way through.

Skip CI please.


crummel:
@dotnet-bot test CI please.

crummel:
@dotnet-bot test CI please.

PR #826, opened 2018-10-22
 Include Azure DevOps badges; generate README table

dagood:
Clean view: https://github.com/dagood/source-build/blob/azdo-badges/README.md

Add the Azure DevOps badges in a new column. Note that the badges are externally visible but the builds the badges link to unfortunately aren't. The build is internal to allow us to work around [a feature gap between Jenkins and AzDO](https://dnceng.visualstudio.com/internal/_workitems/edit/92) by publishing a file to an authenticated endpoint during the build. (Workaround not yet implemented.)

Add a `sh` script to generate the table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,41 +2,26 @@
 
 [![Join the chat at https://gitter.im/dotnet/source-build](https://badges.gitter.im/dotnet/source-build.svg)](https://gitter.im/dotnet/source-build?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-|OS|Release|Debug|
-|--|-------|-----|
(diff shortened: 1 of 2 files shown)


dagood:
OSX failed in Azure DevOps CI. Normally, our `BUILD_BUILDNUMBER` is something like `20181022.1`. Roslyn-tools tries to `split('.')[1]`. The currently-linked build is a retry, which for some reason has the `BUILD_BUILDNUMBER=33769`. Not able to find a `.`, the build fails. (This only shows up in OSX because it's the only place we use the provided shell's env vars. All other legs run in docker containers without any env vars flowing in.)

Our YAML doesn't specify a build number schema. [AzDO is supposed to give us a single unique integer](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/options?view=vsts&tabs=yaml#build-number-format) (like `33769`) when this is the case, so I guess `20181022.1` was the buggy one from AzDO's perspective. I'll submit a followup PR to solidify the `.` format and see if that works.

For now, this PR is ready to merge--[the first AzDO CI build succeeded](https://dnceng.visualstudio.com/public/_build/results?buildId=33350&_a=summary&view=results). (Not sure why I hit retry.)

dagood:
@dseefeld @crummel PTAL

PR #833, opened 2018-10-24
Add build format to YAML

dagood:
This might fix the build number on rebuilds, and if not, it clarifies the scope of the AzDO bug.


dagood:
Doesn't work. We don't have any reason to specify build format because the default works fine (in non-rebuild cases) so I'm closing this PR.

Issue #822, opened 2018-10-18
release/2.2 source-built SDK includes prebuilt binaries from Microsoft.NET.Sdk.Razor preview3


dagood:
https://github.com/dotnet/source-build/pull/821 (2.2.0-preview3) upgraded Microsoft.NET.Sdk.Razor from `2.2.0-preview2-35157` to `2.2.0-preview3-35497`.

Microsoft.NET.Sdk.Razor is built by ASP.NET Core, and the CLI repo uses it as a prebuilt. The `preview2` version was text-only and an acceptable prebuilt, but `preview3` added some task DLLs. These do get into the source-built CLI output:

```
./sdk/2.2.100/Sdks/Microsoft.NET.Sdk.Razor/tasks/net46/Microsoft.NET.Sdk.Razor.Tasks.dll
./sdk/2.2.100/Sdks/Microsoft.NET.Sdk.Razor/tasks/netstandard2.0/Microsoft.NET.Sdk.Razor.Tasks.dll
```

This Microsoft.NET.Sdk.Razor package [*is* on nuget.org](https://www.nuget.org/packages/Microsoft.NET.Sdk.Razor/2.2.0-preview3-35497)… so maybe we can just force CLI not to bundle it while source-building and let the MSBuild NuGet SDK resolver fetch it from the network on demand. I don't know one way or the other whether the mechanism will work as-is.

/cc @dleeapho @dseefeld @crummel

crummel:
I'm getting
```
/home/chris/testing/sdk/2.2.100/Sdks/Microsoft.NET.Sdk.Web/Sdk/Sdk.targets(16,11): error MSB4236: The SDK 'Microsoft.NET.Sdk.Razor' specified could not be found. [/home/chris/testing/app/app.csproj]

The build failed. Please fix the build errors and run again.
```
when I try to run a Razor project so I don't think the easy fix is enough.  Looking into a smarter fix now.

crummel:
Based on the comments in https://github.com/aspnet/Razor/issues/2553 I'm not sure that any fix that doesn't include the binaries in this package is really valid - it seems like we would just bring that issue back for source-build users.  I tried added the Razor repo to source-build but it doesn't seem like an easy repo; we might have to depend on ASP.NET being source-buildable in general.

natemcmaster:
> Is there a way to defer acquiring the Microsoft.NET.Sdk.Razor package until restore/run?

No. This is in-box with Microsoft.NET.Sdk.Web by design.

[...]

PR #834, opened 2018-10-24
[release/2.1] Add aspnet/Razor submodule for Microsoft.NET.Sdk.Razor

dagood:
Add https://github.com/aspnet/Razor to source-build to produce `Microsoft.NET.Sdk.Razor 2.1.2`.

For 2.2, after the auto-merge, I'll apply https://github.com/dotnet/source-build/compare/be27588b9c51c8c9bef9ef60f329eef41cf868f1...dagood:add-razor-sdk-2.2 to get the DLLs building. I decompiled the source-built DLLs and they're identical to the ones on the gallery. Doing a local build of those changes and this 2.1 PR to validate prebuilt baselines.

Once the aspnet repos are merged into aspnet/AspNetCore (not as submodules) these are the steps I'm thinking we'll need to do:
* Remove `src/aspnet-razor`, add `src/aspnetcore`.
* Rename project file to match.
  * Update base path.
  * Add handling of the global.json (present in AspNetCore but not Razor).
* Update patches for new paths.
  * (We actually don't have a mechanism to patch a submodule of a submodule, so we can't go ahead and use AspNetCore right now. The large number of submodules it would add before consolidation is a hindrance for some dev scenarios, too.)

FYI @natemcmaster 

Part of https://github.com/dotnet/source-build/issues/822

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -66,3 +66,6 @@
 [submodule "src/linker"]
 	path = src/linker
 	url = https://github.com/mono/linker.git
+[submodule "src/aspnet-razor"]
+	path = src/aspnet-razor
+	url = https://github.com/aspnet/Razor
--- a/repos/aspnet-razor.proj
+++ b/repos/aspnet-razor.proj
@@ -10,8 +10,7 @@
     <BuildCommandArgs>pack</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) $(RazorSdkProjectFile)</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) /p:Configuration=$(Configuration)</BuildCommandArgs>
-    <!-- A nonfinal build adds "-t000" to the package version. (Prerelease builds are "final".) -->
-    <BuildCommandArgs>$(BuildCommandArgs) /p:IsFinalBuild=true</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:IsFinalBuild=$(UseStableVersions)</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) /p:BuildNumber=$(BuildNumber)</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) /bl</BuildCommandArgs>
 
--- a/repos/aspnet-razor.proj
+++ b/repos/aspnet-razor.proj
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+  <PropertyGroup>
+    <RazorSdkProjectDir>$(ProjectDirectory)src/Microsoft.NET.Sdk.Razor/</RazorSdkProjectDir>
+    <RazorSdkProjectFile>$(RazorSdkProjectDir)Microsoft.NET.Sdk.Razor.csproj</RazorSdkProjectFile>
+
+    <BuildNumber>30932</BuildNumber>
+
+    <BuildCommandArgs>pack</BuildCommandArgs>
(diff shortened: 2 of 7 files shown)


repos/aspnet-razor.proj
```diff
@@ -0,0 +1,28 @@
...
+    <RazorSdkProjectDir>$(ProjectDirectory)src/Microsoft.NET.Sdk.Razor/</RazorSdkProjectDir>
+    <RazorSdkProjectFile>$(RazorSdkProjectDir)Microsoft.NET.Sdk.Razor.csproj</RazorSdkProjectFile>
+
+    <BuildNumber>30932</BuildNumber>
+
+    <BuildCommandArgs>pack</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) $(RazorSdkProjectFile)</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:Configuration=$(Configuration)</BuildCommandArgs>
+    <!-- A nonfinal build adds "-t000" to the package version. (Prerelease builds are "final".) -->
+    <BuildCommandArgs>$(BuildCommandArgs) /p:IsFinalBuild=true</BuildCommandArgs>
```
crummel: Should this be equivalent to one of our `DropSuffix` or `Stabilized` properties?

dagood: My intent behind the comment is to explain why it shouldn't be. 😄 (Specifically, `UseStableVersions` in dependencies.props.) I'll make this clearer.

.gitmodules
```diff
@@ -66,3 +66,6 @@
 [submodule "src/linker"]
 	path = src/linker
 	url = https://github.com/mono/linker.git
+[submodule "src/aspnet-razor"]
+	path = src/aspnet-razor
+	url = https://github.com/aspnet/Razor
```
natemcmaster:
```suggestion
	url = https://github.com/aspnet/Razor
	branch = release/2.1
```

repos/aspnet-razor.proj
dagood: Wait a sec... there's an issue for 2.2 where I ended up building `2.2.0-preview3-final` even though ProdCon built `2.2.0-preview3-35497`. I talked with @natemcmaster about it, who said `dotnet pack /p:BuildNumber=xyz /p:IsFinalBuild=true` (like it does now) should include the build number, but my 2.2 build just got to the point where I can see this still produces `2.2.0-preview3-final`.

Looking closer at version.props, it looks like `IsFinalBuild` should actually be `false`. Then, `PackageVersion` won't be set, and the SDK is free to combine `VersionPrefix` and `VersionSuffix`.

.gitmodules
dagood: AFAIK the only thing this allows is `git submodule update --remote`, and I don't think anyone uses that for source-build. From an infra perspective this is just another value to maintain and resolve conflicts for.

If someone made a case for adding these (to all the submodules that have applicable branches), I could get behind that, but for now I want to leave it out.

repos/aspnet-razor.proj
natemcmaster: For 2.2 we stopped doing the `IsFinalBuild=true` step for previews, as you'll see on nuget.org: https://www.nuget.org/packages/Microsoft.NET.Sdk.Razor/2.2.0-preview3-35497.

But this PR is targeting 2.1, right? So, `IsFinalBuild=true` should be set for an RTM build of 2.1.x

dagood:
[Machine ran out of space.](https://ci.dot.net/job/dotnet_source-build/job/release_2.1/job/OSX10.12_Release_prtest/219/)

dagood:
@dotnet-bot test CentOS7.1 Tarball Release
@dotnet-bot test RHEL7.2 Tarball Release
Straight-up timeouts.


I'm planning to merge regardless of source-build-CI if the Jenkins checks work. I'm working on getting the `source-build-CI` leg to re*try*, but that's blocked by an AzDO bug tracked at https://dnceng.visualstudio.com/internal/_workitems/edit/105. Re*build* is also blocked by a different AzDO bug. Only the `centos71 Offline` leg failed due to `Child node "2" exited prematurely` (somewhat known flakiness) so I'm not concerned.

(The MSBuild premature exit says it logged out some debug data, and I've merged https://github.com/dotnet/source-build/pull/836 to catch it next time. I'm skeptical that it actually wrote anything out, since it happened in Jenkins before and even though it's configured to save those files, it didn't find any.)

PR #841, opened 2018-10-29
Make AzDO badge links specific to branch

dagood:
Now that we have release/2.1 and release/2.2 rolling builds running on AzDO, we can see that this works:

https://dev.azure.com/dnceng/internal/_build/latest?definitionId=114&branchName=release/2.1
https://dev.azure.com/dnceng/internal/_build/latest?definitionId=114&branchName=release/2.2

(Badges are still more specific than the links, going down to phase/matrix leg, where these links just show the whole build without automatically opening the right logs.)


dagood:
Jenkins OSX failed with the usual java.nio.channels.ClosedChannelException, but AzDO OSX was fine. Merging.

am11:
Unlike old Jenkins ones, Azure Pipelines links are 401 for non-MSFT folks. Would we be allowed to access those links sometime in future?

dagood:
There are two projects in the dnceng AzDO instance, `public` and `internal`. CI runs in `public`, so you can see everything when you click on an AzDO PR status, but the rolling build happens in `internal`. There are no plans to give non-Msft folks access to `internal`, since we do signing in there and builds can have access to secrets.

On the other hand, source-build could switch rolling builds to the `public` project. However, that isn't easy due to an AzDO feature gap: https://github.com/dotnet/source-build/issues/851.

As a workaround, we considered running the build twice (once in `public` and once in `internal`) but that means twice the builds to try to keep track of and keep green, so we aren't doing that now.

PR #842, opened 2018-10-30
Update README.md

regisbsb:
Misspelt reproducibility


crummel:
Doc-only fix and CI has been flaky anyway, I think we can go ahead with this without CI.  Thanks for the fix!

PR #845, opened 2018-10-31
Update 2.1.6 for respin

dagood:
Local build in progress to regen baselines if necessary.

New manifest: https://github.com/dotnet/versions/tree/c9bc176e798111c2ad9815cc1ef79b02ea9c8e94/build-info/dotnet/product/cli/release/2.1.6

CLI diff: https://github.com/dotnet/cli/compare/14c7845f4b84f0dc9e0d1e265de8f7f14ba3d8fc...d5e54bf5d801422416a4df16a7dc0f26dd06dd20

### DIFF
--- a/repos/roslyn.proj
+++ b/repos/roslyn.proj
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
 
   <PropertyGroup>
-    <NuGetPackageVersion>2.3.2-beta1-61921-05</NuGetPackageVersion>
+    <NuGetPackageVersion>2.10.0-beta2-63410-10</NuGetPackageVersion>
     <BuildNumber>20180426.3</BuildNumber>
 
     <BuildCommandArgs>build</BuildCommandArgs>
--- a/dependencies.props
+++ b/dependencies.props
@@ -5,7 +5,7 @@
     for which dotnet/versions commit was last used to update the dependency.
   -->
   <PropertyGroup>
-    <ProdConCurrentRef>db136a92478f1489903ef97beb905e235f5cd1da</ProdConCurrentRef>
+    <ProdConCurrentRef>c9bc176e798111c2ad9815cc1ef79b02ea9c8e94</ProdConCurrentRef>
     <CoreClrCurrentRef>1dbfe089b5c3537799927f6541185a8c1fd0adef</CoreClrCurrentRef>
     <BuildToolsCurrentRef>1dbfe089b5c3537799927f6541185a8c1fd0adef</BuildToolsCurrentRef>
   </PropertyGroup>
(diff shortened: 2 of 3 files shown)


repos/roslyn.proj
```diff
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
 
   <PropertyGroup>
-    <NuGetPackageVersion>2.3.2-beta1-61921-05</NuGetPackageVersion>
+    <NuGetPackageVersion>2.10.0-beta2-63410-10</NuGetPackageVersion>
```
dagood: I don't think this has been affecting roslyn package version (anymore?). A build I did before this PR produced e.g. `Microsoft.CodeAnalysis.CSharp.2.8.0` even though this was `2.3.2-beta1-61921-05`. Will comment here again with the version this PR ends up producing.

Doesn't seem like a problem because of forced auto-update, but we could get rid of this to be less misleading.

dagood:
Roslyn fails:

```
    Restore completed in 3.33 sec for /work/src/roslyn/src/Compilers/CSharp/csc/csc.csproj.
  /work/Tools/dotnetcli/sdk/2.1.403/Roslyn/Microsoft.Managed.Core.targets(90,5): error : SourceRoot items must include at least one top-level (not nested) item when DeterministicSourcePaths is true [/work/src/roslyn/src/NuGet/NuGetProjectPackUtil.csproj]
  /work/Tools/dotnetcli/sdk/2.1.403/Roslyn/Microsoft.Managed.Core.targets(90,5): error : SourceRoot items must include at least one top-level (not nested) item when DeterministicSourcePaths is true [/work/src/roslyn/src/Compilers/Core/Portable/CodeAnalysis.csproj]
  /work/Tools/dotnetcli/sdk/2.1.403/Roslyn/Microsoft.Managed.Core.targets(90,5): error : SourceRoot items must include at least one top-level (not nested) item when DeterministicSourcePaths is true [/work/src/roslyn/src/Compilers/Core/MSBuildTask/MSBuildTask.csproj]

  Build FAILED.
```

This update is over 3,000 commits! We missed some updates over time, it looks like. We're doing:
https://github.com/dotnet/roslyn/compare/362ec0eec4c1d61b122e271423840950fedf13ff...45b371170fd8f9f3e9c14ea2a71d17db9469f8b7
but the CLI package version diff only spans 144 commits:
https://github.com/dotnet/roslyn/compare/4b223d581506e85d41fe6b8199965a14fcf870ff...45b371170fd8f9f3e9c14ea2a71d17db9469f8b7

Investigating.

repos/roslyn.proj
dagood: Yeah, produces `Microsoft.CodeAnalysis.CSharp.2.10.0` now.

PR #871, opened 2018-11-14
Add IsDirectDependency and IsAutoReferenced to usage report.

dseefeld:
Add new properties IsDirectDependency and IsAutoReferenced from project.assets.json. 

IsDirectDependency means that the dependency shows up in the $.project.frameworks.<version>.dependencies section of the document.  IsAutoReferenced comes directly from the dependency.

### DIFF
--- a/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/Usage.cs
+++ b/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/Usage.cs
@@ -16,6 +16,10 @@ namespace Microsoft.DotNet.SourceBuild.Tasks.UsageReport
 
         public string AssetsFile { get; set; }
 
+        public bool IsDirectDependency { get; set; }
+
+        public bool IsAutoReferenced { get; set; }
+
         /// <summary>
         /// The Runtime ID this package is for, or null. Runtime packages (are assumed to) have the
         /// id 'runtime.{rid}.{rest of id}'. We can't use a simple regex to grab this value since
@@ -28,6 +32,8 @@ namespace Microsoft.DotNet.SourceBuild.Tasks.UsageReport
             nameof(Usage),
             PackageIdentity.ToXElement().Attributes(),
             AssetsFile.ToXAttributeIfNotNull("File"),
+            IsDirectDependency.ToXAttributeIfTrue(nameof(IsDirectDependency)),
+            IsAutoReferenced.ToXAttributeIfTrue(nameof(IsAutoReferenced)),
             RuntimePackageRid.ToXAttributeIfNotNull("Rid"));
 
         public static Usage Parse(XElement xml) => new Usage
@@ -40,12 +46,16 @@ namespace Microsoft.DotNet.SourceBuild.Tasks.UsageReport
         public static Usage Create(
             string assetsFile,
             PackageIdentity identity,
+            bool isDirectDependency,
+            bool isAutoReferenced,
             IEnumerable<string> possibleRuntimePackageRids)
         {
             return new Usage
             {
                 AssetsFile = assetsFile,
                 PackageIdentity = identity,
+                IsDirectDependency = isDirectDependency,
+                IsAutoReferenced = isAutoReferenced,
                 RuntimePackageRid = possibleRuntimePackageRids
                     .Where(rid => identity.Id.StartsWith($"runtime.{rid}.", StringComparison.Ordinal))
                     .OrderByDescending(rid => rid.Length)
--- a/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/Usage.cs
+++ b/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/Usage.cs
@@ -40,6 +40,8 @@ namespace Microsoft.DotNet.SourceBuild.Tasks.UsageReport
         {
             PackageIdentity = XmlParsingHelpers.ParsePackageIdentity(xml),
             AssetsFile = xml.Attribute("File")?.Value,
+            IsDirectDependency = Convert.ToBoolean(xml.Attribute(nameof(IsDirectDependency))?.Value),
+            IsAutoReferenced = Convert.ToBoolean(xml.Attribute(nameof(IsAutoReferenced))?.Value),
             RuntimePackageRid = xml.Attribute("Rid")?.Value
         };
 
--- a/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
+++ b/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
@@ -180,28 +180,37 @@ namespace Microsoft.DotNet.SourceBuild.Tasks.UsageReport
                 assetFiles,
                 assetFile =>
                 {
-                    var properties = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                    JObject jObj;
 
                     using (var file = File.OpenRead(Path.Combine(RootDir, assetFile)))
                     using (var reader = new StreamReader(file))
                     using (var jsonReader = new JsonTextReader(reader))
                     {
-                        while (jsonReader.Read())
-                        {
-                            if (jsonReader.TokenType == JsonToken.PropertyName &&
-                                jsonReader.Value is string value)
-                            {
-                                properties.Add(value);
-                            }
-                        }
+                        jObj = (JObject)JToken.ReadFrom(jsonReader);
                     }
+
+                    var properties = new HashSet<string>(
+                        jObj.SelectToken("$.targets.*").Select(t => ((JProperty)t).Name).Union(
+                            jObj.SelectToken("$.libraries").Select(t => ((JProperty)t).Name)),
+                        StringComparer.OrdinalIgnoreCase);
+
+                    var directDependencies = jObj.SelectToken("$.project.frameworks.*.dependencies").Select(dep =>
+                        new {
+                            name = ((JProperty)dep).Name,
+                            version = dep.SelectToken("$..version")?.ToString().Replace("[", "").Replace(", )", ""),
+                            autoReferenced = dep.SelectToken("$..autoReferenced")?.ToString() == "True",
+                        });
 
                     foreach (var identity in toCheck
                         .Where(id => properties.Contains(id.Id + "/" + id.Version.OriginalVersion)))
                     {
+                        var directDependency = directDependencies.Where(d => d.name == identity.Id && d.version == identity.Version.OriginalVersion).FirstOrDefault();
+
                         usages.Add(Usage.Create(
                             assetFile,
                             identity,
+                            directDependency != null ? true : false,
+                            directDependency != null ? directDependency.autoReferenced : false,
                             possibleRids));
                     }
                 });
@@ -214,6 +223,8 @@ namespace Microsoft.DotNet.SourceBuild.Tasks.UsageReport
                 usages.Add(Usage.Create(
                     null,
                     restoredWithoutUsagesFound,
+                    false,
+                    false,
                     possibleRids));
             }
 
--- a/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
+++ b/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
@@ -190,18 +190,18 @@ namespace Microsoft.DotNet.SourceBuild.Tasks.UsageReport
                     }
 
                     var properties = new HashSet<string>(
-                        jObj.SelectTokens("$.targets.*")?.Children()
+                        jObj.SelectTokens("$.targets.*").Children()
                             .Concat(jObj.SelectToken("$.libraries"))
                             .Select(t => ((JProperty)t).Name)
                             .Distinct(), 
                         StringComparer.OrdinalIgnoreCase);
 
-                    var directDependencies = jObj.SelectTokens("$.project.frameworks.*.dependencies")?.Children().Select(dep =>
+                    var directDependencies = jObj.SelectTokens("$.project.frameworks.*.dependencies").Children().Select(dep =>
                         new
                         {
-                            name = ((Newtonsoft.Json.Linq.JProperty)dep).Name,
+                            name = ((JProperty)dep).Name,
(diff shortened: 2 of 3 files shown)


dseefeld:
@dotnet-bot test this

dseefeld:
@dotnet-bot test Windows_NT Release

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,37 @@ public override bool Execute()
...
 
+                    var properties = new HashSet<string>(
+                        jObj.SelectToken("$.targets.*").Select(t => ((JProperty)t).Name).Union(
+                            jObj.SelectToken("$.libraries").Select(t => ((JProperty)t).Name)),
+                        StringComparer.OrdinalIgnoreCase);
+
+                    var directDependencies = jObj.SelectToken("$.project.frameworks.*.dependencies").Select(dep =>
+                        new {
+                            name = ((JProperty)dep).Name,
+                            version = dep.SelectToken("$..version")?.ToString().Replace("[", "").Replace(", )", ""),
```
dagood: `VersionRange.Parse` here, then use `.Satisfies` later? That would handle restores that didn't find the exact version, and doesn't require string manipulation.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,37 @@ public override bool Execute()
...
                     {
-                        while (jsonReader.Read())
-                        {
-                            if (jsonReader.TokenType == JsonToken.PropertyName &&
-                                jsonReader.Value is string value)
-                            {
-                                properties.Add(value);
-                            }
-                        }
+                        jObj = (JObject)JToken.ReadFrom(jsonReader);
```
dagood: A reason to do it the earlier way was performance. I don't really think this will be different enough to matter, but once CI is done I think before/after should be compared.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,37 @@ public override bool Execute()
...
+                        jObj.SelectToken("$.targets.*").Select(t => ((JProperty)t).Name).Union(
+                            jObj.SelectToken("$.libraries").Select(t => ((JProperty)t).Name)),
+                        StringComparer.OrdinalIgnoreCase);
+
+                    var directDependencies = jObj.SelectToken("$.project.frameworks.*.dependencies").Select(dep =>
+                        new {
+                            name = ((JProperty)dep).Name,
+                            version = dep.SelectToken("$..version")?.ToString().Replace("[", "").Replace(", )", ""),
+                            autoReferenced = dep.SelectToken("$..autoReferenced")?.ToString() == "True",
+                        });
```
dagood: Unintentional multiple enumeration when it's used later? (Add `ToArray`?)

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,37 @@ public override bool Execute()
...
+
                     foreach (var identity in toCheck
                         .Where(id => properties.Contains(id.Id + "/" + id.Version.OriginalVersion)))
                     {
+                        var directDependency = directDependencies.Where(d => d.name == identity.Id && d.version == identity.Version.OriginalVersion).FirstOrDefault();
```
dagood: Long line.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,37 @@ public override bool Execute()
...
-                            {
-                                properties.Add(value);
-                            }
-                        }
+                        jObj = (JObject)JToken.ReadFrom(jsonReader);
                     }
 
+                    var properties = new HashSet<string>(
+                        jObj.SelectToken("$.targets.*").Select(t => ((JProperty)t).Name).Union(
+                            jObj.SelectToken("$.libraries").Select(t => ((JProperty)t).Name)),
```
dagood: `Concat` earlier then take `Distinct` to factor out `Select`?

```C#
jObj.SelectToken("$.targets.*")
    .Concat(jObj.SelectToken("$.libraries"))
    .Select(t => ((JProperty)t).Name))
    .Distinct()
```

Thought about also suggesting `OfType` for safety, but actually I think it's good to get CI failures if this fails to make sure our prebuilt reports weren't invalidated by some NuGet change.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,37 @@ public override bool Execute()
...
                     foreach (var identity in toCheck
                         .Where(id => properties.Contains(id.Id + "/" + id.Version.OriginalVersion)))
                     {
+                        var directDependency = directDependencies.Where(d => d.name == identity.Id && d.version == identity.Version.OriginalVersion).FirstOrDefault();
+
                         usages.Add(Usage.Create(
                             assetFile,
                             identity,
+                            directDependency != null ? true : false,
+                            directDependency != null ? directDependency.autoReferenced : false,
```
dagood: `directDependency?.autoReferenced ?? false`? (That's easier to read to me, at least.)

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,37 @@ public override bool Execute()
...
+
                     foreach (var identity in toCheck
                         .Where(id => properties.Contains(id.Id + "/" + id.Version.OriginalVersion)))
                     {
+                        var directDependency = directDependencies.Where(d => d.name == identity.Id && d.version == identity.Version.OriginalVersion).FirstOrDefault();
+
                         usages.Add(Usage.Create(
                             assetFile,
                             identity,
+                            directDependency != null ? true : false,
```
dagood: Just `directDependency != null`?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
dagood: `FirstOrDefault` accepts a predicate func, no need for `Where`.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/Usage.cs
```diff
@@ -28,6 +32,8 @@ public class Usage : IEquatable<Usage>
             nameof(Usage),
             PackageIdentity.ToXElement().Attributes(),
             AssetsFile.ToXAttributeIfNotNull("File"),
+            IsDirectDependency.ToXAttributeIfTrue(nameof(IsDirectDependency)),
+            IsAutoReferenced.ToXAttributeIfTrue(nameof(IsAutoReferenced)),
```
dagood: Missing in `Parse`

dagood:
2.1 doesn't trigger Windows normally, it's just here because of the branch switch. 😄

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
dseefeld: Spotcheck of a number of jobs before and after:

Job | Before | After | Diff
-- | -- | -- | --
Ubuntu16.04_Release_prtest | 00:01.7 | 00:02.3 | 00:00.6
CentOS7.1_Release_prtest | 00:01.6 | 00:02.6 | 00:01.0
Debian8.4_Release_prtest | 00:01.5 | 00:02.0 | 00:00.5
RHEL7.2_Release_prtest | 00:01.8 | 00:02.3 | 00:00.5
Ubuntu16.04_Release_prtest | 00:01.7 | 00:02.3 | 00:00.6
  |   | Avg Diff: | 00:00.6

It doesn't seem like this is different enough to be an issue.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,44 @@ public override bool Execute()
...
-                                jsonReader.Value is string value)
-                            {
-                                properties.Add(value);
-                            }
-                        }
+                        jObj = (JObject)JToken.ReadFrom(jsonReader);
                     }
 
+                    var properties = new HashSet<string>(
+                        jObj.SelectTokens("$.targets.*")?.Children()
```
dagood: When does `SelectTokens` return null? The `?.` looks like it'll apply incorrectly and make the whole expression `null` when that happens. In a local repro attempt, it returns empty if there aren't any targets elements.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,44 @@ public override bool Execute()
...
+                            .Concat(jObj.SelectToken("$.libraries"))
+                            .Select(t => ((JProperty)t).Name)
+                            .Distinct(), 
+                        StringComparer.OrdinalIgnoreCase);
+
+                    var directDependencies = jObj.SelectTokens("$.project.frameworks.*.dependencies")?.Children().Select(dep =>
+                        new
+                        {
+                            name = ((Newtonsoft.Json.Linq.JProperty)dep).Name,
```
dagood: Why fully qualified?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
```diff
@@ -180,28 +180,44 @@ public override bool Execute()
...
+                            target = dep.SelectToken("$..target")?.ToString(),
+                            version = VersionRange.Parse(dep.SelectToken("$..version")?.ToString()), //dep.SelectToken("$..version")?.ToString().Replace("[", "").Replace(", )", ""),
```
dagood: More `?.`s on `SelectToken`, but `VersionRange.Parse` throws on null anyway.

dagood: Comment left in?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
dagood: Thanks for checking, yeah, seems fine.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
dseefeld: Agreed.  It doesn't look like SelectTokens() returns null.  I'll remove the ?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
dseefeld: That's a mistake.  I'll remove it.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UsageReport/WritePackageUsageData.cs
dseefeld: Removed.

dseefeld: It's required on SelectToken, because it can return null and I need the ToString() here.

dagood: Ah, didn't notice the plural difference. However, why is it better to wait to throw an exception in `VersionRange.Parse` than throw it on `ToString`?

dseefeld: I'm not sure it's better one way or the other.  Either way, it will fail if we have a dependency that does not have a version.  In the case of `VersionRange.Parse` throwing, we will know it failed trying to parse a version, which may be a quicker indication of the issue, but in the end, it is going to fail.

dagood: To make it extra diagnosable, you can try this syntax, not sure which C# version it was added in though:

```C#
dep.SelectToken("$..version")?.ToString()
?? throw new NotSupportedException($"Direct dependency has no version in '{assetFile}': '{dep}'")
```

I agree that in `?.` vs. `.`, `?.` is more diagnosable in this case, but it's less readable because it's easy to wonder why it's `?.` rather than `.`.

dseefeld: I would argue that it is more readable because it matches all of the other attribute accessors, which 1) select a token, 2) check for null and 3) call `ToString`.

dagood: What makes it less readable to me is that when I see it, I stop and wonder "is this a bug?" because it appears that `?.` is *intended* to make the code robust to `null`s even though it doesn't.

In any case... adding the exception throw expression would make it clear on first reading, clarify the runtime error way more (probably even diagnosable from CI artifacts), and keep `?.` self-similarity.

This is really minor and I don't mind merging anyway.

Issue #851, opened 2018-11-05
Publish AzDO prebuilt report data to public location


dagood:
We have a PowerBI dashboard that uses the xml file of prebuilt usage data.

* Jenkins provides an endpoint that retrieves a file artifact of a given name from the latest build on a given branch. PowerBI checks this to consume the data.
* AzDO doesn't: https://dev.azure.com/dnceng/internal/_workitems/edit/92.

The workaround we have is to use an internal build (to have access to secrets) and use a publish key to upload the file to some public location per branch.

Issue #867, opened 2018-11-13
release/2.2 source-built SDK contains Microsoft.Web.XmlTransform.dll from nuget package


dseefeld:
Microsoft.Web.XmlTransform.dll showed up in release/2.2 as an additional poisoned package. This is not there in release/2.1.  This should be addressed with the issue to Investigate and remove poisoned file leaks.  See #812

crummel:
@crummel to check if this is still relevant

MichaelSimons:
release/2.2 is EOL.  Closing

PR #873, opened 2018-11-14
Add PublishPrebuiltReportData target to make rolling build prebuilt data available

dagood:
Publishes the prebuilt report data for tarball builds to a blob storage location. Just needs the `publish.blobStorage.*` variables specified in the internal build def.

https://github.com/dotnet/source-build/issues/851

### DIFF
--- a/.vsts.pipelines/jobs/ci-linux.yml
+++ b/.vsts.pipelines/jobs/ci-linux.yml
@@ -202,7 +202,7 @@ jobs:
   # Also publish per-build data to a unique location for future use.
   - template: ../steps/publish-prebuilt-data-sh.yml
     parameters:
-      relativeBlobPath: prebuilt-data/$(Build.SourceBranchName)/rolling/$(Build.SourceVersion)/$(Build.DefinitionName)/$(Build.BuildNumber)/$(artifactName)
+      relativeBlobPath: prebuilt-data/$(Build.SourceBranchName)/rolling/$(Build.SourceVersion)/$(Build.BuildNumber)/$(Build.DefinitionName)/$(artifactName)
 
   # Clean up (very large) working directory. root owner makes it difficult for others to remove.
   - script: $(docker.run) $(docker.root.map) $(imageName) bash -c 'rm -rf /root/sb'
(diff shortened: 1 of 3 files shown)


crummel:
Think there's value in keeping historical versions around as well as latest?  @dseefeld and I were looking back at previous reports when we were looking at #867.

dagood:
Hmm. They are kept around in the AzDO build artifacts, and they're reproducible for all practical purposes. However, the builds aren't retained for very long, and in theory, feeds can be mutated, so it wouldn't hurt to save each one to blob storage with a build-versioned path as well. I'll add that.

.vsts.pipelines/jobs/ci-linux.yml
```diff
@@ -195,6 +195,15 @@ jobs:
...
       ArtifactType: Container
 
+  # Publish prebuilt report data to blob storage.
+  - template: ../steps/publish-prebuilt-data-sh.yml
+    parameters:
+      relativeBlobPath: prebuilt-data/$(Build.SourceBranchName)/latest/$(artifactName)
+  # Also publish per-build data to a unique location for future use.
+  - template: ../steps/publish-prebuilt-data-sh.yml
+    parameters:
+      relativeBlobPath: prebuilt-data/$(Build.SourceBranchName)/rolling/$(Build.SourceVersion)/$(Build.DefinitionName)/$(Build.BuildNumber)/$(artifactName)
```
dagood: I figure I have to include at least `Build.SourceVersion`, `Build.DefinitionName`, and `Build.BuildNumber` in here to make sure each build has a unique path, and make it reasonable to know which commit the data is for, I'm not sure about the order though. @dseefeld @crummel Does what I have make sense to you?

crummel: I think that makes sense.  I could see swapping `DefinitionName` and `BuildNumber` too but I'm fine with this, I can't imagine we'll have too many builds for the same commit.

dagood: Hmm, yeah, agreed the swap makes more sense. Thanks.

PR #901, opened 2018-11-16
Change where application-insights writes its output

dseefeld:
App-insights places its output under a path one directory up from its source based on finding an EnlistmentRoot.marker file.  This puts this output right under the source-build/src directory.  Override the EnlistmentRoot property to place it under the source-build/src/application-insights directory.  

Also need to override the RelativeOutputPathBase property because it is calculated from the EnlistmentRoot property.  This is used to differentiate between outputs in app-insight's mapping scheme, but is not needed for source-build.

### DIFF
--- a/repos/application-insights.proj
+++ b/repos/application-insights.proj
@@ -2,7 +2,7 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
   <PropertyGroup>
-    <PackagesOutput>$(SubmoduleDirectory)/bin/$(Configuration)</PackagesOutput>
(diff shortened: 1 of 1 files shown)


repos/application-insights.proj
```diff
@@ -13,6 +13,9 @@
     <PropertyGroup>
       <BuildCommandArgs>$(ProjectDirectory)/Microsoft.ApplicationInsights.csproj</BuildCommandArgs>
       <BuildCommandArgs>$(BuildCommandArgs) /p:Configuration=$(Configuration)</BuildCommandArgs>
+      <!-- Re-assign EnlistmentRoot property so output directories end up under src/application-insights -->
+      <BuildCommandArgs>$(BuildCommandArgs) /p:EnlistmentRoot=$(ProjectDirectory)/src</BuildCommandArgs>
+      <BuildCommandArgs>$(BuildCommandArgs) /p:RelativeOutputPathBase=</BuildCommandArgs>
```
crummel: I think Windows might need an explicit empty string here but I'm not sure.  It's an easy fix when we turn Windows on again if you don't care to experiment.  I'm fine merging as-is.

Issue #898, opened 2018-11-16
Support Windows builds in release/2.1 up to Core-Setup


dagood:
This issue is to support Windows builds up to Core-Setup (Microsoft.NETCore.App), add it to AzDO CI, and flow the work through to `release/2.2` and `master`. We have non-Windows-only change debt, and want to get ahead of it before the more dramatic build flow changes coming soon.

`master` currently has a Windows build, but it is a special case causing difficulties when auto-merging and upgrading.

A few things that can be built on top of this, but aren't included:
* Build the SDK. We know we have some issues to work through in the submodules before we can do this.
* https://github.com/dotnet/source-build/issues/904 Smoke-test support. (Requires a full SDK/CLI build, right now. Maybe we could work around it to use the tooling SDK with source-built Runtime.)
* Building a tarball that can build (offline) on Windows. This doesn't address any known real-world scenario. The only thing I can think of is it would potentially be a convenience for Microsoft devs to work with the offline build flow while on Windows, if they aren't used to Linux.

dagood:
Closing, merged into 2.1. New issues: #904 (smoke-test) #907 (prebuilt baseline)

PR #903, opened 2018-11-19
Fix build on Windows up to Core-Setup, add CI leg

dagood:
https://github.com/dotnet/source-build/issues/898

First commit fixes the build, second commit renames some yaml for better diffing in the third commit, and third adds windows support to YAML. (I suggest looking at them separately.)

Windows CI is fairly hacky. To avoid reimplementing stuff, I used Git bash, which works natively in the `Hosted` pool. However, `Hosted` only has ~10 GB free, and we need ~15GB for the Core-Setup subtree, so I had to switch to `dotnet-internal-temp` which doesn't quite support Git bash. I added a step to rig it up. We should have more space on `Hosted` in the future and we can switch back to it and remove this workaround.

The `dotnet-*-temp` Windows pools are also differently named between internal and public projects. I changed the old `ci.yml` into a `matrix.yml` template, which is now used by `ci.yml` and `ci-internal.yml`. (When I tried using a variable rather than template parameter to determine the pool name, it didn't work. That would have been better to keep the yaml flatter.)

---

This appears to produce a usable runtime. Taking the `dotnet-runtime-2.1.6-win-x64.zip` output, putting the 2.1.403 `sdk` dir in there, disabling multi-level lookup, then `.../dotnet new console` `.../dotnet run` seems to work. Process Explorer says it loaded everything from the source-built shared runtime. Will need to do more to make sure it flows changes, though, and add a smoke test to CI.


.vsts.pipelines/steps/setup-windows-bash.yml
```diff
@@ -0,0 +1,29 @@
...
+    echo "Creating bash.exe from sh.exe in $bashDir..."
+    cp -Recurse "$binDir\*" $bashDir
+    cp "$bashDir\sh.exe" "$bashDir\bash.exe"
+
+    echo "Files in bash dir:"
+    ls $bashDir
+
+    echo "Adding $bashDir to PATH as AzDO variable for bash.exe..."
+    $env:PATH = "$bashDir;" + $env:PATH
+    echo "##vso[task.setvariable variable=PATH;]$($env:PATH)"
```
crummel: Will this eventually run into the `PATH` too long issue that some builds see or does the path change go away after our build?

dagood: Good question, this is only within the context of the running AzDO build. Basically this injects the value of PATH I created in this process as a starting parameter of each next step--it's all process-local. (Not actually sure how to persist it, only ever used the UI for that.)

Issue #909, opened 2018-11-20
Add sanity check: ensure each repo created some artifacts


dagood:
We should make sure each submodule builds roughly what we expect. This would let us detect and diagnose silent failures and block PRs when they're introduced.

For example, CoreCLR fails silently in some cases. When this happens, source-build continues, but downstream repos use Microsoft official build packages when they depend on CoreCLR bits. This manifests as custom changes in the CoreCLR submodule not making it into Microsoft.NETCore.App or the SDK.

dagood:
Merged with https://github.com/dotnet/source-build/pull/910.

PR #910, opened 2018-11-21
Assert each build produces packages

dagood:
> Detects new packages after each repo and prints a hopefully useful error if none were produced.
>
> As a side benefit, stop printing the entire package version props file before every single repo: instead, print the list of package identities each build produced.

The big benefit is catching silent failures in CI and local builds. There might still be a way for a silent failure to happen that leaves partial packages available--detailed baselining would fix this, but we would have to carefully maintain it (investigate whether a change is "real"). I don't see enough value to do that now.

Users who try to `git clone ; build ; [make changes]; build` (fine for normal repos, but we don't support incremental builds) will hit this error, so I tried to make the message actionable.

Example:

```
C:\git\source-build\repos\dir.targets(342,5): error : linker produced no new source-built package identities. Known possible causes: [C:\git\source-build\repos\linker.proj]
C:\git\source-build\repos\dir.targets(342,5): error : ^ There may have been a silent failure in the submodule build. To confirm, check the build log file for undetected errors that may have prevented package creation: C:\git\source-build\bin/logs/linker.log [C:\git\source-build\repos\linker.proj]
C:\git\source-build\repos\dir.targets(342,5): error : ^ This error might be a false positive if linker intentionally builds no nuget packages. If so, set the SkipEnsurePackagesCreated property to true in C:\git\source-build\repos\linker.proj [C:\git\source-build\repos\linker.proj]
C:\git\source-build\repos\dir.targets(342,5): error : ^ The 'bin' directory might be dirty from a previous build and the package files already existed. If so, perform a clean build, or check which packages were already in 'bin' by opening C:\git\source-build\bin/obj/x64/Release/PackageVersions.props.pre.linker.xml [C:\git\source-build\repos\linker.proj]
```

https://github.com/dotnet/source-build/issues/909

### DIFF
--- a/repos/dir.targets
+++ b/repos/dir.targets
@@ -334,9 +334,9 @@
     </ItemGroup>
 
     <ItemGroup>
-      <_PackagesNotCreatedReason Include="^ Silent submodule build failure: check $(RepoConsoleLogFile)" />
-      <_PackagesNotCreatedReason Include="^ $(RepositoryName) doesn't build nuget packages: add SkipEnsurePackagesCreated=true property to $(MSBuildProjectFile)" />
-      <_PackagesNotCreatedReason Include="^ Identities already present in dirty 'bin' directory: perform clean build or check $(_PackageVersionPropsBackupPath)" />
+      <_PackagesNotCreatedReason Include="^ There may have been a silent failure in the submodule build. To confirm, check the build log file for undetected errors that may have prevented package creation: $(RepoConsoleLogFile)" />
+      <_PackagesNotCreatedReason Include="^ This error might be a false positive if $(RepositoryName) intentionally builds no nuget packages. If so, set the SkipEnsurePackagesCreated property to true in $(MSBuildProjectFullPath)" />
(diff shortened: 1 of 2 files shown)


repos/dir.targets
```diff
@@ -322,6 +322,34 @@
...
+    <ItemGroup>
+      <JustSourceBuiltPackages
+        Include="$(SourceBuiltPackagesPath)*.nupkg"
+        Exclude="
+          $(SourceBuiltPackagesPath)*.symbols.nupkg;
+          @(PreviouslySourceBuiltPackages)" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <_PackagesNotCreatedReason Include="^ Silent submodule build failure: check $(RepoConsoleLogFile)" />
```
crummel: I think this could be a bit more clear that these are just possible causes, maybe something like
- "This could be a silent submodule build failure..."
- "If $(RepositoryName) doesn't build nuget packages..."

dagood: True, I was originally thinking they'd show up more clearly as one error message, but since they're split up it's a bit confusing. Thanks.

PR #911, opened 2018-11-21
Fix Windows build: clean submodules

dagood:
AzDO doesn't seem to clean the submodules, so we have to until we can switch off the pool with persistent agents.

Saw this happen in https://github.com/dotnet/source-build/pull/910 as failure to apply patches: https://dnceng.visualstudio.com/public/_build/results?buildId=47321&view=logs. The failure log might go away when I retry it.


dagood:
I'm going ahead with the merge because this only affects OSX and Windows AzDO CI legs, those specific legs have completed already, and the Windows leg would be flaky until this is merged.

PR #916, opened 2018-11-27
Update source-build to use core repos with Arcade

crummel:
This still has some cleanup work to do but should be good enough for a preliminary review.  I also need to open PRs for the patches in submodules.


init-tools.sh
```diff
@@ -193,5 +193,15 @@ ls "$__scriptpath/Tools/scripts/docker/"*.sh | xargs chmod +x
 
 mkdir -p "$(dirname "$__BUILD_TOOLS_SEMAPHORE")" && touch "$__BUILD_TOOLS_SEMAPHORE"
 
+echo "Done initializing BuildTools."
+
+echo "Initializing Arcade..."
+scriptroot=$__scriptpath/eng/common/ \
+DOTNET_INSTALL_DIR=$__DOTNET_PATH \
+DotNetBuildFromSource=true \
+$__scriptpath/eng/common/init-tools.sh
```
dagood: Quotes around this for spaces?

patches/coreclr/0001-wip.patch
```diff
@@ -0,0 +1,28 @@
...
+--- a/global.json
++++ b/global.json
+@@ -7,7 +7,7 @@
+     "python": "2.7.15"
+   },
+   "msbuild-sdks": {
+-    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.18552.21",
+-    "Microsoft.DotNet.Helix.Sdk": "1.0.0-beta.18552.21"
++    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.18556.6",
++    "Microsoft.DotNet.Helix.Sdk": "1.0.0-beta.18556.6"
```
dagood: Would it be reasonable to use global.json updates for this rather than patches?

patches/corefx/0004-wip.patch
```diff
@@ -0,0 +1,31 @@
...
+Subject: [PATCH] wip
+
+---
+ eng/common/tools.sh | 8 --------
+ 1 file changed, 8 deletions(-)
+
+diff --git a/eng/common/tools.sh b/eng/common/tools.sh
+index c7d03ee028..9aaae295bc 100644
+--- a/eng/common/tools.sh
++++ b/eng/common/tools.sh
```
dagood: Just a note: `eng/common/**` is auto-updated from Arcade, so PR will have to go there instead of CoreFX.

repos/coreclr.proj
```diff
@@ -19,6 +19,7 @@
 
     <CleanCommand>$(ProjectDirectory)/clean$(ShellExtension)</CleanCommand>
     <OfficialBuildId>20181014-03</OfficialBuildId>
+    <PackagesOutput>$(ProjectDirectory)/bin/Product</PackagesOutput>
```
dagood: Do you know if this output dir is controlled by Arcade? Will need to make Arcade respect repo API or something like it....

repos/standard.proj
```diff
@@ -3,13 +3,22 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
 
   <PropertyGroup>
-    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) -$(Configuration) --</BuildCommand>
+    <BuildArguments>-$(Configuration)</BuildArguments>
+    <BuildCommand>$(ProjectDirectory)/build$(ShellExtension) $(BuildArguments)</BuildCommand>
```
dagood: Interesting, looks like the `--` isn't required anymore with Arcade to add on `/p:foo` args?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/UpdateJson.cs
```diff
@@ -39,7 +41,11 @@ public override bool Execute()
         private void UpdateAttribute(JToken jsonObj, string[] path, string newValue)
         {
             string pathItem = path[0];
-            if (jsonObj[pathItem] == null) 
+            if (jsonObj[pathItem] == null && SkipUpdateIfMissingKey)
+            {
+                return;
+            }
+            else if (jsonObj[pathItem] == null)
```
dagood: Maybe:

```C#
if (jsonObj[pathItem] == null)
{
    string message = $"Path item [{nameof(PathToAttribute)}] not found in json file.";
    if (SkipUpdateIfMissingKey)
    {
        Log.LogMessage(MessageImportance.Low, $"Skipping update: {message} {pathItem}");
        return;
    }
    throw new ArgumentException(message, pathItem);
}
```

Removes duplicate check, logs message when it happens for diag.

patches/corefx/0001-wip.patch
```diff
@@ -0,0 +1,24 @@
...
+
+diff --git a/global.json b/global.json
+index 808d3a98ec..2c55eb7cde 100644
+--- a/global.json
++++ b/global.json
+@@ -1,6 +1,6 @@
+ {
+   "tools": {
+-    "dotnet": "2.1.401"
++    "dotnet": "2.1.403"
```
dseefeld: Should UpdateCLIVersionInGlobalJson in repos/dir.targets be updated to do this rather than using a patch?

dseefeld: ...and now I see it's marked WIP!

repos/coreclr.proj
dagood: Generally speaking, Arcade controls these, but as far as I can tell, CoreCLR doesn't use Arcade and still respects the Repo API. I'm going to remove this change.

patches/coreclr/0001-wip.patch
dagood: Changing these doesn't seem necessary. Arcade SDK overrides no matter what version, and the Helix package doesn't even seem to be used as it's not in the prebuilt report.

patches/corefx/0004-wip.patch
dagood: I removed this patch because we set `PACKAGES_DIR`: the removed block shouldn't have any effect anyway.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -113,7 +113,7 @@ public override bool Execute()
                         if (destination.EndsWith(".nuspec"))
                         {
                             var fileText = File.ReadAllText(destination);
-                            File.WriteAllText(destination, fileText.Replace("</package>", "<files><file src=\".\\**\\*.*\"/></files>\n</package>"));
+                            File.WriteAllText(destination, fileText.Replace("</package>", "<files><file src=\".\\**\\*\"/></files>\n</package>"));
```
dagood: FYI @dseefeld, this commit explains: https://github.com/dotnet/source-build/pull/916/commits/45efd65430147c26e1c45da6c3bc7a18944f89f3. This only addresses the repack problem, but I think native binaries with no extension are in general being overlooked by this task. I think we should fix in all branches.

init-tools.cmd
```diff
@@ -12,6 +12,7 @@ set BUILD_TOOLS_PATH=%PACKAGES_DIR%Microsoft.DotNet.BuildTools\%BUILDTOOLS_VERSI
 set INIT_TOOLS_RESTORE_PROJECT=%~dp0init-tools.msbuild
 set BUILD_TOOLS_SEMAPHORE_DIR=%TOOLRUNTIME_DIR%\%BUILDTOOLS_VERSION%
 set BUILD_TOOLS_SEMAPHORE=%BUILD_TOOLS_SEMAPHORE_DIR%\init-tools.completed
+set BUILD_TOOLS_ARCADE_SEMAPHORE=%TOOLRUNTIME_DIR%/%TOOLRUNTIME_DIR%\%BUILDTOOLS_VERSION%.init-tools.completed"
```
dseefeld: Is this new Arcade semaphore ever used?

dagood: Not by us, but it prevents CoreFX from trying to initialize BuildTools again.

This will get less confusing when we switch to really onboard to Arcade and start initializing BuildTools the same way CoreFX does.

init-tools.sh
```diff
@@ -145,7 +146,7 @@ fi
 
 if [ ! -e "$__BUILD_TOOLS_PATH" ]; then
     echo "Restoring BuildTools version $__BUILD_TOOLS_PACKAGE_VERSION..."
-    echo "Running: $__DOTNET_CMD restore \"$__INIT_TOOLS_RESTORE_PROJECT\" --no-cache --packages $__PACKAGES_DIR --source $__BUILDTOOLS_SOURCE /p:BuildToolsPackageVersion=$__BUILD_TOOLS_PACKAGE_VERSION /p:ToolsDir=$__TOOLRUNTIME_DIR" >> "$__init_tools_log"
+    echo "Running: $__DOTNET_CMD restore \"$__INIT_TOOLS_RESTORE_PROJECT\" --no-cache --packages $__PACKAGES_DIR --source $__BUILDTOOLS_SOURCE --source https://api.nuget.org/v3/index.json /p:BuildToolsPackageVersion=$__BUILD_TOOLS_PACKAGE_VERSION /p:ToolsDir=$__TOOLRUNTIME_DIR" >> "$__init_tools_log"
```
dseefeld: A new source was added just in the log?

patches/corefx/0001-Skip-CoreFx.Private.TestUtilities-in-source-build.patch
```diff
@@ -14,11 +14,11 @@ Patch PR: https://github.com/dotnet/corefx/pull/33217
...
 
 diff --git a/Directory.Build.props b/Directory.Build.props
-index eb59fd5c25..3fdd594ffa 100644
+index 9d96e84da0..9c300dabaa 100644
 --- a/Directory.Build.props
 +++ b/Directory.Build.props
-@@ -320,6 +320,14 @@
-     <IncludeDllSafeSearchPathAttribute Condition="'$(TargetGroup)' == 'netstandard1.0'">false</IncludeDllSafeSearchPathAttribute>
+@@ -401,6 +401,14 @@
+     <VersionFileForPackages Condition="'$(VersionFileForPackages)' == ''">$(ArtifactsObjDir)version.txt</VersionFileForPackages>
```
dseefeld: I don't understand this patch and why it changed from one thing to something completely different?

dagood: This is just the patch's context changing when I updated it from one commit to another--the patch is the same, but diffs of diffs are really confusing. Looking at the patch itself is much clearer: https://github.com/dotnet/source-build/blob/6190f47699b9b0dfef549de3e72e5766ce4a3612/patches/corefx/0001-Skip-CoreFx.Private.TestUtilities-in-source-build.patch

```patch
@@ -401,6 +401,14 @@
     <VersionFileForPackages Condition="'$(VersionFileForPackages)' == ''">$(ArtifactsObjDir)version.txt</VersionFileForPackages>
   </PropertyGroup>
 
+  <!--
+    Remove test-only dependencies during source-build: disable project that brings in XUnit.Runtime
+    dependencies through 'ReferenceFromRuntime' property.
+  -->
+  <ItemGroup Condition="'$(DotNetBuildFromSource)' == 'true' and '$(BuildTests)' != 'true'">
+    <ProjectExclusions Include="$(SourceDir)CoreFx.Private.TestUtilities/**/*.csproj" />
+  </ItemGroup>
+
   <!-- Properties related to multi-file mode for ILC tests -->
   <PropertyGroup Condition="'$(TargetGroup)' == 'uapaot'">
     <!-- Only enable multi-file for Release-x64 and Debug-x86 for now -->
```

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
dseefeld: Good catch.  Yes, I can fix this in the other branches.

patches/corefx/0001-Skip-CoreFx.Private.TestUtilities-in-source-build.patch
dseefeld: Oh OK, Got it.  Thanks!

init-tools.cmd
dseefeld: Cool.  Thanks.

PR #924, opened 2018-12-06
Publish burndown report when it is generated

dseefeld:


.vsts.pipelines/steps/publish-prebuilt-data-sh.yml
```diff
@@ -17,3 +17,18 @@ steps:
...
+      set -x
+      # Publish prebuilt data if burndown report was generated
+      $(docker.run) $(docker.src.map) $(docker.src.work) $(imageName) ./build.sh \
+        /t:PublishPrebuiltReportData \
+        "/p:RelativeBlobPath=$relativeBlobPath" \
+        /p:ContainerName=$(publish.blobStorage.container) \
+        /p:AzureAccountName=$(publish.blobStorage.account) \
+        /p:AzureAccessToken=$(publish.blobStorage.accessToken) \
+        /clp:v=D
+    displayName: ${{ format('Publish prebuilt report data ({0})', parameters.relativeBlobPath) }}
```
dagood:
```suggestion
    displayName: ${{ format('Publish prebuilt burndown data ({0})', parameters.relativeBlobPath) }}
```

Since these steps have variable conditions rather than a template condition, it'll always show up in the list of steps, so I think the name should be unique.

dseefeld: Yes.  Thanks for the update.

PR #930, opened 2018-12-12
Update master to build 3.0 preview submodules

dagood:
This updates the submodules to what was used for the 3.0 preview Microsoft build.

Some cleanup I have in mind for future PRs:
* Stop smoke-test replacing `netcoreapp2.2` with `netcoreapp3.0`, originally there to work around the templates being old. Running `dotnet new console` with the source-built SDK, for example, creates a csproj with `netcoreapp3.0`.
* Look through the prebuilts this PR to try to nip them in the bud.
* Merge patches into repos. Some patches might already have PRs out or be merged, but we can't use the merged commits yet because they aren't 3.0 preview.
* Look for patches that could be replaced with different build args or a pre-build modification.


dagood:
CI error seems really weird, still working on it. The first project (C# console) hits an error:

```
    running run
/work/testing-smoke/builtCli/sdk/3.0.100-preview-009812/Roslyn/Microsoft.CSharp.Core.targets(52,5):
  error : Roslyn compiler server reports different hash version than build task.
  [/work/testing-smoke/C#_console/C#_console.csproj]
```

I'm able to repro with this command:

```
docker run --rm -it \
  -v /home/dagood/sb/source-build:/work:z -w /work \
  microsoft/dotnet-buildtools-prereqs:centos-7-b46d863-20180719033416 \
  ./smoke-test.sh --projectOutput
```

but not when I add `-u=1000:1000` to become not root:

```
docker run --rm -it -u=1000:1000 \
  -v /home/dagood/sb/source-build:/work:z -w /work \
  microsoft/dotnet-buildtools-prereqs:centos-7-b46d863-20180719033416 \
  ./smoke-test.sh --projectOutput
```

dagood:
(The user thing seemed to be a red herring. My guess would be that for some reason, Roslyn decided not to use the compiler server at all in that case.)

I tracked this down to this new feature: https://github.com/dotnet/roslyn/pull/29711, specifically [here](https://github.com/dotnet/roslyn/blob/f51703e0b1faae077ebb5d6acc1ae8f060bed0d2/src/Compilers/Shared/BuildServerConnection.cs#L121-L125):

```C#
// early check for the build hash. If we can't find it something is wrong; no point even trying to go to the server
if (string.IsNullOrWhiteSpace(BuildProtocolConstants.GetCommitHash()))
{
    return new IncorrectHashBuildResponse();
}
```

That's based on the assembly's `CommitHashAttribute`, which ilspy tells me is `[assembly: CommitHash("")]` in our case.

Roslyn has a target that generates the attribute here: https://github.com/dotnet/roslyn/blob/d7e493956ffcbf0e932097f6f3e1fa3fa1699087/eng/targets/Imports.targets#L94. We set `OfficialBuild` to `true`, but we don't provide a `SourceRevisionId` so it stays empty.

It looks like `SourceRevisionId` normally comes from sourcelink, which we don't use: https://github.com/dotnet/sourcelink/blob/master/docs/README.md#sourcerevisionid

I'm going to try setting `SourceRevisionId` to the `GitCommitHash` we have saved.

repos/msbuild.proj
```diff
@@ -3,26 +3,25 @@
...
   <PropertyGroup>
-    <!-- Override value from commit to match expected build. -->
-    <OfficialBuildId>20180426.1</OfficialBuildId>
     <PackagesOutput>$(ProjectDirectory)/artifacts/$(Configuration)/packages</PackagesOutput>
 
-    <OutputVersionArgs>/p:PackageVersion=15.9.0-preview-000008</OutputVersionArgs>
-    <OutputVersionArgs>$(OutputVersionArgs) /p:VersionPrefix=15.9.8</OutputVersionArgs>
+    <OutputVersionArgs>/p:PackageVersion=16.0.0-preview.225</OutputVersionArgs>
+    <!-- Set assembly version. Otherwise, defaults to 1.0.0 because we don't have nerdbank versioning. -->
+    <OutputVersionArgs>$(OutputVersionArgs) /p:VersionPrefix=15.1.0</OutputVersionArgs>
```
dseefeld: Is this correct?  VersionPrefix changed from 15.9.8 to 15.1.0?

dagood:
Now we get a "this should never happen" error on two legs of CI--one online tarball build and one offline tarball build. Seems like flakiness, maybe race condition? It doesn't repro for me locally. I don't see any way to drill into what actually failed to happen (especially without a local repro). Message:

```
[SourceBuiltSdkResolver] Looking for SDK Microsoft.Build.CentralPackageVersions. Detected config(s) in env: 'REPO_TOOLSET' for 'RoslynTools.RepoToolset/1.0.0-source-built' at '/tb/tarball_57394/Tools/source-built/RoslynTools.RepoToolset/sdk/'
/tb/tarball_57394/src/msbuild/Directory.Build.targets(14,3): error : [SourceBuiltSdkResolver] 0 overrides found for 'Microsoft.Build.CentralPackageVersions'
/tb/tarball_57394/src/msbuild/Directory.Build.targets(14,3): error : Failed to resolve SDK 'Microsoft.Build.CentralPackageVersions'. Package restore was successful but a package with the ID of "Microsoft.Build.CentralPackageVersions" was not installed.
/tb/tarball_57394/src/msbuild/Directory.Build.targets(14,33): error MSB4236: The SDK 'Microsoft.Build.CentralPackageVersions' specified could not be found. [/tb/tarball_57394/src/msbuild/src/Build/Microsoft.Build.csproj]
```

(The `SourceBuiltSdkResolver` not finding any overrides is normal, it's just that MSBuild only shows the message when all resolvers fail to acquire the requested SDK.)

[Source of the NuGet error message:](https://github.com/NuGet/NuGet.Client/blob/7e85ef267561d0aee4831fc2ab1d2456c5c1985c/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs#L164-L166)

```
// This should never happen because we were told the restore succeeded.
// If we can't find the package from GetAllInstalled(), we probably did something wrong with the NuGet API
errors.Add(string.Format(CultureInfo.CurrentCulture, Strings.PackageWasNotInstalled, sdk, sdk.Name));
```

`Strings.PackageWasNotInstalled` is `Failed to resolve SDK '{0}'. Package restore was successful but a package with the ID of "{1}" was not installed.`

I'm going to work around this by using `SourceBuiltSdkResolver` to find `Microsoft.Build.CentralPackageVersions`, bypassing the NuGet resolver entirely. It's a ref-only prebuilt so we should have easy access to it.

dagood:
The override workaround includes a change to make the resolver detect the `sdk` directory case insensitively because the CentralPackageVersions package has it in uppercase.

PR #934, opened 2018-12-14
Add check for files with no extensions.  

dseefeld:
If files with no extensions are binaries, they're not eligible to be included in reference-only packages.  This adds backa handfull of prebuilts that shouldn't have been eliminated with the original referenceOnlyPackages PR.


tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CopyReferenceOnlyPackages.cs
```diff
@@ -128,5 +123,36 @@ public override bool Execute()
...
 
+        /// <summary>
+        /// Enumerate all files in a directory and its sub-directories.
+        /// </summary>
+        private static IEnumerable<string> EnumerateAllFiles(string path, string searchPattern)
+        {
+            return Directory.EnumerateFiles(path, searchPattern, SearchOption.AllDirectories);
+        }
+
+
```
dagood: \<nit> Extra newline.

Issue #933, opened 2018-12-14
Remove workaround that inserts Microsoft.Build.CentralPackageVersions as source-built SDK override


dagood:
I'm filing this issue to link from a code comment describing this workaround. https://github.com/dotnet/source-build/pull/930#issuecomment-447041144

> Now we get a "this should never happen" (per NuGet source code) error on two legs of CI--one online tarball build and one offline tarball build. Seems like flakiness, maybe race condition? It doesn't repro for me locally. I don't see any way to drill into what actually failed to happen (especially without a local repro). Message:
> 
> ```
> [SourceBuiltSdkResolver] Looking for SDK Microsoft.Build.CentralPackageVersions. Detected config(s) in env: 'REPO_TOOLSET' for 'RoslynTools.RepoToolset/1.0.0-source-built' at '/tb/tarball_57394/Tools/source-built/RoslynTools.RepoToolset/sdk/'
> /tb/tarball_57394/src/msbuild/Directory.Build.targets(14,3): error : [SourceBuiltSdkResolver] 0 overrides found for 'Microsoft.Build.CentralPackageVersions'
> /tb/tarball_57394/src/msbuild/Directory.Build.targets(14,3): error : Failed to resolve SDK 'Microsoft.Build.CentralPackageVersions'. Package restore was successful but a package with the ID of "Microsoft.Build.CentralPackageVersions" was not installed.
> /tb/tarball_57394/src/msbuild/Directory.Build.targets(14,33): error MSB4236: The SDK 'Microsoft.Build.CentralPackageVersions' specified could not be found. [/tb/tarball_57394/src/msbuild/src/Build/Microsoft.Build.csproj]
> ```
> (The `SourceBuiltSdkResolver` not finding any overrides is normal, it's just that MSBuild only shows the message when all resolvers fail to acquire the requested SDK.)
> 
> [Source of the NuGet error message:](https://github.com/NuGet/NuGet.Client/blob/7e85ef267561d0aee4831fc2ab1d2456c5c1985c/src/NuGet.Core/Microsoft.Build.NuGetSdkResolver/NuGetSdkResolver.cs#L164-L166)
> 
> ```
> // This should never happen because we were told the restore succeeded.
> // If we can't find the package from GetAllInstalled(), we probably did something wrong with the NuGet API
> errors.Add(string.Format(CultureInfo.CurrentCulture, Strings.PackageWasNotInstalled, sdk, sdk.Name));
> ```
[...]

PR #935, opened 2018-12-17
Revert "Workaround: CentralPackageVersions sdk as override"

dagood:
This reverts commit 0df70cc344d264ea792b60c1511bd7a4f700c426 per [my comment in #933](https://github.com/dotnet/source-build/issues/933#issuecomment-447919547):

> A few ways forward:
> 
> 1. Keep the workaround (using the source-built SDK resolver) that ensures this won't repro.
> 2. Remove the workaround because it doesn't seem to repro. However, if it does repro, we're in no better position to find out what happened. The problem becomes a source of flakiness.
>    * I'm relatively OK with this, we can add the workaround from 1 back in if it repros again.

### DIFF
--- a/repos/dir.targets
+++ b/repos/dir.targets
@@ -398,7 +398,7 @@
     <ItemGroup>
       <EnvironmentVariables Include="SOURCE_BUILT_SDK_ID_%(UseSourceBuiltSdkOverride.Group)=%(UseSourceBuiltSdkOverride.Identity)" />
       <EnvironmentVariables Include="SOURCE_BUILT_SDK_VERSION_%(UseSourceBuiltSdkOverride.Group)=%(UseSourceBuiltSdkOverride.Version)" />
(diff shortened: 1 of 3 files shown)


dagood:
Ha, this PR hit the problem again already (https://dnceng.visualstudio.com/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_build/results?buildId=59681):

```
2018-12-17T20:54:20.4750312Z   /tb/tarball_59681/src/msbuild/Directory.Build.targets(14,3): error : [SourceBuiltSdkResolver] 0 overrides found for 'Microsoft.Build.CentralPackageVersions'
2018-12-17T20:54:20.4750753Z   /tb/tarball_59681/src/msbuild/Directory.Build.targets(14,3): error : Failed to resolve SDK 'Microsoft.Build.CentralPackageVersions'. Package restore was successful but a package with the ID of "Microsoft.Build.CentralPackageVersions" was not installed.
```

Closing this PR: workaround stays in.

I have an idea for a different approach to inject the diagnostic nuget resolver to try to repro. Also, I'll submit them as PRs instead of isolated runs to make sure it's in the exact same environment.

PR #936, opened 2018-12-17
[WIP] Attempt to repro nuget sdk resolver failure with diagnostic info

dagood:
This PR is just to run the code through the exact flow as a normal PR.

Revert a workaround for a nuget msbuild sdk resolver failure, and attempt to add more diagnostic info to the restore attempt to figure out what's happening.


dagood:
In addition to the respin commits, I hit rebuild a number of times in AzDO. The problem has never repro'd with a diagnostic nuget resolver.

PR #941, opened 2019-01-05
DOC: Use all processors for clone

deinok:


Documentation/README.md
```diff
@@ -11,7 +11,7 @@ The scripts are supported on Windows, macOS and Linux. The scripts are based on
 This repo uses submodules so remember on fresh clones or when updating/switching branches to run
 
 ```
-git submodule update --init --recursive
+git submodule update --init --recursive --jobs $(nproc)
```
dagood: This particular line isn't specific to running in a unix shell, so I think it would be better as a note on a next line. Something like this, maybe:

---

Add a `--jobs <n>` parameter to update submodules in parallel, such as `--jobs $(nproc)` in Bash. You can also set this option permanently by configuring `submodule.fetchJobs`.

PR #951, opened 2019-01-26
Update for 3.0 preview 2.

crummel:


crummel:
Might still have a couple things falling out from the offline build but I think this is mostly to review when you guys have time.

patches/cli/0002-Exclude-test-projects-from-source-build.patch
```diff
@@ -36,5 +36,5 @@ index 406c4f8c1..62d043078 100644
 +  </PropertyGroup>
  </Project>
 -- 
-2.17.1.windows.2
+2.18.0
```
dagood: Consider doing a pass to revert patch changes that don't actually change the patch effect?

patches/corefx/0001-Don-t-overwrite-PVP-versions.patch
```diff
@@ -0,0 +1,54 @@
...
+-    <MicrosoftDotNetCodeAnalysisPackageVersion>1.0.0-beta.19072.5</MicrosoftDotNetCodeAnalysisPackageVersion>
+-    <MicrosoftDotNetGenAPIPackageVersion>1.0.0-beta.19072.5</MicrosoftDotNetGenAPIPackageVersion>
+-    <MicrosoftDotNetGenFacadesPackageVersion>1.0.0-beta.19072.5</MicrosoftDotNetGenFacadesPackageVersion>
+-    <MicrosoftDotNetXUnitExtensionsPackageVersion>2.4.0-beta.19072.5</MicrosoftDotNetXUnitExtensionsPackageVersion>
+-    <MicrosoftDotNetBuildTasksPackagingPackageVersion>1.0.0-beta.19072.5</MicrosoftDotNetBuildTasksPackagingPackageVersion>
+-    <MicrosoftDotNetCoreFxTestingPackageVersion>1.0.0-beta.19072.5</MicrosoftDotNetCoreFxTestingPackageVersion>
+-    <MicrosoftDotNetBuildTasksConfigurationPackageVersion>1.0.0-beta.19072.5</MicrosoftDotNetBuildTasksConfigurationPackageVersion>
+-    <MicrosoftDotNetBuildTasksFeedPackageVersion>2.2.0-beta.19072.5</MicrosoftDotNetBuildTasksFeedPackageVersion>
++    <MicrosoftDotNetApiCompatPackageVersion Condition="'$(MicrosoftDotNetApiCompatPackageVersion)' == ''">1.0.0-beta.19072.5</MicrosoftDotNetApiCompatPackageVersion>
```
dagood: Typically the PVP is imported after these so they override, does that not work here?

repos/dir.targets
```diff
@@ -167,6 +167,12 @@
                 PathToAttribute="%(_PossibleCliVersionJsonPath.Identity)"
                 NewAttributeValue="$(SDK_VERSION)"
                 SkipUpdateIfMissingKey="true" />
+    <!-- we always produce an arcade with version 1.0.0-dev -->
+    <UpdateJson JsonFilePath="$(GlobalJsonFile)"
```
dagood: The source built sdk resolver should tell msbuild what version to use... is this needed now for some other reason?

build.proj
```diff
@@ -78,6 +78,63 @@
     <ItemGroup>
       <ReferenceOnlyPackageDlls Remove="$(ReferencePackagesSourceDir)**/System.Runtime.dll" />
       <ReferenceOnlyPackageDlls Remove="$(ReferencePackagesSourceDir)**/netstandard.dll" />
+      <!-- Microsoft.AspNet.App DLLs that include .resource files -->
```
dseefeld: I can't see why these would be in the list of ReferenceOnlyPackageDlls.  I spot-checked a number of them and they all seem like normal packages.

crummel: Microsoft.AspNet.App is recognized as a reference package because it only has `_._` in its lib folder, but it includes a few dozen full assemblies in a `ref` folder that cause issues when we try to ildasm/ilasm them.  I'll reach out to ASP.Net and see if that should actually be happening.

dseefeld: Is it all dlls in Microsoft.AspNet.App?  If so, this could be shortened to 
`<ReferenceOnlyPackageDlls Remove="$(ReferencePackagesSourceDir)microsoft.aspnet.app/*" />`
to remove all of them.

repos/dir.targets
crummel: No, this was an early attempt.  Removed.

build.proj
crummel: Changed as suggested.

patches/corefx/0001-Don-t-overwrite-PVP-versions.patch
crummel: I went through the binlog on this one and the problem is that it's imported both before and after this file, which I think is caused by using both Arcade and BuildTools.  The second time MSBuild will skip the import.  I think this patch will go away after CoreFX gets rid of BuildTools.

patches/cli/0002-Exclude-test-projects-from-source-build.patch
crummel: Good point, fixing these.

build.proj
crummel: Looks like these also end up in other directories - I had to revert this change.

PR #984, opened 2019-03-03
Additional properties for the latest patch version of previous runtimes.

crummel:
Please do not merge yet - waiting to hear if this is necessary.  Wanted to get CI going though.


crummel:
These variables are present but unused in 2.2 - this change isn't needed.

PR #986, opened 2019-03-05
Update for 3.0 preview 3.

crummel:
New repos participating in prodcon:
- fsharp
- standard

Other changes:
- Updated patches.
- Disable new optimization in Roslyn.
- Enable binlogging in CoreCLR.
- Some F# patches that need to be propagated back (changes made to their normal build but not their build-from-source build).
- BuildTools update.
- Some new ASP.Net DLLs with resources that we have to exclude from ILAsm/ILDasm.
- Update prebuilt baselines.
- Manual patching for coherency - need to propagate DotNetPackageVersionProps back to some repos.


crummel:
I think this is close enough to start taking a look at.  There's a little bit of work left to do around smoke-test - I have a workaround locally that does not require the RID changes to smoke-test but I want to confirm with some runtime host people on how this should be handled.  Post release we may want to look at building portable by default to match the official builds.

patches/fsharp/0002-Add-type-annotation-to-F.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/src/utils/reshapedreflection.fs b/src/utils/reshapedreflection.fs
+index b1f0aeca8..33e9b0c73 100644
+--- a/src/utils/reshapedreflection.fs
++++ b/src/utils/reshapedreflection.fs
+@@ -360,7 +360,7 @@ module internal ReflectionAdapters =
+             this.GetType().GetMethod(methodName, argTypes, BindingFlags.Public)
+ 
+     type System.Char with
+-        static member GetUnicodeCategory(c) = System.Globalization.CharUnicodeInfo.GetUnicodeCategory(c)
++        static member GetUnicodeCategory(c) = System.Globalization.CharUnicodeInfo.GetUnicodeCategory(c : char)
```
dseefeld: I'm curious why this is needed?  Is it a product issue?

crummel: Yes, an unmodified F# at this commit does not build for me outside of source-build.  I think the buildfromsource projects there have suffered from some bitrot, I'm going to make a PR in F# to fix them.

PR #989, opened 2019-03-18
Add all indirect dependencies to Version.Details.xml.

crummel:


eng/Version.Details.xml
```diff
@@ -7,4 +7,122 @@
...
+      <Sha>6f29a34e3e2e429a379c9de1b0e917a524d0be16</Sha>
+    </Dependency>
+
+    <!-- external dependencies, not handled by Maestro/Arcade -->
+    <Dependency Name="Newtonsoft.Json" Version="9.0.1">
+      <Uri>https://github.com/dseefeld/Newtonsoft.Json</Uri>
+      <Sha>e43dae94c26f0c30e9095327a3a9eac87193923d</Sha>
+    </Dependency>
+
+    <!-- not yet implemented -->
```
dseefeld: Do we want these in here if they're not yet implemented?

eng/Version.Details.xml
```diff
@@ -7,4 +7,122 @@
...
+    <Dependency Name="Microsoft.NET.Sdk.WindowsDesktop" Version="3.0.0-preview4-27511-44">
```
dseefeld: What is this dependency?

eng/Version.Details.xml
crummel: I think for ASP.Net stuff it makes sense to keep it in here since that is planned to eventually be source-buildable, but I'll remove the Windows Desktop SDK since we're not planning on supporting it.  Does that sound reasonable?

eng/Version.Details.xml
crummel: Windows Desktop SDK, see other comment - I think we should remove this.

eng/Version.Details.xml
dseefeld: Yes.  That makes sense.

Issue #928, opened 2018-12-12
Add incremental builds by increasing isolation


dagood:
I think with a few infra changes (borrowing some ideas from the Microsoft build ProdCon infra), we can enable incremental builds in source-build:

### 1. Use a unique blob feed output directory for each repo build.

This allows us to understand which artifacts came from which repos and easily delete (or ignore) stale assets.

### 2. Give each repo build its required inputs only.

Do this by either merging the upstream blob feeds into an input blob feed (like ProdCon) or pointing repos to multiple blob feeds.

Providing only the required inputs improves reliability: if repos can only use the artifacts they explicitly depend on via `RepositoryReference`, we will likely detect and fix any unknown dependencies and end up with a more correct graph.

### 3. Use an isolated, fresh package cache directory for each repo build.

This lets us clear the cache to consume new packages from upstream repos. Currently, we can't simply delete it because the source-build infra also shares the single package cache.

### 4. After each build, make cached packages available to downstream builds.

Without this, each submodule would download a fresh copy of every package even if an earlier repo already downloaded it, slowing the build down considerably and increasing the impact of network flakiness.

### 5. After harvesting a package cache directory, delete it.

The duplicate packages in each isolated cache would take a lot of space, and our build's already too large for some CI infra now.

For diagnosability, when we delete a package cache, we should note which packages were in it. If any changes were made to a specific package in-place (e.g. by BuildTools init), we should leave it alone in case it's important to understand a problem. (There are likely very few instances of this.)

### 6. Add a semaphore for each repo build completion.

This way, if a build fails, fix it and rerun the default build command to pick up where it left off.

When a repo starts building, all downstream repo semaphores should be removed so the new build's dependencies flow through to the end without question. This will help when making changes to e.g. CoreFX and expecting them show up in the Runtime without fully cleaning the repo first.

### 7. We misuse `/tmp` as general staging.

`prep.sh` extracts `PackageVersions.props` to `/tmp` explictly.  This creates a problem with re-running the build or running as a different user than the first time you built it.  We probably have over such abuses of `/tmp` and should use `mktemp()` or similar instead.

---

### Non-goals

I don't think we should try to detect submodule changes and automatically rebuild. A false positive would potentially rebuild a *lot* and be very frustrating. Deleting the semaphore doesn't seem like a terrible experience. We can provide utility props to invalidate builds if that seems easier than deleting semaphores.

### Future benefits

[...]

Issue #613, opened 2018-06-14
Change check-submodules behavior: notify by default, configurable prompt or disable


dagood:
`check-submodules.sh/ps1` is helpful if you mistakenly didn't `git submodule update` or have some dev changes you haven't cleaned up. However, it gets in the way if you're actually trying to make a custom runtime. We could change the current on/off to three modes to keep some of the benefit without getting in the way:

 * default: detect submodule checkout mismatches and notify the user. Notify the user how to set up the environment for the other two modes.
 * "prompt": the current behavior, where the user is prompted for what to do for each mismatch repo.
 * "disabled": don't check for mismatches.

This way, an observant user is notified of the mismatch, and they can set the default mode they like best by setting an env variable on their machine.

dleeapho:
This is likely obviated by Arcade 'uber-clone'.

MichaelSimons:
[Triage] This issue is obsolete, source-build no longer utilizes submodules.

PR #321, opened 2018-02-08
Update PlatformApis.cs and others from core-setup

dseefeld:
Fixes #320

PR #992, opened 2019-03-21
Clone submodules from Version.Details.xml and use them to build.

crummel:
This change removes `src/` submodules from source-build and clones them as the first step in the build.
Additional changes:
- Fix the hashes in Version.Details.xml to match submodules from master.
- Add metadata to Version.Details.xml to allow skipping repos and specify directory name.
- Some of our build tasks were depending on core-setup files.  This seems janky anyway, so I copied these files over to source-build.

Remaining work:
- Do some detection to make sure `git` is callable in the production build.  We already depend on this, so the change isn't making it worse, but we could have a better error for it.
- Fix incorrect core-sdk package name so we get auto-PRs (want the build to work first so we're testing updates).
- Tarball build currently depends on these being submodules, needs a small fix.
- Use version numbers from Version.Details.xml (future PR).

Open questions:
- If the source directories already exist, what should we do?
  - Offer to delete and reclone?
  - Offer to force checkout the hash so patches will apply?
  - Try to detect whether things are in good shape for us?  (correct hash and patches either not applied or identical to `patches/` directory - could set `SkipPatches=true` if we can detect this)
    - Or reverse the patches and let the regular build apply them again?
  - Warn and do one of the above?
  - Silently do one of the above?
- Doing a `git clean -fxd` does not remove these `src/` directories because they are git repos.
  - Should a `./clean.sh` or `./build.sh /t:Clean` remove these directories?
  - Should we delete the `.git` directory so they are not recognized as git repos?
    - If so, should these go into `.gitignore`?
  - Or should we have a new directory for `cloned-src` that's treated as a build artifact?
- Preferences on clone console output?  Only in the case of errors or always?
- Darc may do some of this work for us in the future - I think it should be pretty easy to adapt to that.


crummel:
@dseefeld @dagood if either of you want to chime in on any of these questions.  I will fix the tarball build before this is ready for review though.

dagood:
Didn't look through the code, my thoughts on the questions though:

> * If the source directories already exist, what should we do?

There's a plan in https://github.com/dotnet/source-build/issues/613 that we never got around to: notify the user that it's not clean and continue without changing anything. They can pass properties to configure, turning on auto-clean, prompting, or disabling the checks altogether to get rid of the perf load. (Environment variables can be used for permanent config for free because MSBuild carries env vars into the build context.)

I'm not sure what to do to make the patches more reasonable with a dirty repo. Might not be worth too much time thinking about, considering people should be making efforts to remove them. Also, I think https://github.com/dotnet/source-build/issues/928 would help with this by making it less interesting to start from a dirty repo anyway.

> * Preferences on clone console output?  Only in the case of errors or always?

Always visible in binlog (for silent failure diag), ideally fully visible in error message if it fails in a detectable way (exit code). Writing output as it comes doesn't seem too useful because the clones (should?) happen in parallel and output'd be scrambled.

> * Doing a `git clean -fxd` does not remove these `src/` directories because they are git repos.
>   * Should a `./clean.sh` or `./build.sh /t:Clean` remove these directories?
>   * Should we delete the `.git` directory so they are not recognized as git repos?
>     * If so, should these go into `.gitignore`?
>   * Or should we have a new directory for `cloned-src` that's treated as a build artifact?

I think it makes sense to go as far as putting the repos in e.g. `bin/src/`. So it's gitignored, and (in a way that's clear to devs!) subject to deletion when a clean script/target runs.

But we do need to make sure the git data isn't deleted--it takes a painfully long time to fetch, and it will only get worse. Submodules do that by keeping the git data outside its working tree with a `.git` file pointing to the actual data in the source-build `.git` dir, and I think doing the exact same thing here would make sense. That lets ordinary git commands work as they did with submodules, too.

(This also blends into a future approach where source-build tooling is in Arcade SDK, and someone can walk up to Core-SDK and source-build without purposefully going out and setting up a specific side-by-side dir full of repos. Everything needed would be self-contained, and if enabled (like for CI builds), cached source-built outputs would be pulled down rather than just that repo's sources.)

> `git clean -fxd`

Adding another `f` removes contained git dirs, like `git clean -xdff`.

dseefeld:
> - If the source directories already exist, what should we do?

I think warn and offer to delete and reclone. 

> - Doing a `git clean -fxd` does not remove these `src/` directories because they are git repos.
>   - Should a `./clean.sh` or `./build.sh /t:Clean` remove these directories?
>   - Should we delete the `.git` directory so they are not recognized as git repos?
>     - If so, should these go into `.gitignore`?
>   - Or should we have a new directory for `cloned-src` that's treated as a build artifact?

I agree with Davis' recommendation.  If these are cloned under bin/src, it's clear that they are build artifacts and will be cleaned.

> - Some of our build tasks were depending on core-setup files.  This seems janky anyway, so I copied these files over to source-build.

This was done intentionally in https://github.com/dotnet/source-build/pull/321 because the source was getting out of date.  If these are moved back to copied source, I think we need a plan for keeping them up to date so we don't run into the same issue.

dagood:
Here's something to consider:

https://github.com/dotnet/source-build/blob/a4beac6c72e8d7609c9f22d539035d00dec2716b/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/GetHostInformation.cs#L13-L17

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CloneRepos.cs
```diff
@@ -0,0 +1,155 @@
...
+                FileName = "git",
+                Arguments = args,
+                CreateNoWindow = true,
+                WorkingDirectory = workingDir,
+                RedirectStandardError = true,
+                RedirectStandardOutput = true,
+            };
+            Log.LogCommandLine($"{psi.FileName} {psi.Arguments}");
+            var p = Process.Start(psi);
+            p.WaitForExit();
```
dagood: This looks like the classic hang trap to me: https://stackoverflow.com/q/139593. Annoying how easy it is to code it this way.

Maybe reuse MSBuild's Exec task instead of implementing it yourself? I did it that way here: https://github.com/dotnet/buildtools/blob/29c184401b9d65166475e5630ca8459246cfb4d6/src/Microsoft.DotNet.Build.Tasks/ExecWithRetriesForNuGetPush.cs#L100-L112

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CloneRepos.cs
```diff
@@ -0,0 +1,155 @@
+﻿using Microsoft.Build.Framework;
```
dagood: Missing copyright header.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CloneRepos.cs
```diff
@@ -0,0 +1,155 @@
...
+                if (other == null)
+                {
+                    return false;
+                }
+                return other.Url == this.Url && other.Name == this.Name;
+            }
+
+            public override int GetHashCode()
+            {
+                return this.Url.GetHashCode() ^ this.Name.GetHashCode();
```
dagood: Are these really necessary?

If so, maybe use tuple hashcode and equality to avoid implementing your own? `^` is also not a great hash combiner.

(Also, looks out of date.)

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CloneRepos.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            internal string Url { get; set; }
+            internal string Name { get; set; }
+            internal string Hash { get; set; }
+
+            public RepoSpec(string url, string hash) : this(url, url.Substring(url.LastIndexOf('/') + 1), hash)
+            {
+            }
+
+            public RepoSpec(string url, string name, string hash)
```
dagood: Same-positioned args should have the same meaning especially if the type is the same.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CloneRepos.cs
```diff
@@ -0,0 +1,155 @@
...
+                    yield return new RepoSpec(dep.Element("Uri").Value, dep.Element("Sha").Value);
+                }
+            }
+        }
+
+        private class RepoSpec
+        {
+            internal string Url { get; set; }
+            internal string Name { get; set; }
+            internal string Hash { get; set; }
```
dagood: Why `internal`? Everything else is `public` (which I think makes sense).

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/CloneRepos.cs
```diff
@@ -0,0 +1,155 @@
...
+                Log.LogError("git does not appear to be installed or working.  See error message for details.");
+                Log.LogErrorFromException(e, true, true, null);
+                throw;
+            }
+        }
+
+        private static IEnumerable<RepoSpec> ReadSpecFile(string filePath)
+        {
+            var doc = XDocument.Load(filePath);
+            foreach (var dep in doc.Element("Dependencies").Element("ToolsetDependencies").Elements("Dependency").Concat(doc.Element("Dependencies").Element("ProductDependencies").Elements("Dependency")))
```
dagood: Huge line!

eng/Version.Details.xml
```diff
@@ -1,84 +1,88 @@
...
-      <Sha>0951c0fd4c38a6d01affe53b042111f8f7b91e28</Sha>
+      <Sha>26042347ff79a11869af8e951dd2888051c369f0</Sha>
     </Dependency>
   </ToolsetDependencies>
   <ProductDependencies>
     <Dependency Name="ILLink.Tasks" Version="0.1.5-preview-1461378">
       <Uri>https://github.com/mono/linker.git</Uri>
-      <Sha>d72abce2133c7e37c524f7933bce09a2f44e788f</Sha>
+      <Sha>9e8bcb40755c4ec007b6ca89618968b7c5f6fe58</Sha>
+      <RepoName>linker</RepoName>
```
dagood: Does Darc leave these alone when auto-updates happen?

crummel:
Closing and reopening to try and get CI to trigger...

crummel:
Again...

eng/Version.Details.xml
crummel: Yes, but these are no longer used.

crummel:
@dseefeld @adaggarwal This should be good enough to review - the only change I'm expecting is updating arcade-services to the final version.

build-source-tarball.sh
```diff
@@ -59,6 +59,7 @@ fi
 
 export SCRIPT_ROOT="$(cd -P "$( dirname "$0" )" && pwd)"
 SDK_VERSION=$(cat $SCRIPT_ROOT/DotnetCLIVersion.txt)
+DARC_DLL="$SCRIPT_ROOT/tools-local/arcade-services/artifacts/bin/Microsoft.DotNet.Darc/Release/netcoreapp2.1/Microsoft.DotNet.Darc.dll"
```
dseefeld: Do we build darc to this location?  Do we need to?  Can we just install it?  My naive thought is that we don't need it in the tarball, so we don't need the source for it.

eng/Version.Details.xml
```diff
@@ -1,15 +1,130 @@
...
+      <Uri>https://github.com/aspnet/AspNetCore</Uri>
+      <Sha>5c6e1aa1d95bb79809c99736acb3b8abd36d08d4</Sha>
+      <SkipClone>True</SkipClone>
+    </Dependency>
+    <Dependency Name="Microsoft.NET.Sdk.Razor" Version="3.0.0-preview4.19167.2">
+      <Uri>https://github.com/aspnet/AspNetCore-Tooling</Uri>
+      <Sha>08651acc42d11510f7a9c47d47f9135efd1d1d12</Sha>
+      <SkipClone>True</SkipClone>
+    </Dependency>
+    <Dependency Name="dotnet-ef" Version="3.0.0-preview4.19167.2">
```
dseefeld: Is this a new dependency?  What brought this in?

repos/application-insights.proj
```diff
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <PreferredHash>53b80940842204f78708a538628288ff5d741a1d</PreferredHash>
+    <SourceDirectory>ApplicationInsights-dotnet</SourceDirectory>
```
dseefeld: Is this directory dictated by something else?  Is there a reason that it's changing from what we had before?

repos/application-insights.proj
```diff
@@ -1,6 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <PreferredHash>53b80940842204f78708a538628288ff5d741a1d</PreferredHash>
```
dseefeld: Just so I understand.  To move a repo forward, do each of these PreferredHashes need to be updated?  Are these auto-updated in some way?

src/reference-assemblies/build.proj
```diff
@@ -8,6 +8,6 @@
   <Target Name="Restore" />
   <Target Name="Build">
     <Copy SourceFiles="$(MSBuildProjectDirectory)/_._" DestinationFolder="$(IntermediatePath)reference-assemblies/" />
-    <MSBuild Projects="@(AllProjects)" Targets="Build" BuildInParallel="true" />
+    <MSBuild Projects="@(AllProjects)" Targets="Build" BuildInParallel="false" />
```
dseefeld: Was this causing an issue?

tools-local/prebuilt-baseline-offline.xml
```diff
@@ -70,6 +69,7 @@
     <PackageIdentity Id="Microsoft.VisualStudio.Services.Client" Version="16.138.0-preview" />
     <PackageIdentity Id="Microsoft.Win32.Registry" Version="4.4.0" />
     <PackageIdentity Id="Newtonsoft.Json" Version="11.0.2" />
+    <PackageIdentity Id="Newtonsoft.Json" Version="12.0.2" />
```
dseefeld: I wonder who's including this.  It'd be good if we could get everyone to include the version we build.

eng/Version.Details.xml
crummel: Removed extra dependencies.

repos/application-insights.proj
crummel: All of the source directories now use the same name as the repos being cloned.  This ended up being hard to work around in the uberclone and easier to work around for us, so I went in this direction.  We could also rename the `repo.proj`s if you think that would be better - I'm fine with either.

repos/application-insights.proj
crummel: Correct, the `PreferredHash` is what actually has to move forward.  Currently this will look like:
1. One of our submodules updates.
2. We get an auto-PR that changes `eng/Version.Details.xml` and nothing else.
3. CI build fails.
4. Someone updates `PreferredHash`.
5. CI build hopefully passes and is merged.

Using the version numbers and hashes out of `eng/Version.Details.xml` for our build directly is part of what I was planning for phase 2 of the dependency changes.

src/reference-assemblies/build.proj
crummel: No, reverted.

build-source-tarball.sh
crummel: I don't think we really need to build Darc, however, Darc is not published very often and the published version is currently quite far behind.  I will check on whether this is planned to change anytime soon, otherwise I think we need to stick with it for now.

tools-local/prebuilt-baseline-offline.xml
crummel: Checking on this.

crummel: This came from the Darc update - logged https://github.com/dotnet/source-build/issues/1040 to address.

build-source-tarball.sh
crummel: Logged https://github.com/dotnet/source-build/issues/1041 for this.

Issue #994, opened 2019-03-26
Fix dotnet/templating package flow


crummel:
I removed dotnet/templating from the build in https://github.com/dotnet/source-build/pull/951 because packages were not actually flowing into the final product, despite our PackageVersion props passing the correct version.  We should figure out what the issue is and re-enable the templating build.

crummel:
Fixed in https://github.com/dotnet/source-build/pull/1008.

Issue #895, opened 2018-11-16
Analyze prebuilts for templating


dseefeld:
The following prebuilt packages are included as dependencies in source-build by **templating**.  These need to be analyzed to determine how to avoid requiring packages that source-build doesn’t build.

1. Change to use the source-built version
1. Package is not critical to product and can be removed.  (i.e.test-specific, windows-specific, etc)
1. The package is used as reference only
1. The package should be built in source-build
1. Other – please elaborate on what you think should be done.

For more detailed usage information for these, please see the _Usage Exploration_ tab of the [SourceBuildPrebuiltReport](https://msit.powerbi.com/groups/dc6359c5-e96a-44ce-9d86-0af7fab1c15e/reports/64e989dd-8072-4d84-8268-140bde0cbc7d/ReportSectionc64da6d8334068f41e23).

## ~~Prebuilt Table Updated 2019-05-07~~

PackageName | PackageVersion | SourceBuiltVersion | Category
-- | -- | -- | --
Microsoft.DotNet.Cli.CommandLine | 0.1.1-alpha-167 | 1.0.0-dev.19208.1 
Microsoft.Extensions.CommandLineUtils | 1.1.0 | 1.0.1 
System.Runtime.Loader | 4.3.0 |  

</br>

## Original Prebuilt Table

PackageName | PackageVersion | SourceBuiltVersion | Category
-- | -- | -- | --
Microsoft.DiaSymReader.Pdb2Pdb | 1.1.0-beta1-62506-02 |  
Microsoft.DotNet.Cli.CommandLine | 0.1.1-alpha-167 | 1.0.0-dev.19127.1 
Microsoft.DotNet.SignTool | 1.0.0-beta.19067.6 |  
Microsoft.Extensions.CommandLineUtils | 1.1.0 | 1.0.1 
Newtonsoft.Json | 10.0.3 | 9.0.1 
System.Diagnostics.Process | 4.3.0 |  
System.Runtime.Loader | 4.3.0 |  
sn | 1.0.0 |  
vswhere | 2.5.2 |

vijayrkn:
@dseefeld - The above dependencies list was generated before the arcade. Can we re-run this to see what are the pending source build dependencies?

dseefeld:
@vijayrkn We are currently not building templating with our 3.0 build.  I will update our source and get and new build and then update this issue.  For now, this will have to be on hold.  I'll notify you when done.

dseefeld:
[...]

PR #996, opened 2019-03-26
Reset templating pre-built status

nguerrera:
The repo should not actually have been removed. See #994. I've reopened #895


nguerrera:
FYI, @mlorbetske.

PR #1080, opened 2019-06-14
Remove arcade dependency on newtonsoft so arcades gets the version it…

### LINKED ISSUE #594: Update Newtonsoft.Json and build it without patches/fork
In https://github.com/dotnet/source-build/pull/582, the newtonsoft-json submodule was pointed at @dseefeld's fork instead of using patches checked into source-build. This makes it easier to work with as a dev because we don't have to worry about `git apply` and cleaning/resetting submodules at the right times.

However, in the past, we've been bitten by pointing at "unofficial" forks of one of the submodules. The repo/branch might disappear or the branch might accidentally be force pushed, making the original commit inaccessible. This interrupts development by making it impossible to freshly clone source-build. The risk of *losing data*, however, is small: everyone who has a long-lived clone of the repo still has the old commit on their machine and just needs to push it to a new location.

Related: the source-build product isn't reliably using the source-built Newtonsoft.Json, tracked by https://github.com/dotnet/source-build/issues/593.

---

Short term, I think we're ok as-is. We could look into making a newtsonsoft-json fork in the dotnet org to store these patches, do something weird like storing it in an source-build orphan branch, or go back to patch files. I don't think that's worthwhile.

Long term, we should work with newtonsoft-json to get the project source-buildable on the latest sources.

adaggarwal:
Issue #594.

- Change newtonsoft Version Details to point to adaggarwal's newtonsoft fork
- Workaround to remove arcade's dependency on newtonsoft so it picks up the version that it wants
- Pass LibraryFrameworks via environment variable to bypass default portable frameworks that newtonsoft tries to build as they are not compatible with sb

### DIFF
--- a/repos/arcade.proj
+++ b/repos/arcade.proj
@@ -52,10 +52,6 @@
     <EnvironmentVariables Include="DotNetUseShippingVersions=false" />
   </ItemGroup>
 
-  <ItemGroup>
-    <RepositoryReference Include="newtonsoft-json" />
-  </ItemGroup>
-
   <PropertyGroup>
     <ArcadeBootstrapVersion Condition="'$(OfflineBuild)' == 'true'">1.0.0-dev</ArcadeBootstrapVersion>
     <ArcadeBootstrapVersion Condition="'$(OfflineBuild)' != 'true'">1.0.0-beta.19306.12</ArcadeBootstrapVersion>
--- a/repos/newtonsoft-json.proj
+++ b/repos/newtonsoft-json.proj
@@ -6,19 +6,22 @@
   </PropertyGroup>
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
-
+  
+  <!--The Library Frameworks was added to environment variables so as to override the frameworks in newtonsoft csproj-->
+  <!--The semicolons were added with encoding so as to avoid conflicts in MSBuild shell-->
   <ItemGroup>
-    <EnvironmentVariables Include="LibraryFrameworks=netstandard1.0%3bnetstandard1.3%3bnetstandard2.0" />
+    <EnvironmentVariables Include="LibraryFrameworks=netstandard1.0%3Bnetstandard1.3%3Bnetstandard2.0" />
+    <RepositoryReference Include="arcade" />
   </ItemGroup>   
- 
+
   <PropertyGroup>
     <NuGetConfigFile>$(ProjectDirectory)/Src/NuGet.Config</NuGetConfigFile>
     <NewtonsoftJsonKeyFilePath>$(KeysDir)Newtonsoft.Json.snk</NewtonsoftJsonKeyFilePath>
     <NewtonsoftJsonDirectory>$(ProjectDirectory)/Src/Newtonsoft.Json/</NewtonsoftJsonDirectory>
     <NewtonsoftJsonProjectPath>$(NewtonsoftJsonDirectory)Newtonsoft.Json.csproj</NewtonsoftJsonProjectPath>
     <DotnetToolCommandArguments>"/p:DotnetOnly=true" "/p:Configuration=$(Configuration)" "/p:AssemblyOriginatorKeyFile=$(NewtonsoftJsonKeyFilePath)" "/p:SignAssembly=true" "/p:PublicSign=true" "/p:TreatWarningsAsErrors=false" "/p:AdditionalConstants=SIGNED"</DotnetToolCommandArguments>
-    <BuildCommand>$(DotnetToolCommand) build $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments)</BuildCommand>
-    <BuildPackagesCommand>$(DotnetToolCommand) pack $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments)</BuildPackagesCommand>
+    <BuildCommand>$(DotnetToolCommand) build $(NewtonsoftJsonProjectPath) /bl:restore.binlog $(DotnetToolCommandArguments)</BuildCommand>
+    <BuildPackagesCommand>$(DotnetToolCommand) pack $(NewtonsoftJsonProjectPath) /bl:restore.binlog $(DotnetToolCommandArguments)</BuildPackagesCommand>
     <CleanCommand>$(DotnetToolCommand) clean $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments)</CleanCommand>
     <PackagesOutput>$(NewtonsoftJsonDirectory)bin/$(Configuration)/</PackagesOutput>
     <RepoApiImplemented>false</RepoApiImplemented>
--- a/repos/newtonsoft-json.proj
+++ b/repos/newtonsoft-json.proj
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <PreferredHash>252b0a02a539d144a246267010b91870c86744e8</PreferredHash>
+    <PreferredHash>cac0690ad133c5e166ce5642dc71175791404fad</PreferredHash>
     <SourceDirectory>Newtonsoft.Json</SourceDirectory>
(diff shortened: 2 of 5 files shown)


repos/newtonsoft-json.proj
```diff
@@ -1,17 +1,21 @@
...
   <PropertyGroup>
-    <PreferredHash>e43dae94c26f0c30e9095327a3a9eac87193923d</PreferredHash>
+    <PreferredHash>cac0690ad133c5e166ce5642dc71175791404fad</PreferredHash>
     <SourceDirectory>Newtonsoft.Json</SourceDirectory>
   </PropertyGroup>
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
 
+  <ItemGroup>
+    <EnvironmentVariables Include="LibraryFrameworks=netstandard1.0%3bnetstandard1.3%3bnetstandard2.0" />
```
crummel: Please add a comment here about how we're doing this to avoid the semicolons getting parsed by the shell.  Also one nitpick - please capitalize the B in `%3B` so it stands out from the TFMs more.

repos/newtonsoft-json.proj
```diff
@@ -24,7 +28,7 @@
   </PropertyGroup>
 
   <Target Name="Restore" BeforeTargets="Build" DependsOnTargets="UpdateNuGetConfig">
-    <Exec Command="$(DotnetToolCommand) restore $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments) /v:$(LogVerbosity) $(RedirectRepoOutputToLog)"
+    <Exec Command="$(DotnetToolCommand) restore $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments) /v:$(LogVerbosity) /bl $(RedirectRepoOutputToLog)"
```
crummel: Suggest adding binlogging to the `BuildCommand` and `BuildPackagesCommand` as well - you'll have to name them differently (e.g. `/bl:restore.binlog`).

tools-local/prebuilt-baseline-online.xml
```diff
@@ -260,6 +260,7 @@
     <Usage Id="NETStandard.Library" Version="2.0.1" />
     <Usage Id="NETStandard.Library" Version="2.0.3" />
     <Usage Id="NETStandard.Library.NETFramework" Version="2.0.1-servicing-26011-01" IsDirectDependency="true" />
+    <Usage Id="Newtonsoft.Json" Version="9.0.1" />
     <Usage Id="Newtonsoft.Json" Version="10.0.3" IsDirectDependency="true" />
     <Usage Id="Newtonsoft.Json" Version="11.0.2" IsDirectDependency="true" />
     <Usage Id="Newtonsoft.Json" Version="12.0.2" IsDirectDependency="true" />
```
crummel: Newtonsoft.Json 12.0.2 should no longer be prebuilt since we're building it - suggest comparing the generated prebuilt report with yours (bin/prebuilt-report/generated-new-baseline.xml).

repos/arcade.proj
```diff
@@ -52,10 +52,6 @@
     <EnvironmentVariables Include="DotNetUseShippingVersions=false" />
   </ItemGroup>
 
-  <ItemGroup>
-    <RepositoryReference Include="newtonsoft-json" />
```
crummel: We need to actually enforce Arcade building before Newtonsoft.Json, right?  We should add a `<RepositoryReference Include="arcade" />` to newtonsoft-json.proj with a comment that it's not a real dependency and for ordering purposes only.

tools-local/prebuilt-baseline-online.xml
adaggarwal: Yes. It had a suggestion to remove it. I probably had a misconception that 12.0.2 should also be a part of prebuilt. I'll rectify this.

tools-local/prebuilt-baseline-offline.xml
```diff
@@ -260,6 +260,7 @@
     <Usage Id="optimization.windows_nt-x64.PGO.CoreCLR" Version="99.99.99-master-20190505.1" />
     <Usage Id="optimization.windows_nt-x86.IBC.CoreCLR" Version="99.99.99-master-20190505.1" />
     <Usage Id="optimization.windows_nt-x86.PGO.CoreCLR" Version="99.99.99-master-20190505.1" />
+    <Usage Id="Newtonsoft.Json" Version="9.0.1" />
```
dseefeld: I want to make sure I understand why this is here.  Is it because there are repos that are hardcoded to Newtonsoft.Json version 9.0.1 and in the past, we were building that version so it wasn't a prebuilt?  What is the plan for eliminating this prebuilt in the future?

adaggarwal: Arcade utilizes Newtonsoft.json 9.0.1, since it was a part of source build built packages; it needn't be a part of prebuilts before. After sb was updated to build ns.j 12.0.2 and arcade couldn't use it due to sdk stuff it intrinsically still depends on 9.0.1, we needed this 9.0.1 as a prebuilt. As we get closer to preview 7, we hope to have arcade either use systemtext.json or use newer 12.0.2 as sdk updates. 

@crummel , would you like to add anything to the stuff that I probably am vaguely stating?

crummel: Yes, that's pretty much it.  Don't forget we need to file issues for removing Newtonsoft.Json as a prebuilt for repos that aren't using the source-built version (something like https://github.com/dotnet/cli/issues/11441).

repos/newtonsoft-json.proj
```diff
@@ -1,20 +1,27 @@
...
     <SourceDirectory>Newtonsoft.Json</SourceDirectory>
   </PropertyGroup>
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+  
+  <!--The Library Frameworks was added to environment variables so as to override the frameworks in newtonsoft csproj-->
+  <!--The semicolons were added with encoding so as to avoid conflicts in MSBuild shell-->
+  <ItemGroup>
+    <EnvironmentVariables Include="LibraryFrameworks=netstandard1.0%3Bnetstandard1.3%3Bnetstandard2.0" />
+    <RepositoryReference Include="arcade" />
```
crummel: Please comment about how this is a fake RepositoryReference only used for build ordering.

repos/newtonsoft-json.proj
```diff
@@ -1,20 +1,27 @@
...
   <PropertyGroup>
     <NuGetConfigFile>$(ProjectDirectory)/Src/NuGet.Config</NuGetConfigFile>
     <NewtonsoftJsonKeyFilePath>$(KeysDir)Newtonsoft.Json.snk</NewtonsoftJsonKeyFilePath>
     <NewtonsoftJsonDirectory>$(ProjectDirectory)/Src/Newtonsoft.Json/</NewtonsoftJsonDirectory>
-    <NewtonsoftJsonProjectPath>$(NewtonsoftJsonDirectory)Newtonsoft.Json.Dotnet.csproj</NewtonsoftJsonProjectPath>
+    <NewtonsoftJsonProjectPath>$(NewtonsoftJsonDirectory)Newtonsoft.Json.csproj</NewtonsoftJsonProjectPath>
     <DotnetToolCommandArguments>"/p:DotnetOnly=true" "/p:Configuration=$(Configuration)" "/p:AssemblyOriginatorKeyFile=$(NewtonsoftJsonKeyFilePath)" "/p:SignAssembly=true" "/p:PublicSign=true" "/p:TreatWarningsAsErrors=false" "/p:AdditionalConstants=SIGNED"</DotnetToolCommandArguments>
-    <BuildCommand>$(DotnetToolCommand) build $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments)</BuildCommand>
-    <BuildPackagesCommand>$(DotnetToolCommand) pack $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments)</BuildPackagesCommand>
+    <BuildCommand>$(DotnetToolCommand) build $(NewtonsoftJsonProjectPath) /bl:restore.binlog $(DotnetToolCommandArguments)</BuildCommand>
```
crummel: Please make the binlogs' names match what's happening - something like `restore.binlog`, `build.binlog`, `pack.binlog`.

PR #1102, opened 2019-06-24
Add coherency info to Version.Details.xml.

crummel:
When I had the option of something higher or lower in the graph, I went with higher - this keeps the dependency chains shorter since core-sdk or toolset depend on most things.  `core-sdk` is still a placeholder - I need to implement the workaround since core-sdk only publishes blobs, not packages.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -24,7 +24,7 @@
       <Uri>https://github.com/dotnet/coreclr</Uri>
       <Sha>48431cc037776ca359de36bf71bda8c154cc2aa9</Sha>
     </Dependency>
-    <Dependency Name="NETStandard.Library" Version="2.1.0-prerelease.19281.5" CoherentParentDependency="core-sdk">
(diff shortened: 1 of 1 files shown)


eng/Version.Details.xml
```diff
@@ -57,11 +57,11 @@
       <Uri>https://github.com/aspnet/Extensions</Uri>
       <Sha>54d51a340698b6883dd3e47be372c07e0acf75bc</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.NET.Sdk" Version="3.0.100-preview5.19303.1">
+    <Dependency Name="Microsoft.NET.Sdk" Version="3.0.100-preview6.19303.1" CoherentParentDependency="Microsoft.Dotnet.Toolset.Internal">
```
dseefeld: Why the manual change from preview5 to preview6 here?

crummel: `preview5` was an error from manual updates (nothing in source-build looks at these version numbers yet).  The 19303.1 build was always a `preview6` tag.

PR #1106, opened 2019-06-25
Remove corefx patches

ericstj:
Neither of these should be required.  Individual commits indicate where these were fixed in product.


crummel:
We're still on https://github.com/dotnet/corefx/commit/a28176b5ec68b6da1472934fe9493790d1665cae so we don't have the Build.Tasks.Feed fix yet and that's why this is failing.  I'll see if we can update to https://github.com/dotnet/corefx/commit/c4c1985531cc1f25c805bbc4801b85c98e0dfee0.

dseefeld:
Subsequent changes have removed these two patches.  This is complete now.

Issue #316, opened 2018-02-02
UpdateNuGetConfig should add <clear/> if not present


crummel:
https://github.com/dotnet/source-build/blob/48763cf4d08e02d727c50f249121ee4d42c9f980/repos/dir.targets#L79

Prevents us from accidentally hitting nuget.org/myget.org.

PR #1108, opened 2019-06-25
UpdateNuGetConfig adds <clear/> if not present

adaggarwal:
Fixes Issue #316


crummel:
You should be able to remove part of https://github.com/dotnet/source-build/blob/release/3.0/patches/linker/0003-Don-t-add-sources-when-building-offline.patch, that would help validate that this is working.

adaggarwal:
@crummel thanks for the suggestion. I have removed the part as you suggested. Are there any other patches that try to accomplish what this issue resolves?

adaggarwal:
Retrying the failed leg. Although that was a timeout issue and not a real one.

crummel:
That's the only patch doing that right now, I think.  Looks like the offline legs will pass without it though, so that looks good.

PR #1113, opened 2019-06-26
Enable tests for CoreFX

NikolaMilosavljevic:


NikolaMilosavljevic:
Hmm, it seems that I'm bringing bunch of changes that are not mine. I have merged main repo's master branch into my fork last week. Now, I'm merging all this into release/3.0, which brings changes from master branch. Is this OK? If not what is the best way to resolve?

crummel:
I cleaned up the merge, let's see if that fixes the TargetRid issues too.

crummel:
`OutputGitInfoPropsFile` is being incorrectly set to `/src/bin/obj/x64/Release /p:PortableBuild=false /p:ArchiveDownloadedPackages=false /p:FailOnPrebuiltBaselineError=true /p:ProdConBlobFeedUrlPrefix=/TargetInfo.props` for some reason.  My guess is the argument parsing stuff you changed.

crummel:
Yeah, `Configuration` is being set incorrectly.  I think you want this line to always be a space:
https://github.com/dotnet/source-build/blob/a7597a11dfc8778fcd2efda0e6c73b0d6e716a50/build.sh#L20

It's probably not showing up locally because you're not passing `/p:Configuration=Release` like the CI does.

NikolaMilosavljevic:
Taking a look at command line parsing - I had tested that before, when I originally implemented. Not sure why it's broken now.

repos/corefx.proj
```diff
@@ -18,16 +18,20 @@
     <OverrideTargetRid Condition="'$(TargetOS)' == 'Windows_NT'">win-x64</OverrideTargetRid>
 
     <BuildArguments>$(FlagParameterPrefix)configuration $(Configuration)</BuildArguments>
-    <BuildArguments>$(BuildArguments) $(FlagParameterPrefix)restore $(FlagParameterPrefix)build $(FlagParameterPrefix)pack</BuildArguments>
-    <BuildArguments>$(BuildArguments) /p:SkipTests=true</BuildArguments>
+    <BuildArguments Condition="'$(RunTests)' != 'true' AND '$(PrepForTests)' != 'true'">$(BuildArguments) $(FlagParameterPrefix)restore $(FlagParameterPrefix)build $(FlagParameterPrefix)pack</BuildArguments>
+    <BuildArguments Condition="'$(RunTests)' != 'true'">$(BuildArguments) /p:SkipTests=true</BuildArguments>
+    <BuildArguments Condition="'$(PrepForTests)' == 'true'">$(BuildArguments) $(FlagParameterPrefix)build /p:IncludeTestUtils=true</BuildArguments>
+    <BuildArguments Condition="'$(RunTests)' == 'true'">$(BuildArguments) -buildtests -test</BuildArguments>
+    <BuildArguments Condition="'$(RunTests)' == 'true'">$(BuildArguments) /p:TestRspFile=$(TestExclusionsDir)corefx/linux.docker.rsp</BuildArguments>
```
NikolaMilosavljevic: Not sure what happened, but I'm missing a commit here that conditioned this for Linux on Docker only.

adaggarwal:
@NikolaMilosavljevic, it seems I can't add you as reviewer as the PR belongs to you. But it would be nice to receive comments from you, if any. I think its ready for review now.

patches/corefx/0003-Enable-CoreFX-tests.patch
```diff
@@ -0,0 +1,388 @@
...
++      <_NetStandard21Files 
++        Include="$(_NETStandard21RefFolder)\*.dll"
++        Exclude="@(ExcludeNetStandardRefs -> '$(_NETStandard21RefFolder)\%(Identity).dll')"  />
+     </ItemGroup>
+ 
+     <Error Condition="'@(_NetStandard21Files)' == ''" Text="Could not find package assets for netstandard2.1" />
+diff --git a/src/System.ComponentModel.Composition.Registration/tests/System.ComponentModel.Composition.Registration.Tests.csproj b/src/System.ComponentModel.Composition.Registration/tests/System.ComponentModel.Composition.Registration.Tests.csproj
+deleted file mode 100644
+index 9abbcb3..0000000
+--- a/src/System.ComponentModel.Composition.Registration/tests/System.ComponentModel.Composition.Registration.Tests.csproj
```
crummel: Could we file an issue to fix these tests?

repos/arcade.proj
```diff
@@ -7,6 +7,7 @@
     <BuildCommandArgs />
     <BuildCommandArgs>$(BuildCommandArgs) $(FlagParameterPrefix)pack</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) $(FlagParameterPrefix)configuration $(Configuration)</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:IncludeTestUtils=true</BuildCommandArgs>
```
dseefeld: What is this flag? Should it be conditional on building tests?  Does it bring in prebuilts?

.vsts.pipelines/jobs/ci-linux.yml
```diff
@@ -93,6 +94,16 @@ jobs:
         /p:ProdConBlobFeedUrlPrefix=$(prodConBlobFeedUrlPrefix)
     displayName: Run smoke-test
 
+  # Run unit tests that we support.
```
dseefeld: Where do we see the result of these unit tests?

patches/corefx/0003-Enable-CoreFX-tests.patch
ericstj: I'd be curious to understand what's wrong with them... First I've heard of this.

patches/corefx/0003-Enable-CoreFX-tests.patch
```diff
@@ -0,0 +1,388 @@
...
+diff --git a/src/dirs.proj b/src/dirs.proj
+index af8d308..30932b8 100644
+--- a/src/dirs.proj
++++ b/src/dirs.proj
+@@ -13,7 +13,7 @@
+     <Project Include="src.builds" />
+     <Project Include="shims\manual\*.csproj" />
+     <Project Include="shims\ApiCompat.proj" Condition="'$(DotNetBuildFromSource)' != 'true'" />
+-    <Project Include="pretest.proj" Condition="'$(DotNetBuildFromSource)' != 'true' and '$(BuildAllConfigurations)' != 'true'" />
++    <Project Include="pretest.proj" Condition="('$(DotNetBuildFromSource)' != 'true' and '$(BuildAllConfigurations)' != 'true') or '$(IncludeTestUtils)' == 'true'" />
```
ericstj: Instead of doing this why not just manually build it after you build binaries?  Alternatively, consider removing the BuildFromSource condition.  /cc @ViktorHofer

test/exclusions/corefx/linux.docker.rsp
```diff
@@ -0,0 +1,61 @@
+# Ignore the tests that are to be circumvented for CI
+-notrait category=IgnoreForCI
+
+# Ignore any other failing tests for SB coreFX tests
+-notrait category=failing
+
+# System.Diagnostics.Tests.ProcessTests
```
ericstj: Why all these static lists?  /cc @ViktorHofer

patches/corefx/0003-Enable-CoreFX-tests.patch
adaggarwal: What I observed was that [#40426](https://github.com/dotnet/corefx/pull/40426) resolved [#40417](https://github.com/dotnet/corefx/issues/40417) and the built could reach the test stage. But the System.ComponentModel.Composition still ends up being empty, so the tests which depend on this dll would fail. That's the reason I had to leave this patch intact.

repos/arcade.proj
adaggarwal: That's true. And it doesn't bring in any additional prebuilts.

.vsts.pipelines/jobs/ci-linux.yml
adaggarwal: The results for each respective project ends up being in corefx artifacts/bin under source-build. They generate a testresults.xml file.

patches/corefx/0003-Enable-CoreFX-tests.patch
adaggarwal: We didn't want test binaries in non-test source-build builds. AFAIK that's the reason we wanted to have a flag just to enable test utils in separate source-build test builds. cc @crummel

test/exclusions/corefx/linux.docker.rsp
adaggarwal: Oh good catch! I can try it without mentioning the failed tests separately. I think `category=failing` should take care of it.

patches/corefx/0003-Enable-CoreFX-tests.patch
ericstj: This doesn't sound right.  After that fix the built S.CM.C should be winning.  We need to take another look if it is not.

adaggarwal: ref package of that binary is still ~5Kb. The current state of this PR can repro it. 

`[root@20b7e9a283da source-build]# ls -lrth bin/src/corefx.1bb52e6a3db7f3673a3825f3677b9f27b9af99aa/artifacts/bin/ref/System.ComponentModel.Composition/netstandard-Release/System.ComponentModel.Composition.dll
-rwxr--r--. 5 root root 4.5K Aug 22 15:23 bin/src/corefx.1bb52e6a3db7f3673a3825f3677b9f27b9af99aa/artifacts/bin/ref/System.ComponentModel.Composition/netstandard-Release/System.ComponentModel.Composition.dll
[root@20b7e9a283da source-build]#
`

repos/arcade.proj
crummel: I'm not sure Arcade sure care about this at all actually.  Why is this needed?

.vsts.pipelines/jobs/ci-linux.yml
crummel: It'd be nice if we could upload these as artifacts in the build, don't block on it though.

adaggarwal: That makes sense. Would you mean copying all the testResults.xml files from coreFX artifacts to source-build root bin directory? Perhaps create some directory in source-build/bin and copy all of them over there?

crummel: Yeah, I don't think the build generally has access to artifacts in the docker container so we might have to copy them out to bin/logs or something, but I'm fine with adding this later and just getting this in for now so we at least get an overall pass/fail.

patches/corefx/0003-Enable-CoreFX-tests.patch
crummel: Yeah, we generally see the test builds bring in more prebuilts than we would like.  Once our prebuilt process is improved that might the better way to go about this though.

.vsts.pipelines/jobs/ci-linux.yml
adaggarwal: Sure! I can start an issue to track this as this PR gets merged.

PR #1118, opened 2019-06-28
Add newtonsoft.json 9.0.1 to source-build

### LINKED ISSUE #1115: Add Newtonsoft.Json 9.0.1
We will not be able to supply Newtonsoft.Json 9.0.1 as a reference-only assembly, and taking 12.0.2 on the repo side has a long dependency path (NuGet -> SDK -> CLI -> core-sdk -> core-setup).  To ease the transition, we should build both 9.0.1 version and 12.0.2 versions.

adaggarwal:
- Adding newtonsoft.json removes the 9.0.1 prebuilt issue for the time-being; as some repos depend on it. 

- Fixes #1115


eng/Version.Details.xml
```diff
@@ -102,6 +102,12 @@
       <Sha>cac0690ad133c5e166ce5642dc71175791404fad</Sha>
       <RepoName>newtonsoft-json</RepoName>
     </Dependency>
+    <!-- additional Newtonsoft version to ease out individual repo updates to use current source-build version-->
+    <Dependency Name="Newtonsoft.Json" Version="9.0.1">
+      <Uri>https://github.com/adaggarwal/Newtonsoft.Json</Uri>
+      <Sha>e43dae94c26f0c30e9095327a3a9eac87193923d</Sha>
+      <RepoName>newtonsoft-json</RepoName>
```
crummel: This should match the `newtonsoft-json901` project name.

repos/newtonsoft-json901.proj
```diff
@@ -0,0 +1,34 @@
...
+    <NewtonsoftJsonProjectPath>$(NewtonsoftJsonDirectory)Newtonsoft.Json.Dotnet.csproj</NewtonsoftJsonProjectPath>
+    <DotnetToolCommandArguments>"/p:DotnetOnly=true" "/p:Configuration=$(Configuration)" "/p:AssemblyOriginatorKeyFile=$(NewtonsoftJsonKeyFilePath)" "/p:SignAssembly=true" "/p:PublicSign=true" "/p:TreatWarningsAsErrors=false" "/p:AdditionalConstants=SIGNED"</DotnetToolCommandArguments>
+    <BuildCommand>$(DotnetToolCommand) build $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments)</BuildCommand>
+    <BuildPackagesCommand>$(DotnetToolCommand) pack $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments)</BuildPackagesCommand>
+    <CleanCommand>$(DotnetToolCommand) clean $(NewtonsoftJsonProjectPath) $(DotnetToolCommandArguments)</CleanCommand>
+    <PackagesOutput>$(NewtonsoftJsonDirectory)bin/$(Configuration)/</PackagesOutput>
+    <RepoApiImplemented>false</RepoApiImplemented>
+    <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
+    <!-- this is built in the Arcade build and will not produce packages in the main build -->
+    <SkipEnsurePackagesCreated>true</SkipEnsurePackagesCreated>
```
crummel: Is this correct?  It seems like this build should be producing the Newtonsoft.Json.9.0.1 nupkg.

repos/known-good.proj
```diff
@@ -35,6 +35,7 @@
         <RepositoryReference Include="common" />
         <RepositoryReference Include="netcorecli-fsc" />
         <RepositoryReference Include="newtonsoft-json" />
+        <RepositoryReference Include="newtonsoft-json901" />
```
crummel: Was core-setup also one that depended on 9.0.1 or was that not a thing anymore?

adaggarwal: It did have a dependency afaik

repos/newtonsoft-json901.proj
adaggarwal: Yes in my local build, it did produce 901 nupkg

adaggarwal: This is the same project which sb was utilizing before I merged #1080

crummel: Ok, I don't think we should have `SkipEnsurePackagesCreated` true here then.

adaggarwal: What does ```SkipEnsurePackagesCreated``` does?

crummel: Normally we check to make sure that a project actually created nupkgs - this was added when we found some silent failures in CoreCLR and core-setup.  Setting this to true just means we won't check it, which is used for toolset and core-sdk because they produce .tar.gzs instead of nupkgs.  If anything shows up in the build as "New NuGet packages after building ..." then it isn't needed.

adaggarwal:
@crummel , can I request a review update?

repos/known-good.proj
crummel: There should be a `RepositoryReference` in repos/core-setup as well then.

crummel:
One more change, then I think we're good.

PR #1124, opened 2019-07-03
Add sourcelink repo

### LINKED ISSUE #1112: Add SourceLink repo

adaggarwal:
Fixes #1112


repos/dir.targets
```diff
@@ -372,6 +372,7 @@
       <_PackagesNotCreatedReason Include="^ There may have been a silent failure in the submodule build. To confirm, check the build log file for undetected errors that may have prevented package creation: $(RepoConsoleLogFile)" />
       <_PackagesNotCreatedReason Include="^ This error might be a false positive if $(RepositoryName) intentionally builds no nuget packages. If so, set the SkipEnsurePackagesCreated property to true in $(MSBuildProjectFullPath)" />
       <_PackagesNotCreatedReason Include="^ The 'bin' directory might be dirty from a previous build and the package files already existed. If so, perform a clean build, or check which packages were already in 'bin' by opening $(_PackageVersionPropsBackupPath)" />
+      <_PackagesNotCreatedReason Include="^ The packages may have been created in the artifacts. Check if that is the output packages path in $(MSBuildProjectFullPath)" />
```
dagood: I'm not sure what this message is saying... it isn't clear to me from the first sentence why it might be a bad thing that packages may have been created. Is this what it means?

```suggestion
      <_PackagesNotCreatedReason Include="^ The packages may have been written to an unexpected directory. For example, some repos used bin/ and changed to artifacts/ to match Arcade. Check PackagesOutput in $(MSBuildProjectFullPath) (currently '$(PackagesOutput)')" />
```

repos/known-good.proj
```diff
@@ -32,6 +32,7 @@
         <!-- Tier 1 -->
         <RepositoryReference Include="application-insights" />
         <RepositoryReference Include="aspnet-xdt" />
+        <RepositoryReference Include="sourcelink" />
```
dagood: Adding this to the all-up proj is fine to get it building, but to support building from other root project nodes I'd expect a reference from some other project, maybe the Arcade repo?

repos/sourcelink.proj
```diff
@@ -0,0 +1,57 @@
...
+    <OfficialBuildId>20190623.1</OfficialBuildId>
+
+    <BuildCommandArgs/>
+    <BuildCommandArgs>$(BuildCommandArgs) --pack</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) --configuration $(Configuration)</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) --binaryLog</BuildCommandArgs>
+    <BuildCommand>$(ProjectDirectory)build$(ShellExtension) $(BuildCommandArgs)</BuildCommand>
+
+    <GlobalJsonFile>$(ProjectDirectory)global.json</GlobalJsonFile>
+    <NuGetConfigFile>$(ProjectDirectory)/NuGet.config</NuGetConfigFile>
```
dagood: Redundant `/`? (By convention, MSBuild directory properties always already have a trailing directory separator.)

repos/sourcelink.proj
```diff
@@ -0,0 +1,57 @@
...
+    <!-- reset SDK location variables to use 3.0 SDK override -->
+    <EnvironmentVariables Remove="DOTNET_TOOL_DIR=$(DotNetCliToolDir)" />
+    <EnvironmentVariables Remove="DotNetCoreSdkDir=$(DotNetCliToolDir)" />
+    <EnvironmentVariables Remove="DotNetRoot=$(DotNetCliToolDir)" />
+    <EnvironmentVariables Remove="DotNetToolPath=$(DotNetCliToolDir)" />
+    <EnvironmentVariables Remove="DotNetTool=$(DotNetCliToolDir)dotnet" />
+    <EnvironmentVariables Remove="DOTNET_HOST_PATH=$(DotNetCliToolDir)dotnet" />
+    <EnvironmentVariables Remove="_InitializeDotNetCli=$(DotNetCliToolDir)" />
+
+    <EnvironmentVariables Include="DOTNET_TOOL_DIR=$(DotNet3CliToolDir)" />
```
dagood: The removes then includes look very brittle to me, I'd expect switching between CLIs to be implemented in `dir.props`/`dir.targets`. If this is copy-paste from an existing place it's implemented this way, though, maybe fixing this shouldn't block the PR.

repos/sourcelink.proj
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <PreferredHash>51310e65e75010467993f793e1739d1a1dad50c5</PreferredHash>
+    <SourceDirectory>sourcelink</SourceDirectory>
```
dagood: Doesn't `SourceDirectory` default to the proj name? I wouldn't expect to need a custom name here because there are no special characters.

repos/known-good.proj
adaggarwal: Actually sourcelink depends on Arcade and if I add it to arcade proj it might result in cyclic dependency. And that may become a problem. I think roslyn depends on it. But not sure about that.

repos/sourcelink.proj
adaggarwal: I think I got this from a merge update from the weekly sb update. I'll go ahead and fix this one.

repos/sourcelink.proj
adaggarwal: I utilized the source for sourcelink.proj from xdt.proj, if this seems redundant I can go ahead and remove this.

repos/known-good.proj
dagood: My assumption there was that the intention of adding sourcelink is to include it as a tool in the Arcade SDK, but I am lacking context. The dependency of `sourcelink.proj` on `arcade.proj` could be removed and the prebuilt/N-1 Arcade SDK used, AFAIK.

repos/sourcelink.proj
adaggarwal: I had it for consistency. I can try to remove it and see if it picks up without explicitly assigning it the value.

repos/sourcelink.proj
dagood: Ok, it doesn't appear redundant, just a fragile implementation. It would take some extra work to rework both places so I don't think it needs to be improved in this PR.

repos/known-good.proj
adaggarwal: I tried to remove it in the beginning. But then the build errors out saying that it was expecting arcade directory which it couldn't find. Later after adding arcade as a repoRef, the error vanished.

repos/sourcelink.proj
dagood: Here's some precedent for the default, FWIW:

https://github.com/dotnet/source-build/blob/3a522d7ee800df9a01bd9272445e21431c256eb8/repos/corefx.proj#L3-L7

https://github.com/dotnet/source-build/blob/3a522d7ee800df9a01bd9272445e21431c256eb8/repos/core-setup.proj#L3-L7

https://github.com/dotnet/source-build/blob/3a522d7ee800df9a01bd9272445e21431c256eb8/repos/arcade.proj#L3-L7

repos/known-good.proj
dagood: Maybe this was still in there, which tries to use a source-built Arcade SDK:

```
  <ItemGroup>
    <UseSourceBuiltSdkOverride Include="@(ArcadeSdkOverride)" />
  </ItemGroup>
```

Either way, no point in changing this without knowing the intended usage.

patches/sourcelink/0001-Exclude-test-projects.patch
```diff
@@ -0,0 +1,116 @@
+From 4a051438d29bd96ede45069031462522313ed1bc Mon Sep 17 00:00:00 2001
+From: adaggarwal <aditya.aggarwal@microsoft.com>
+Date: Tue, 2 Jul 2019 16:43:39 +0000
+Subject: [PATCH] Exclude test projects.
+Note: This patch is necessary in fixing arcade #3051 and awaiting uptake.
```
dagood: The phrasing makes me think the patch is needed to fix something in Arcade, but it's kind of the other way around. Maybe:

```suggestion
Note: This patch is necessary to work around https://github.com/dotnet/arcade/pull/3051. Removing this patch is awaiting uptake of a new Arcade version.
```

adaggarwal: Thanks for the suggestion.

repos/known-good.proj
dseefeld: If there are no repo references for this repo, I'm not sure why we're adding it.  It would be good to sync with @tmat, since he requested it, to understand what the intention is for this and get this worked out before merging this PR.

tmat: SourceLink should be referenced by all repos via Arcade SDK. Currently Arcade SDK references a [specific version](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/tools/DefaultVersions.props#L93) of the package.

The package is referenced here: https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/tools/RepositoryInfo.targets#L7-L10 (currently excluded when building from source, a condition we want to remove).

I think we should flow SourceLink to Arcade like we do Roslyn compiler package, so that all repos get the latest SourceLink with Arcade update.

patches/sourcelink/0001-Exclude-test-projects.patch
```diff
@@ -0,0 +1,115 @@
+From 4a051438d29bd96ede45069031462522313ed1bc Mon Sep 17 00:00:00 2001
+From: adaggarwal <aditya.aggarwal@microsoft.com>
+Date: Tue, 2 Jul 2019 16:43:39 +0000
+Subject: [PATCH] Exclude test projects.
+
```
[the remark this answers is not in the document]
tmat: I'm not sure why this is needed. SourceLink uses the latest Arcade.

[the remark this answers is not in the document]
dagood: Sure, when SourceLink builds standalone and pulls its own Arcade SDK off the network, but source-build doesn't use the latest. (Or at least it didn't, as of writing.) This may also not be the latest SourceLink due to coherency issues, but maybe not since it's newly introduced.

This patch would be classified as a source-build infra issue.

repos/known-good.proj
crummel: I think Arcade uses a frozen version of Roslyn, do we want to do the same for SourceLink?  I do think it makes sense to flow it using a Version.Details.xml entry and subscription and everything instead, we'll have to make some Arcade changes for that.

tmat: Not sure what do you mean by "frozen". Arcade SDK captures the version of Roslyn that flows into Arcade repo here: https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/Microsoft.DotNet.Arcade.Sdk.csproj#L71, so that everyone using that Arcade SDK build gets that version.

we should do the same for SourceLink.

tmat: That means Arcade repo would subscribe for SourecLink updates.

crummel: There's no Roslyn in [Version.Details.xml](https://github.com/dotnet/arcade/blob/0c2c6f3121c2c1fa5a4391b72a4df13ddc9e9959/eng/Version.Details.xml) though, and the last few updates at least to that variable are manual PRs (like https://github.com/dotnet/arcade/pull/3146), how is it getting updates?

repos/sourcelink.proj
crummel: Yeah, at this point we should be defaulting to the 3.0 SDK and making the exception the 2.x one if it's still needed.  This should happen in a different PR, I logged https://github.com/dotnet/source-build/issues/1129 for this.

repos/known-good.proj
tmat: > how is it getting updates?

Apparently manually at this point. Seems like we are missing subscription of Arcade repo to some Roslyn channel.

tmat: Whether the update is manual or automatically flows is secondary. The important thing is that if it is updated it will flow to all repos where Arcade SDK flows.

dseefeld: > The package is referenced here: https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/tools/RepositoryInfo.targets#L7-L10 (currently excluded when building from source, ***a condition we want to remove***).

@tmat Does SourceLink require source-control?  I'm not clear about including SourceLink with source-build.  What will happen when we build offline and there is no source-control?

tmat: It does not require the full `.git` dir, just a couple of files that can be created by a script. See
https://github.com/dotnet/sourcelink/tree/master/docs#minimal-git-repository-metadata

crummel: Cool, I will add that as part of my daily build work.

repos/sourcelink.proj
crummel: Still have one extra slash here.

repos/sourcelink.proj
```diff
@@ -0,0 +1,58 @@
...
+    <PreferredHash>51310e65e75010467993f793e1739d1a1dad50c5</PreferredHash>
+  </PropertyGroup>
+
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+
+  <PropertyGroup>
+    <OfficialBuildId>20190623.1</OfficialBuildId>
+
+    <BuildCommandArgs/>
+    <BuildCommandArgs>$(BuildCommandArgs) --pack</BuildCommandArgs>
```
crummel: CI looks like it's running into `-` vs `--` - you can use `$(FlagParameterPrefix)` instead.

repos/sourcelink.proj
```diff
@@ -0,0 +1,58 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <ArcadeBootstrapVersion Condition="'$(OfflineBuild)' == 'true'">1.0.0-dev</ArcadeBootstrapVersion>
+    <ArcadeBootstrapVersion Condition="'$(OfflineBuild)' != 'true'">1.0.0-beta.19306.12</ArcadeBootstrapVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- bootstrap our version of Arcade using the version of Arcade that source-build retrieves.
+         This prevents getting into a dependency loop of Arcade needing Arcade to build and keeps us from having multiple prebuilt versions of Arcade.  -->
```
crummel: Comment is slightly off since it isn't Arcade trying to build Arcade here.

tools-local/init-build.proj
```diff
@@ -105,12 +105,10 @@
   <Target Name="ExtractToolPackage"
           DependsOnTargets="BuildTasks"
           Condition="'$(OfflineBuild)' == 'true'">
-          <Message Importance="High" Text="Arcversion ->> $(ArcadeSdkOverride.Version)    SBPackPath ->> $(SourceBuiltPackagesPath)" />
     <ItemGroup>
       <_ToolPackage
-        Condition="'%(ArcadeSdkOverride.Version)' != ''"
-        Include="$(SourceBuiltPackagesPath)%(ArcadeSdkOverride.Identity).%(ArcadeSdkOverride.Version).nupkg"
-        Exclude="$(SourceBuiltPackagesPath)%(ArcadeSdkOverride.Identity).%(ArcadeSdkOverride.Version).symbols.nupkg"
+        Include="$(SourceBuiltPackagesPath)%.1.0.0-dev.nupkg"
```
dseefeld: A couple of comments:
1. SourceBuildPackagesPath should be PrebuiltSourceBuiltPackagesPath, since this is a prebuilt from the previous source-build build.
2. You need the name of the package, and no "%".  Microsoft.DotNet.Arcade.Sdk.1.0.0-dev.nupkg

tools-local/init-build.proj
```diff
@@ -105,12 +105,10 @@
...
+        Exclude="$(SourceBuiltPackagesPath)%.1.0.0-dev.symbols.nupkg"
         Id="%(ArcadeSdkOverride.Identity)" />
```
dseefeld: ArcadeSdkOverride is not defined here.  You really don't need the Id property here at all.

tools-local/init-build.proj
```diff
@@ -105,12 +105,10 @@
...
     </ItemGroup>
```
dseefeld: For the ZipFileExtractToDirectory task, the destination should be "$(PackagesDir)microsoft.dotnet.arcade.sdk/1.0.0-dev/".

Also, I'm not sure you need the _ToolPackage property.  You could just specify the SourceArchive here.  "$(PrebuiltSourceBuiltPackagesPath)Microsoft.DotNet.Arcade.Sdk.1.0.0-dev.nupkg"

tools-local/init-build.proj
dseefeld: Also, once you get the tool package extracted to the packages dir, you'll need to update the SOURCE_BUILT_SDK  EnvironmentVariables in arcade.proj to not be conditional and you can also remove the "alternate-workaround-for-bootstrapping" patch from patches/arcade.

tools-local/init-build.proj
```diff
@@ -107,16 +107,15 @@
...
     <ZipFileExtractToDirectory SourceArchive="%(_ToolPackage.Identity)"
-                               DestinationDirectory="$(ToolPackageExtractDir)%(_ToolPackage.Id)\"
+                               DestinationDirectory="$(PackagesDir)microsoft.dotnet.arcade.sdk/1.0.0-dev/"
                                OverwriteDestination="true" />
     <ItemGroup>
-      <ExtractedToolFiles Include="$(ToolPackageExtractDir)%(_ToolPackage.Id)/**/*netcore*/*.dll" />
+      <ExtractedToolFiles Include="$(PackagesDir)microsoft.dotnet.arcade.sdk/1.0.0-dev/**/*netcore*/*.dll" />
     </ItemGroup>
 
     <Copy SourceFiles="@(ExtractedToolFiles)" DestinationFolder="$(ToolPackageExtractDir)/" />
```
adaggarwal: Is the destination folder correct?

crummel: `$(ToolPackageExtractDir)` is `Tools/source-built`, so I think you want the package name as well: `$(ToolPackageExtractDir)Microsoft.DotNet.Arcade.Sdk/`.  There might be some [RecursiveDir](https://blogs.msdn.microsoft.com/msbuild/2005/11/07/how-to-recursively-copy-files-using-the-copy-task/) trick you need to do if that doesn't look right.

dseefeld: I don't think the folder is correct.  I don't think we want $(ToolPackageExtractDir).  I think we want $(PackagesDir).  i.e. I think the destination should be "$(PackagesDir)microsoft.dotnet.arcade.sdk/1.0.0-dev/".

crummel: The `ZipFileExtractToDirectory` call has `DestinationDirectory="$(PackagesDir)microsoft.dotnet.arcade.sdk/1.0.0-dev/"` already though right?  We want it in both the package cache and the `source-built` folder, I think.

adaggarwal: In that case, won't source and destination become the same?

https://github.com/adaggarwal/source-build/blob/7cbbc77254f6928866a8e8b253530e01e5c56da9/tools-local/init-build.proj#L121

tools-local/init-build.proj
```diff
@@ -101,6 +102,21 @@
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Done packing reference-only packages." />
   </Target>
 
+
+  <Target Name="ExtractToolPackage"
+          DependsOnTargets="BuildTasks"
+          Condition="'$(OfflineBuild)' == 'true'">
+    <ItemGroup>
+      <_ToolPackage
+        Include="$(PrebuiltSourceBuiltPackagesPath)Microsoft.DotNet.Arcade.1.0.0-dev.nupkg"
```
crummel: Missing `.Sdk` - `Microsoft.DotNet.Arcade.Sdk`.

crummel: Should be `Microsoft.DotNet.Arcade.Sdk.1.0.0-dev.nupkg`.

dir.props
```diff
@@ -21,6 +21,7 @@
     <DotNet3CliToolDir>$(ProjectDir)Tools/dotnetcli3/</DotNet3CliToolDir>
     <PatchesDir>$(ProjectDir)patches/</PatchesDir>
     <PackagesDir>$(ProjectDir)packages/</PackagesDir>
+    <ArcadeBootstrapPackageDir>$(PackagesDir)ArcadeBootStrapPackage/</ArcadeBootstrapPackageDir>
```
dagood: \<nitpick> Standardize on `ArcadeBootstrapPackage` casing?

repos/dir.props
```diff
@@ -72,7 +72,7 @@
     <EnvironmentVariables Include="DOTNET_HOST_PATH=$(DotNetCliToolDir)dotnet" />
     <!-- _InitializeDotNetCli is used by websdk and templating to decide whether to init the SDK -->
     <EnvironmentVariables Include="_InitializeDotNetCli=$(DotNetCliToolDir)" />
-    <EnvironmentVariables Include="_InitializeToolset=$(ProjectDir)Tools/source-built/Microsoft.DotNet.Arcade.Sdk/tools/Build.proj" Condition="'$(RepositoryName)' != 'arcade'" />
+    <EnvironmentVariables Include="_InitializeToolset=$(ProjectDir)Tools/source-built/Microsoft.DotNet.Arcade.Sdk/tools/Build.proj" Condition="'$(RepositoryName)' != 'arcade' AND '$(RepositoryName)' != 'sourcelink'" />
```
dagood: I don't think the `dir.props` should condition on repo names... seems like it should use a property name focused on the intent that `proj` files can set. When I see this, it seems analogous to having this in a base class:

```C#
if (!(foo is Arcade) && !(foo is SourceLink)) { addEnv(...); }
```

Looks like this is existing (so I suppose non-blocking) but something to consider.

repos/sourcelink.proj
```diff
@@ -0,0 +1,63 @@
...
+
+    <GlobalJsonFile>$(ProjectDirectory)global.json</GlobalJsonFile>
+    <NuGetConfigFile>$(ProjectDirectory)NuGet.config</NuGetConfigFile>
+    <LogVerbosityOptOut>true</LogVerbosityOptOut>
+    <OutputPlacementRepoApiImplemented>false</OutputPlacementRepoApiImplemented>
+    <PackagesOutput>$(ProjectDirectory)artifacts/packages/$(Configuration)/Shipping/</PackagesOutput>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(OfflineBuild)' != 'true'">
+    <ArcadeBootstrapVersion>1.0.0-beta.19359.6</ArcadeBootstrapVersion>
```
dagood: Having this specific version hard-coded makes me concerned for maintenance, is there any way to get it into `Versions.props` or grab it from `global.json` or something like that?

tools-local/init-build.proj
dseefeld: I think the second part of this comment still applies.  I don't think you need the _ToolPackage property.

adaggarwal:
@crummel Is the windows production not passing due to preview 7 pull? Any thoughts?

crummel:
Ah, my bad.  The Windows build.cmd needs to be updated with something similar to https://github.com/dotnet/source-build/pull/1124/files#diff-0b83f9dedf40d7356e5ca147a077acb4R8.

adaggarwal:
Will I need to change update sourcelink version as well? [preview 7 update versions](https://github.com/dotnet/source-build/pull/1159#issuecomment-516135170)
@crummel

crummel:
No, I think we can leave SourceLink at this version for now.  We still have SourceLink turned off in all the repos, that'll be what we can fix next.

PR #1148, opened 2019-07-18
Remove ApplyNgenOptimization settings

tmat:
NGEN optimization is not enabled when building from source.

See
https://github.com/dotnet/arcade/blob/a1805103791e43031355e11c0d037bca803a9593/src/Microsoft.DotNet.Arcade.Sdk/tools/DefaultVersions.props#L68

https://github.com/dotnet/arcade/blob/a1805103791e43031355e11c0d037bca803a9593/src/Microsoft.DotNet.Arcade.Sdk/tools/Imports.targets#L27

Roslyn is being fixed to use the right conditions: https://github.com/dotnet/roslyn/pull/37327


crummel:
CI failures were infra, kicked it off again.

PR #1159, opened 2019-07-29
Fix preview7 versions.

crummel:


crummel:
I missed some updates between the non-final and final core-sdk builds, this takes care of those.

PR #1164, opened 2019-08-01
Update to 3.0 preview8.

crummel:


crummel:
The remaining work here is:
- possibly a workaround for core-setup version numbers.
- updating prebuilts.

Other than that I expect this to pass.

patches/core-setup/0002-Try-ExcludeFromSourceBuild-in-windowsdesktop.patch
```diff
@@ -0,0 +1,24 @@
+From 1cc0f98a7afd3dd49a95ea8f4a31f80f6805f75c Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Sun, 4 Aug 2019 21:28:18 -0500
+Subject: [PATCH 2/2] Try ExcludeFromSourceBuild in windowsdesktop
```
dagood: I like the `[PATCH]` note here, making it clear which commits are from `git am`, but are the numbers useful? Are they kept up to date with some automation?

patches/websdk/0001-Add-PVP-import.patch
```diff
@@ -1,26 +1,26 @@
...
      <SystemXmlXmlDocumentPackageVersion>4.3.0</SystemXmlXmlDocumentPackageVersion>
+     <MicrosoftAspNetCoreAnalyzersPackageVersion>3.0.0-preview8.19381.6</MicrosoftAspNetCoreAnalyzersPackageVersion>
    </PropertyGroup>
 +
 +  <Import Project="$(DotNetPackageVersionPropsPath)" Condition="Exists('$(DotNetPackageVersionPropsPath)')" />
 +
    <PropertyGroup>
      <VersionPrefix>3.0.100</VersionPrefix>
-     <PreReleaseVersionLabel>preview7</PreReleaseVersionLabel>
+     <PreReleaseVersionLabel>preview8</PreReleaseVersionLabel>
```
dagood: Call this out in patch subject? Maybe this can be overridden as a `/p` property rather than in a patch?

repos/core-setup.proj
```diff
@@ -9,25 +9,30 @@
...
     <OverridePortable Condition="'$(TargetOS)' == 'OSX'">true</OverridePortable>
 
-    <BuildArguments>-ConfigurationGroup=$(Configuration)</BuildArguments>
-    <BuildArguments>$(BuildArguments) -PortableBuild=$(OverridePortable)</BuildArguments>
-    <BuildArguments>$(BuildArguments) -SkipTests=true</BuildArguments>
+    <BuildArguments>--restore --build</BuildArguments>
+    <BuildArguments>$(BuildArguments) --configuration $(Configuration)</BuildArguments>
+    <BuildArguments>$(BuildArguments) --ci</BuildArguments>
+    <BuildArguments>$(BuildArguments) /p:PortableBuild=$(OverridePortable)</BuildArguments>
     <BuildArguments Condition="$(Platform.Contains('arm'))">$(BuildArguments) -TargetArchitecture=$(Platform) -DisableCrossgen=true -CrossBuild=true</BuildArguments>
```
dagood: As a min-effort potential fix, can change this to new arg format, but don't know if it actually works:

```suggestion
     <BuildArguments Condition="$(Platform.Contains('arm'))">$(BuildArguments) /p:TargetArchitecture=$(Platform) /p:DisableCrossgen=true /p:CrossBuild=true</BuildArguments>
```

repos/core-setup.proj
```diff
@@ -9,25 +9,30 @@
...
-    <BuildArguments>$(BuildArguments) --</BuildArguments>
     <BuildArguments>$(BuildArguments) /p:BuildDebPackage=false</BuildArguments>
     <BuildArguments>$(BuildArguments) /p:BuildAllPackages=true</BuildArguments>
+    <BuildArguments>$(BuildArguments) /p:RestoreAllBuildRids=false</BuildArguments>
```
dagood: Filed https://github.com/dotnet/core-setup/issues/7629 to track being able to remove this.

patches/core-setup/0002-Try-ExcludeFromSourceBuild-in-windowsdesktop.patch
crummel: It looks like this was a default that was changed in Git.  We can standardize on one or the other but we'll probably see changes regularly until everyone is on a new enough version.

patches/websdk/0001-Add-PVP-import.patch
crummel: We already override this with an environment variable, but since this is also included here in websdk, the patch won't apply without updating it.  I will submit this patch to upstream after this is merged.

repos/core-setup.proj
crummel: Thanks!

repos/core-setup.proj
crummel: I'll change this but yeah, ARM is a crapshoot at this point anyway.

patches/core-setup/0002-Try-ExcludeFromSourceBuild-in-windowsdesktop.patch
dagood: Ah, that's fine then, just wanted to make sure this isn't adding pointless manual work on top of patch maintenance.

patches/websdk/0001-Add-PVP-import.patch
dagood: Oh! I just misread the double-diff. Carry on. 😄 

It's possible to shrink patch regions to avoid Git flagging the change as incompatible, but to me that only seems worthwhile in e.g. 2.1/2.2 where the patches are more stable and the patches won't be upstreamed.

repos/core-setup.proj
```diff
@@ -9,25 +9,30 @@
...
     <OverridePortable Condition="'$(TargetOS)' == 'OSX'">true</OverridePortable>
 
-    <BuildArguments>-ConfigurationGroup=$(Configuration)</BuildArguments>
-    <BuildArguments>$(BuildArguments) -PortableBuild=$(OverridePortable)</BuildArguments>
-    <BuildArguments>$(BuildArguments) -SkipTests=true</BuildArguments>
-    <BuildArguments Condition="$(Platform.Contains('arm'))">$(BuildArguments) -TargetArchitecture=$(Platform) -DisableCrossgen=true -CrossBuild=true</BuildArguments>
-    <BuildArguments>$(BuildArguments) --</BuildArguments>
+    <BuildArguments>--restore --build</BuildArguments>
+    <BuildArguments>$(BuildArguments) --configuration $(Configuration)</BuildArguments>
+    <BuildArguments>$(BuildArguments) --ci</BuildArguments>
```
adaggarwal: Although it probably wouldn't matter, but could FlagParameterPrefix replace '--', I am not sure though.

crummel: Yes, fixed.

crummel:
Windows started failing while Windows CI was down, logged https://github.com/dotnet/source-build/issues/1171 to fix.

PR #1166, opened 2019-08-02
Fix version templating property for Microsoft.DotNet.Cli.CommandLine

dseefeld:
Adding patch to fix templating property name so it picks up the source-built version of Microsoft.DotNet.Cli.CommandLine.


dseefeld:
Patch was added as part of Preview8 update.

PR #1170, opened 2019-08-08
Copy only the packages that we need the stage1 version of into the tarball

crummel:
Experimenting with a few different things to try and get the size of the tarball down.


crummel:
This gets it down to ~1.9GB, which seems like a good start.

crummel:
Windows started failing earlier while Windows CI was down, logged https://github.com/dotnet/source-build/issues/1171 to track.

crummel:
Windows was the only failing leg here so this is ready for review.

PR #1176, opened 2019-08-13
Separate out XPlat tasks that don't require core-setup so we can build them before cloning sources.

crummel:
This gets rid of one of the last few remaining manual steps for taking updates.


tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj
```diff
@@ -7,8 +7,8 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <CoreSetupPlatformAbstractionsPath Condition="'$(OfflineBuild)' != 'true'">$(ClonedSubmoduleDirectory)core-setup.$(CoreSetupPreferredVersion)/src/managed/Microsoft.DotNet.PlatformAbstractions/</CoreSetupPlatformAbstractionsPath>
-    <CoreSetupPlatformAbstractionsPath Condition="'$(OfflineBuild)' == 'true'">$(SubmoduleDirectory)core-setup.$(CoreSetupPreferredVersion)/src/managed/Microsoft.DotNet.PlatformAbstractions/</CoreSetupPlatformAbstractionsPath>
+    <CoreSetupPlatformAbstractionsPath Condition="'$(OfflineBuild)' != 'true'">$(ClonedSubmoduleDirectory)core-setup.$(coresetupGitCommitHash)/src/managed/Microsoft.DotNet.PlatformAbstractions/</CoreSetupPlatformAbstractionsPath>
```
dseefeld: Where does coresetupGitCommitHash come from?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat/WriteSourceRepoProperties.cs
```diff
@@ -175,16 +174,7 @@ private string GetCommitCount(string gitDir, string hash)
...
-            if (RuntimeEnvironment.OperatingSystemPlatform == Platform.Windows)
-            {
-                format = "%%cd";
-            }
-            else
-            {
-                format = "%cd";
-            }
-            return RunGitCommand($"log -1 --format={format} --date=short {hash}", gitDir: gitDir);
+            return RunGitCommand($"log -1 --format=%cd --date=short {hash}", gitDir: gitDir);
```
dseefeld: If we go back to supporting Windows, this will have to be added back?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj
crummel: `WriteSourceLinkMetadata` writes out an AllRepoVersions.props file that includes the hash and version number from eng/Version.Details.xml.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat/WriteSourceRepoProperties.cs
crummel: Kind of.  It would have to be messier since this can't rely on core-setup to tell it the platform anymore - we'd probably have to change this to a try-catch and attempt both ways.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj
dseefeld: OK.  Got it.  Thanks!

PR #1183, opened 2019-08-17
Remove some uneeded patches.

crummel:


crummel:
This gets rid of I believe all the ASP.NET patches we decided were questionable.  The baseline changes are expected because we're reverting to building on ASP.NET's expected target framework.

patches/aspnet-extensions/0004-Exclude-two-non-shipping-projects-from-source-build.patch
```diff
@@ -12,7 +12,7 @@ index 0198da33a..d358b96e7 100644
 --- a/src/TargetingPack/Microsoft.Internal.Extensions.Refs/ref/Microsoft.Internal.Extensions.Refs.csproj
 +++ b/src/TargetingPack/Microsoft.Internal.Extensions.Refs/ref/Microsoft.Internal.Extensions.Refs.csproj
 @@ -4,6 +4,7 @@
-     <TargetFramework>netcoreapp2.0</TargetFramework>
+     <TargetFramework>netcoreapp3.0</TargetFramework>
```
JunTaoLuo: Hmm why is this change needed if we are excluding this from source build?

crummel: Sorry, this is just the confusing double-patch notation.  `netcoreapp3.0` is what you actually have, previously we were changing it to `netcoreapp2.0` and then disabling it, now we just disable it.

PR #1185, opened 2019-08-20
Fix nuget-client build to fix dependency flow

dseefeld:


crummel:
Looks like we might need to patch NuGet to pin the MSBuild version to our ref assembly version.  Fix looks good in theory to me though.

crummel:
Did you mean to revert https://github.com/dotnet/source-build/commit/9b2927494af0a1c7515d2956021df39ec2128136?

dseefeld:
> Did you mean to revert [9b29274](https://github.com/dotnet/source-build/commit/9b2927494af0a1c7515d2956021df39ec2128136)?

Yes.  The change does not require dependency flow.

PR #1195, opened 2019-08-26
Fix MSBuild version to reference-package version

dseefeld:


dseefeld:
This change was made with 3.0 update.

PR #1213, opened 2019-08-30
Stop building MSBuild assemblies for .NET Fx

NikolaMilosavljevic:
Mostly changes in msbuild repo, but also some that were needed in few others to pass the build (sourcelink, fsharp)

Tests:
- smoke tests
- VS Code tests (debugging, intelisense, omnisharp)
- dotnet commands


adaggarwal:
Just wondering if `Version="15.7.179"` could be something that sourcelink patches pick up from source-build props? Otherwise LGTM.

NikolaMilosavljevic:
> Just wondering if `Version="15.7.179"` could be something that sourcelink patches pick up from source-build props? Otherwise LGTM.

Yes, this will be fixed in my pending sourcelink changes.

PR #1222, opened 2019-09-10
Do not build .NET Fx binaries in source build

NikolaMilosavljevic:
This cover all 3 binaries in aspnet-extensions repo.


NikolaMilosavljevic:
For some reason patch is not applicable. I'll send the new PR with updated changes.

Issue #1200, opened 2019-08-27
[Non-Deterministic behavior] CentOS7 Online CI leg failing infrequently


adaggarwal:
Fails at 'Run smoke tests in tarball stage' 

Example failure : [Link](https://dev.azure.com/dnceng/public/_build/results?buildId=327204) CentOS Online

![image](https://user-images.githubusercontent.com/22264147/63808376-69d2cf00-c8e5-11e9-937d-3d37b0a67c19.png)

adaggarwal:
Closing this for the time being as the issue is fairly infrequent and fetching ip info from docker containers in CI could involve tweaks to add ifconfig utility.

PR #1228, opened 2019-09-16
Do not build for NET Fx in source build - aspnet-xdt repo

NikolaMilosavljevic:
Removing a single NET Fx binary in source build


NikolaMilosavljevic:
'osx' and 'centos71 Online' failures are expected - these are failing for the same reasons in official builds, without my changes.

crummel:
OSX and CentOS Production are expected right now (OSX - prebuilts, CentOS production - unit tests), CentOS online is the intermittent failure we're having on smoke-test.  /cc @dseefeld for the smoke-test issue.

adaggarwal:
For the CentOS online, I think its again the ipv6 configuration issue. #1200

PR #1233, opened 2019-09-17
Add source built artifacts tarball

dseefeld:
Package up source-built artifacts and publish them with CI so they can be used for subsequent builds.

### DIFF
--- a/repos/package-source-build.proj
+++ b/repos/package-source-build.proj
@@ -4,7 +4,6 @@
 
   <PropertyGroup>
     <RepoApiImplemented>false</RepoApiImplemented>
-    <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
     <ProjectDirectory>$(SubmoduleDirectory)$(RepositoryName)/</ProjectDirectory>
     <SkipEnsurePackagesCreated>true</SkipEnsurePackagesCreated>
     <BuildCommand>$(ProjectDirectory)pack$(ShellExtension)</BuildCommand>
--- a/repos/package-source-build.proj
+++ b/repos/package-source-build.proj
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" DefaultTargets="Pack" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+
+  <PropertyGroup>
+    <RepoApiImplemented>false</RepoApiImplemented>
+    <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
+    <ProjectDirectory>$(SubmoduleDirectory)$(RepositoryName)/</ProjectDirectory>
+    <SkipEnsurePackagesCreated>true</SkipEnsurePackagesCreated>
+    <BuildCommand>$(ProjectDirectory)pack$(ShellExtension)</BuildCommand>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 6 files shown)


NikolaMilosavljevic:
Would this produce different results if build was started with plain ./build.sh, without /p:ArchiveDownloadedPackages=true? (This would be for non-CI scenario.)

dseefeld:
> Would this produce different results if build was started with plain ./build.sh, without /p:ArchiveDownloadedPackages=true? (This would be for non-CI scenario.)

No.  It should package up any of the source-built artifacts.  It doesn't depend on downloaded packages.

repos/package-source-build.proj
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" DefaultTargets="Pack" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+
+  <PropertyGroup>
+    <RepoApiImplemented>false</RepoApiImplemented>
+    <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
```
crummel: Think `OrchestratedManifestBuildName` is unused now.

src/package-source-build/pack.sh
```diff
@@ -0,0 +1,19 @@
...
+  source="$(readlink "$source")"
+
+  # if $source was a relative symlink, we need to resolve it relative to the path where the
+  # symlink file was located
+  [[ $source != /* ]] && source="$scriptroot/$source"
+done
+
+scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
+
+"$scriptroot/../../eng/common/build.sh" --pack -c Release --warnaserror false $@ /p:Projects="$scriptroot/package.proj"
```
crummel: Not sure if this is what you intended - this uses our bootstrap version of Arcade, not the version we build.  I'm fine with either since this is infra.

src/package-source-build/package.proj
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" DefaultTargets="Pack" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+  <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
+
+  <PropertyGroup>
+    <RepoApiImplemented>false</RepoApiImplemented>
+    <OrchestratedManifestBuildName>N/A</OrchestratedManifestBuildName>
+    <ProjectDirectory>$(SubmoduleDirectory)$(RepositoryName)/</ProjectDirectory>
+    <SkipEnsurePackagesCreated>true</SkipEnsurePackagesCreated>
```
crummel: Source-build-specific properties shouldn't be needed here, right?

repos/package-source-build.proj
dseefeld: Removed.

src/package-source-build/package.proj
dseefeld: Agreed.  Removed.

src/package-source-build/pack.sh
dseefeld: Is this any different than other repos?  i.e. repos have their own version of arcade eng/common scripts with correspond to the arcade version that they use.  But, when built with source-build, they use the source-built version of arcade.

dseefeld:
osx failure is expected.

PR #1234, opened 2019-09-17
Do not build .NET Fx binaries in source build - roslyn repo

NikolaMilosavljevic:
Few more changes for roslyn repo.


crummel:
This one I'm a bit concerned about - at one point Roslyn decided whether to include build props and targets based on the target framework, and we need those props and targets included for downstream builds to actually pick our Roslyn up instead of using the SDK compiler.  If you haven't looked at it yet, could you confirm that `Microsoft.Net.Compilers.Toolset` and family still include these props and get used downstream?

NikolaMilosavljevic:
> This one I'm a bit concerned about - at one point Roslyn decided whether to include build props and targets based on the target framework, and we need those props and targets included for downstream builds to actually pick our Roslyn up instead of using the SDK compiler. If you haven't looked at it yet, could you confirm that `Microsoft.Net.Compilers.Toolset` and family still include these props and get used downstream?

I've checked the Compilers.Toolset package and csc binary being used - everything looks fine, we are using what is built in source-build.

PR #1243, opened 2019-09-20
Update nuget-client to use source-built versions of msbuild.

dseefeld:


patches/nuget-client/0001-Pin-newtonsoft-json-to-a-specific-version.patch
```diff
@@ -0,0 +1,26 @@
...
+index 1332abb..52176e0 100644
+--- a/build/config.props
++++ b/build/config.props
+@@ -48,8 +48,7 @@
+ 
+   <!-- Dependency versions -->
+   <PropertyGroup>
+-    <NewtonsoftJsonVersionCore Condition="$(NewtonsoftJsonPackageVersion) == ''">9.0.1</NewtonsoftJsonVersionCore>
+-    <NewtonsoftJsonVersionCore Condition="$(NewtonsoftJsonPackageVersion) != ''">$(NewtonsoftJsonPackageVersion)</NewtonsoftJsonVersionCore>
++    <NewtonsoftJsonVersionCore>9.0.1</NewtonsoftJsonVersionCore>
```
NikolaMilosavljevic: Can this version be defined in common props file (for source-built packages)?

dseefeld: This should be a stable version, not floating, because we're building a newer version (12.0.2), which was getting picked up, but they're using functionality from the 9.0.1 version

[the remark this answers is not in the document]
dseefeld: No.  Its using functionality from 9.0.1.

PR #1255, opened 2019-09-24
Work around hanging unit tests.

crummel:
Make the timeout 40 minutes - they usually take 30 when successful but have been hanging to three hours.  Also turn continueOnError on so if they timeout the build will keep going.


dleeapho:
LGTM.

PR #1257, opened 2019-09-26
Switch to PROD pool

NikolaMilosavljevic:


.vsts.pipelines/builds/ci-internal.yml
```diff
@@ -1,5 +1,5 @@
 jobs:
 - template: matrix.yml
   parameters:
-    windowsPoolName: NetCoreInternal-Int-Pool
+    windowsPoolName: NetCoreInternal-Pool
```
adaggarwal: This might only affect windows pool 
cc @crummel

crummel:
There's one more reference here, I think: https://github.com/dotnet/source-build/blob/6df8641e6f585bc6566c0ded29b6773a3a726f3f/.vsts.pipelines/builds/matrix.yml#L9.

NikolaMilosavljevic:
> There's one more reference here, I think: 
>   
>     
>       source-build/.vsts.pipelines/builds/matrix.yml
>     
>     
>          Line 9
>       in
>       6df8641
>     
>     
>     
>     
> 
>         
>           
>            windowsPoolName: NetCorePublic-Int-Pool 
>         
>     
>   
> 
> .

I've updated the PR.

crummel:
/azp run
(I don't think this is a bad idea even if it wasn't the immediate problem)

PR #1258, opened 2019-09-26
Move packages instead of copy them to the blob feed to save space.

crummel:
This should save about 18GB while creating the tarball in CI.


crummel:
Closing in favor of https://github.com/dotnet/source-build/pull/1259.

PR #1259, opened 2019-09-26
Add an option to aggressively delete things the tarball doesn't need and turn it on for CI.

crummel:
This should not be used normally, as it will break the repo and delete local work for the sake of minimizing space.

### DIFF
--- a/build-source-tarball.sh
+++ b/build-source-tarball.sh
@@ -99,30 +99,30 @@ mkdir -p "$TARBALL_ROOT"
 # This list of extensions is everything over 6MB or so.
 if [ $MINIMIZE_DISK_USAGE -eq 1 ]; then
     find $SCRIPT_ROOT/bin/src \( -type f \( \
-        -iname *.dll -o \
-        -iname *.exe -o \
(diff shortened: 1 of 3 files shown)


crummel:
This take gives us 3GB to spare at the point we're using the most space, so still not great, but at least as an interim step it gets builds off the ground.  OSX is a separate issue I'm still working on.

build-source-tarball.sh
```diff
@@ -124,6 +176,11 @@ for srcDir in `find bin/src -name '.git' -type d`; do
 done
 popd
 
+# now we don't need bin/src at all
+if [ $MINIMIZE_DISK_USAGE -eq 1 ]; then
+    rm -rf "$SCRIPT_ROOT/bin/src/"
```
dseefeld: I think this deletes all production repo binlogs before they get copied in the next step.

crummel: Nice catch, thanks.  Removed this part.

crummel:
OSX failure is expected; see https://github.com/dotnet/source-build/pull/1262 for that fix.

PR #1270, opened 2019-10-01
Remove reference to Microsoft.CodeAnalysis,2.8.2 in arcade Packaging project

### LINKED ISSUE #1265: Prebuilt Removal: Investigate Microsoft.CodeAnalysis,2.8.2 in arcade
Arcade defines a static version package reference for Microsoft.CodeAnalysis,2.8.2 here:
https://github.com/dotnet/arcade/blob/14abaee3dba41fbe608431fb6a4f4b2435dcac33/src/Microsoft.DotNet.Build.Tasks.Packaging/src/Microsoft.DotNet.Build.Tasks.Packaging.csproj#L60-L61

This is causing multiple indirect prebuilts.  There is a comment which states:
`Need to reference version 2.8.2 of CodeAnalysis so that we don't have downgrade issues with System.Reflection.Metadata version`

This should be investigated to see if source-build can use a source-built version for this package.

dseefeld:
Fixes #1265


dseefeld:
@crummel @adaggarwal PTAL

PR #1278, opened 2019-10-02
Use updated uberclone to ignore cloning of source-build-ref-pkgs repo

dseefeld:


dseefeld:
OSX failure is expected.  Good to merge.

PR #1297, opened 2019-10-08
Fix arcade patch to update NuGet versions

dseefeld:


dseefeld:
Fixed with most recent Arcade update

PR #1305, opened 2019-10-10
Update MSBuild version to ref-pkg version in fsharp

dseefeld:


dseefeld:
OSX failure is a known issue.  Good to merge.

PR #1316, opened 2019-10-16
Update vstest projects to netcoreapp2.1 to eliminate prebuilts

dseefeld:


dseefeld:
OSX failure is known.

PR #1318, opened 2019-10-16
Add PackageVersionProps to env variables

dseefeld:
AspNetCore's build script (build.sh) builds eng/tools/RepoTasks before building the main project.  The build of this project doesn't use the passed in build params from source-build.  This change set DotNetPackageVersionPropsPath in env variables so the RepoTasks build can pick up the SourceBuilt version of Microsoft.Net.Compilers.Toolset.


dseefeld:
OSX failure is expected.

PR #1331, opened 2019-10-24
Add patches to fix Arcade prebuilts

dseefeld:


patches/arcade/0014-Import-PackageVersions-props-if-exists-and-update-Co.patch
```diff
@@ -0,0 +1,26 @@
...
+index 0389531..a8a2a0c 100644
+--- a/eng/Versions.props
++++ b/eng/Versions.props
+@@ -72,4 +72,8 @@
+     <MicrosoftDotNetMaestroClientVersion>1.1.0-beta.19454.3</MicrosoftDotNetMaestroClientVersion>
+     <MicrosoftDotNetSwaggerGeneratorMSBuildVersion>1.0.0-beta.19459.10</MicrosoftDotNetSwaggerGeneratorMSBuildVersion>
+   </PropertyGroup>
++  <Import Project="$(DotNetPackageVersionPropsPath)" Condition="'$(DotNetPackageVersionPropsPath)' != ''" />
++  <PropertyGroup>
++    <MsbuildTaskMicrosoftCodeAnalysisCSharpVersion Condition="'$(DotNetBuildFromSource)' == 'true'">$(MicrosoftCodeAnalysisCSharpVersion)</MsbuildTaskMicrosoftCodeAnalysisCSharpVersion>
```
crummel: Since we're already patching here I don't think this matters, but I've been doing these one-off properties here: https://github.com/dotnet/source-build/blob/9dd007e685b5e76724492688d347461095a0654d/dir.props#L198.

PR #1342, opened 2019-11-01
Remove app host prebuilt from aspnetcore

dseefeld:


patches/aspnetcore/0011-Remove-samples-and-tests-from-ProjectToBuild.patch
```diff
@@ -0,0 +1,27 @@
...
+
+diff --git a/eng/Build.props b/eng/Build.props
+index fa5e1d2..5e0ebf8 100644
+--- a/eng/Build.props
++++ b/eng/Build.props
+@@ -119,6 +119,10 @@
+ 
+         <ProjectToBuild Condition=" '$(BuildJava)' == 'true'" Include="@(JavaProjects)" Exclude="@(ProjectToExclude)" />
+         <ProjectToExclude Condition=" '$(BuildJava)' != 'true'" Include="@(JavaProjects)" />
++        <ProjectToExclude Condition=" '$(DotNetBuildFromSource)' == 'true'" Include="$(RepoRoot)src\**\samples\**\*.*proj" />
```
JunTaoLuo: This shouldn't be needed. The logic is encoded here: https://github.com/aspnet/AspNetCore/blob/master/Directory.Build.targets#L14. Is that not working in source build?

dseefeld: Agreed.  I removed the patch.  This was an in-progress change.

PR #1357, opened 2019-11-08
PlatformAbstractions reference should be included for source-build only

NikolaMilosavljevic:
Conditioning PlatformAbstractions reference for source-build only.

I had to redo 2 patches, touching the same file, one was dependent on the other.


crummel:
OSX failure looks like a network issue - if this last retry doesn't work I'm okay with merging anyway.

PR #1362, opened 2019-11-13
Initial support for FreeBSD.

wfurt:
contributes to #1139.
With limited ability to commit to repose, I focus on passing parameters to the build.
Using patches when necessary. 

cc: @joperator @jasonpugsley  @am11


patches/corefx/0006-Do-not-fail-on-test-chmod.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/external/runtime/runtime.depproj b/external/runtime/runtime.depproj
+index 3ce0ed89bc..1298019efd 100644
+--- a/external/runtime/runtime.depproj
++++ b/external/runtime/runtime.depproj
+@@ -60,7 +60,7 @@
+           SkipUnchangedFiles="true"
+           UseHardlinksIfPossible="$(UseHardlink)" />
+ 
+-    <Exec Command="chmod +x $(TestHostRootPath)%(DotnetExe.Filename)%(DotnetExe.Extension)" Condition="'$(OS)' != 'Windows_NT'"/>
++    <Exec Command="chmod +x $(TestHostRootPath)%(DotnetExe.Filename)%(DotnetExe.Extension) || true" Condition="'$(OS)' != 'Windows_NT'"/>
```
omajid: Has someone from corefx looked at this change? We are (or were) hitting this on Fedora 32 and Arch Linux as well: https://github.com/dotnet/source-build/issues/1310#issuecomment-543163427. Maybe this should be merged into corefx if it's the correct general fix.

wfurt: yes, that is my plan. but I'll do it after repo merge. I did testing and we don't need it at all any more. The dotnet is executable as it should on its own.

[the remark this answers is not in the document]
wfurt: This normally comes from TestHost and matching runtime entry. The problem is that if you get standard package (or if it lives in your nuget cache) it does not have platform reference and then dotnet is not restored and chmod fails. 
At this point you may not even want to run tests but it breaks build it self. Conceptually I feel that is not correct. 
To get to point where you can run tests on FreeBSD, I copy bootstrap dotnet over. (and related native libraries) You can pass --buildtests to top level script to build but not run tests @am11. This will go away and everything will work as other platforms once we have builds and packages flowing.

wfurt:
OSX failures seems like infrastructure issue
```
It was not possible to find any compatible framework version
  The specified framework 'Microsoft.NETCore.App', version '3.0.0' was not found.
    - Check application dependencies and target a framework version installed at:
        /Users/runner/.dotnet
```

crummel:
Yup, OSX failure is expected, I'm looking at it.

patches/corefx/0006-Do-not-fail-on-test-chmod.patch
crummel: Does the build work without this change?  I wouldn't expect test code to be getting hit in our normal build.

omajid: @crummel The build fails here for Fedora 32 and Arch Linux too: https://github.com/dotnet/source-build/issues/1310#issuecomment-543163427. This code path is definitely invoked in source-build.

wfurt: no, it does not build @crummel 
It seems like it got moved to Arcade in master. 
I'll take a look after repo merge is over. 
cc: @safern

wfurt: BTW I think this is old time paranoia when we did not know how to set and preserve executable bits on Unix.

crummel:
@wfurt Is this ready to go in?

repos/vstest.proj
```diff
@@ -3,10 +3,13 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
 
   <PropertyGroup>
+    <OverrideTargetRid>$(TargetRid)</OverrideTargetRid>
+    <OverrideTargetRid Condition="'$(TargetOS)' == 'FreeBSD'">freebsd-x64</OverrideTargetRid>
```
omajid: Sorry if this is already obvious to everyone else: but what is the original value of `TargetRid` here that's being overridden?

crummel: We generate this at the very beginning of the build here: https://github.com/dotnet/source-build/blob/8ebb1219998b0c360ec9ba3381c9c01255d0a1ef/tools-local/init-build.proj#L111
That calls into [our task](https://github.com/dotnet/source-build/blob/8ebb1219998b0c360ec9ba3381c9c01255d0a1ef/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/GetHostInformation.cs) which in turn is calling into [core-setup code](https://github.com/dotnet/source-build/blob/8ebb1219998b0c360ec9ba3381c9c01255d0a1ef/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj#L16) to get the RID the same way the product does.  This will by default be a non-portable RID but it looks like FreeBSD, like Windows and OSX, always uses a portable RID - @wfurt will have to be the one to explain if you're interested in why that is.

omajid: Oh, so this is just forcing a portable RID instead of the non-portable RID? That answers my question.

crummel: Yes, that's correct.

wfurt: yes, we decided that last year but that change never made it to all repose. Also repos now are inconsistent so source-build is the glue to force same and expect RID. 
One day we should make it simpler but for now we do want portable RID as all versions are binary and API compatible. (at least for parts we need)

wfurt:
it would be great if anybody can confirm that this works. But this should be good to go as it works at least in cases I tested and it does not break other platforms.

crummel:
@dseefeld think we can go ahead with this?  we can't verify that it remains working without CI but it doesn't seem to break anything else.

dseefeld:
> @dseefeld think we can go ahead with this? we can't verify that it remains working without CI but it doesn't seem to break anything else.

Yes.  I think we should merge this.

PR #1365, opened 2019-11-15
Remove source-built dir from tarball in favor of prev source-built archive

### LINKED ISSUE #1358: Remove copied nupkgs from tarball
There are a set of nupkgs that are copied into packages/source-built when building a tarball.  These should be removed in favor of packages from the previously source-built artifacts archive.

dseefeld:
- Remove copying of assets to source-built dir in tarball
- On initial tarball build, lookup boostrap versions of arcade, sourcelink
and DotNetHost from previously source-built archive and use those
versions when building arcade, sourcelink and corefx.

Fixes #1358


dseefeld:
This work is in 3.1 with #1397.  Closing this for 3.0.

PR #1367, opened 2019-11-18
Build tasks in tarball

### LINKED ISSUE #1351: tools-local tasks should build in tarball rather than being copied
The source-build task binaries in tools-local get copied to the tarball rather than being built in the tarball.  These should be built rather than copied to eliminate these binaries.

adaggarwal:
Fixes #1351


build-source-tarball.sh
```diff
@@ -173,6 +173,11 @@ find $TARBALL_ROOT/src \( -type f \( \
     -iname *.zip -o \
     -iname *.nupkg \) \) -exec rm {} \;
 
+#Remove binaries from tools-local as we don't want them to end up in tarball, the intent is to have them built within the context of tarball
```
dseefeld: Maybe you want to remove the bin and obj directories from Tasks and Task.XPlat instead?  There are dlls and pdbs in the obj directory.

adaggarwal: Thanks for the catch.

crummel: Could we remove these from the tarball copy instead like here? https://github.com/dotnet/source-build/blob/1e937de0b9e256bc9b4107b2e9ab3f3f9b067a69/build-source-tarball.sh#L206

Then we don't trash stuff from someone's working directory that we don't need to.

adaggarwal:
> One small thing, otherwise LGTM.

Good idea.

PR #1383, opened 2019-11-21
Merge 3.0 => 3.1

crummel:


support/allowed-prebuilts.txt
```diff
@@ -0,0 +1 @@
+
```
dseefeld: There are still some roslyn prebuilts in 3.1.

support/additional-prebuilts-to-delete.txt
```diff
@@ -0,0 +1,40 @@
+# this is a list of packages that end up in the tarball's prebuilts folder but are not actually needed.
```
dseefeld: Will this list need to be updated for 3.1?

support/allowed-prebuilts.txt
crummel: I disabled actually prebuilt enforcement with https://github.com/dotnet/source-build/pull/1383/commits/dc9a11ced5fd7fbf8bed5d8d9b460f0c708ed54c, I figured once we hit zero it should be easy to flip that back on and with your improvement we'll only need to run CI once to find all the extras.

support/additional-prebuilts-to-delete.txt
crummel: Yes, but I was going to wait until we hit zero in the baseline, I think that will be easier.

support/allowed-prebuilts.txt
dseefeld: Got it.  Makes sense.

crummel:
CentOS Production passed on re-run but AzDo isn't updating the status (https://dev.azure.com/dnceng/public/_build/results?buildId=434521&view=logs&jobId=935c00b8-5de8-597d-a783-42e7f079746c).  Merging.

PR #1390, opened 2019-11-22
Add tool bootstrapping script

dseefeld:
Add a script for bootstrapping the toolset in a distro.

### DIFF
--- a/support/tarball/tool-bootstrapping/bootstrap.sh
+++ b/support/tarball/tool-bootstrapping/bootstrap.sh
@@ -209,14 +209,12 @@ fi
 
 # Check network access
 if command -v wget 2>/dev/null; then
-    wget -q --spider http://www.microsoft.com
-    if [ $? -eq 0 ]; then
+    if wget -q --spider http://www.microsoft.com 2>/dev/null; then
         echo "Network can be off for bootstrapping..."
         read -n1 -r -p "Press space to continue..." key
         echo ""
     fi
-    wget -q --spider http://www.microsoft.com
-    if [ $? -eq 0 ]; then
+    if wget -q --spider http://www.microsoft.com 2>/dev/null; then
         LogMessage "Network is ON"
     else
         LogMessage "Network is OFF"
--- a/support/tarball/tool-bootstrapping/bootstrap.sh
+++ b/support/tarball/tool-bootstrapping/bootstrap.sh
@@ -18,8 +18,8 @@ usage() {
 
 function LogMessage {
     local size=(`df -h .`)
-    local message="[${size[10]} - ${size[11]}] : $1"
-    #local message=$1
+    #local message="[${size[10]} - ${size[11]}] : $1"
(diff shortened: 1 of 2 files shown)


support/tarball/tool-bootstrapping/bootstrap.sh
```diff
@@ -0,0 +1,231 @@
...
+    sdkVer=`ls $refPkgsDir/.dotnet/sdk`
+    LogMessage "    Sdk version = [$sdkVer]"
+    sed -i "s|\"dotnet\": \".*|\"dotnet\": \"$sdkVer\"|g" $refPkgsDir/global.json
+
+    LogMessage "Build stage1 ref-pkgs"
+    cd $refPkgsDir
+    ./build.sh
+
+    LogMessage "Package built reference packages into new tarball"
+    cd $refPkgsDir/artifacts/reference-packages
```
crummel: Suggestion: end with the current directory either the same as what we started with, or a consistent current directory across all three steps.  It can get confusing if the script leaves you somewhere you didn't expect.

support/tarball/tool-bootstrapping/bootstrap.sh
```diff
@@ -0,0 +1,231 @@
...
+    echo "exiting..."
+    exit 1
+fi
+
+if [ ! -d "$bootstrapDir" ]; then
+    mkdir "$bootstrapDir"
+fi
+
+# Check network access
+wget -q --spider http://www.microsoft.com
```
crummel: I know this is just a quick check but do we want to check that wget exists before relying on it?  I think several of the dotnet-buildtools-prereqs images don't include it.

support/tarball/tool-bootstrapping/bootstrap.sh
```diff
@@ -0,0 +1,231 @@
...
+    echo "  Steps:"
+    echo "      1. Build stage1 sdk and arcade from tarball using stage0 sdk and ref-pkgs in the tarball"
+    echo "      2. Build final ref-pkgs using stage1 sdk and arcade"
+    echo "      3. Build final sdk using final ref-pkgs and stage1 sdk and arcade"
+    echo ""
+}
+
+function LogMessage {
+    local size=(`df -h .`)
+    #local message="[${size[10]} - ${size[11]}] : $1"
```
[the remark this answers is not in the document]
dseefeld: Fixed

support/tarball/tool-bootstrapping/bootstrap.sh
dseefeld: I'll do that.  Thanks.

support/tarball/tool-bootstrapping/bootstrap.sh
dseefeld: I think after the call to the function, it is left in the same dir that it was in when the function was called.  But, I changed to pushd/popd just to be clear.

PR #1395, opened 2019-11-25
Add more build artifacts and automatically publish to BAR.

crummel:
- Upgrade Arcade scripts.
- Add build steps to:
  - Collect all artifacts into one tarball (so it includes all RID-specific as well as non-RID-specific packages).
  - Publish this tarball (and an empty nupkg for dependency uptake) to BAR.
- Add offline legs for Ubuntu and Debian so we consistently use the Offline-built packages for this tarball.
- Add more artifacts to the list to publish:
  - runtime and SDK product tarballs
  - generated git-info and PackageVersions.props
  - list of all files in the repo after the build

Example builds: (old) https://dev.azure.com/dnceng/internal/_build/results?buildId=438340, https://dev.azure.com/dnceng/internal/_build/results?buildId=438343, (latest) https://dev.azure.com/dnceng/internal/_build/results?buildId=438639, https://dev.azure.com/dnceng/internal/_build/results?buildId=438643.

@dagood, added you if you don't mind taking a look, I think you're more familiar with this than any of us.


.vsts.pipelines/jobs/collect-artifacts.yml
```diff
@@ -0,0 +1,104 @@
...
+          # do CentOS last to prefer those versions of any duplicates
+          for os in portable debian fedora ubuntu centos; do
+            # DownloadArtifact creates another directory - use '*' to not deal with this
+            tar xzf ../$os/*/Private.SourceBuilt.Artifacts*.tar.gz
+            rm -rf ../$os
+          done
+          echo $(Build.SourceVersion) > .version
+          echo $(Build.BuildNumber) > .buildnumber
+          echo $(Build.BuildId) > .buildid
+          tar czf ../Private.SourceBuilt.Artifacts.$version.tar.gz * .version .buildnumber .buildid
```
dseefeld: This combines the nupkgs into the tarball, but doesn't combine the corresponding PackageVersions.props.  I'm not sure that it's an issue, though.

.vsts.pipelines/jobs/ci-linux.yml
```diff
@@ -77,6 +77,23 @@ jobs:
...
+      $(docker.run) $(docker.drop.map) $(docker.src.map) $(docker.src.work) $(imageName) /bin/bash -c "
+        mkdir -p /drop/artifacts/source-build
+        cp ./bin/*/*/*.tar.gz /drop/artifacts/source-build/
+        cp ./bin/*/*/runtime/*.tar.gz /drop/artifacts/source-build/
+        cp ./bin/obj/*/*/blobs/aspnetcore/Runtime/*/*.tar.gz /drop/artifacts/source-build/
+        cp ./bin/obj/*/*/blob-feed/packages/PackageVersions.props /drop/artifacts/source-build/
+        cp ./bin/obj/x64/Release/*.props /drop/artifacts/source-build/
+        cp -r ./bin/git-info /drop/artifacts/source-build/
+        cp -r ./bin/obj/*/*/build-info /drop/artifacts/source-build/"
+      du -h $(rootDirectory) | sort -h | tail -n 50
```
dagood: I wonder if it would be better to have in the build itself... this seems like a useful diagnosis artifact someone can give us when source-build doesn't work for them.

The paths also might also be clearer in that context--using the project props rather than writing out full names in multiple places.

.vsts.pipelines/jobs/collect-artifacts.yml
crummel: I feel like we should do that but I don't think we currently have a use for it.  We also can work around it using something like
https://github.com/dotnet/source-build/blob/8d17a1046da907571c285ae144db4dc48ac9c9d2/dir.props#L211

Any opinion on whether we should fix this now or log an issue and do it if it ever comes up?

.vsts.pipelines/jobs/ci-linux.yml
```diff
@@ -255,6 +291,7 @@ jobs:
   # Copy artifacts to staging - Copy to VSTS owned folder is done outside of docker so copied files
   # have correct ownership so VSTS can clean them up later.
   - task: CopyFiles@2
+    displayName: Copy artifacts from Docker-owned staging directory to VSTS-owned artifacts directory
```
dagood:
```suggestion
    displayName: Copy artifacts from Docker-owned staging directory to AzDO-agent-owned artifacts directory
```

.vsts.pipelines/jobs/collect-artifacts.yml
```diff
@@ -0,0 +1,104 @@
...
+
+  pool: ${{ parameters.pool }}
+
+  steps:
+  - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+    - task: DownloadBuildArtifacts@0
+      displayName: Download CentOS artifacts
+      inputs:
+        artifactName: Tarball centos71 Offline
+        downloadPath: '$(Build.StagingDirectory)/centos'
```
dagood: IMO you should use an [`each`](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#iterative-insertion) and pass a list of objects to this template to keep this out of the implementation yaml. You can get this down to only have a a single input list of objects that you reprocess to create the `dependsOn` and the sequence of download steps. The overall point being to keep the platform listings in a single file.

.vsts.pipelines/jobs/collect-artifacts.yml
```diff
@@ -0,0 +1,104 @@
...
+      inputs:
+        artifactName: Tarball centos71 Offline Portable
+        downloadPath: '$(Build.StagingDirectory)/portable'
+
+    - task: Bash@3
+      inputs:
+        targetType: 'inline'
+        workingDirectory: '$(Build.StagingDirectory)'
+        script: |
+          version=`echo centos/*/Private.SourceBuilt.Artifacts*.tar.gz | sed 's/\.tar\.gz$//' | sed 's/^centos\/.*\/Private\.SourceBuilt\.Artifacts\.//'`
```
dagood: `$(` is preferred over <code>`</code> generally. But this is also quite complex to be here and something that might belong as build logic (where I would *think* version would be accessible?).

.vsts.pipelines/jobs/collect-artifacts.yml
```diff
@@ -0,0 +1,104 @@
...
+          done
+          echo $(Build.SourceVersion) > .version
+          echo $(Build.BuildNumber) > .buildnumber
+          echo $(Build.BuildId) > .buildid
+          tar czf ../Private.SourceBuilt.Artifacts.$version.tar.gz * .version .buildnumber .buildid
+          rm -rf *
+          mv ../Private.SourceBuilt.Artifacts.*.tar.gz .
+      displayName: Collect RID-specific packages into one tarball
+      condition: succeeded()
+    - task: PowerShell@2
```
dagood: A newline between each step makes them significantly more readable to me.

.vsts.pipelines/jobs/collect-artifacts.yml
```diff
@@ -0,0 +1,104 @@
...
+            # DownloadArtifact creates another directory - use '*' to not deal with this
+            tar xzf ../$os/*/Private.SourceBuilt.Artifacts*.tar.gz
+            rm -rf ../$os
+          done
+          echo $(Build.SourceVersion) > .version
+          echo $(Build.BuildNumber) > .buildnumber
+          echo $(Build.BuildId) > .buildid
+          tar czf ../Private.SourceBuilt.Artifacts.$version.tar.gz * .version .buildnumber .buildid
+          rm -rf *
+          mv ../Private.SourceBuilt.Artifacts.*.tar.gz .
```
dagood: It isn't clear to me what this artifact will be used for... sounds like more than diag if it's ending in BAR, so I'm a little concerned since point of source-build is that you can't get a combined artifact like this.

.vsts.pipelines/jobs/collect-artifacts.yml
```diff
@@ -0,0 +1,104 @@
...
+      displayName: Collect RID-specific packages into one tarball
+      condition: succeeded()
+    - task: PowerShell@2
+      inputs:
+        targetType: 'filePath'
+        filePath: $(Build.SourcesDirectory)\publish.ps1
+        arguments: /p:OfficialBuildId=$(Build.BuildNumber) /p:AzureAccountName=$(publish.blobStorage.account) /p:ContainerName=$(publish.blobStorage.container) "/p:AzureAccessToken=$(publish.blobStorage.accessToken)" "/p:ManifestName=$(BUILD.REPOSITORY.NAME)" "/p:ManifestBuildId=$(BUILD.BUILDID)" "/p:ManifestBranch=$(BUILD.SOURCEBRANCHNAME)" /p:ManifestCommit=$(Build.SourceVersion)
```
dagood: Huge line, consider `>-` or similar to split it nicely onto more. (I refer to https://yaml-multiline.info/ usually.)

eng/Versions.props
```diff
@@ -4,6 +4,11 @@
   <PropertyGroup>
     <VersionPrefix>0.1.0</VersionPrefix>
     <PreReleaseVersionLabel>alpha.1</PreReleaseVersionLabel>
+    <VersionSuffix>3.1.100-$(PreReleaseVersionLabel)</VersionSuffix>
+    <VersionSuffix Condition="'$(BUILD_BUILDNUMBER)' != ''">$(VersionSuffix).$(BUILD_BUILDNUMBER)</VersionSuffix>
+    <VersionSuffix Condition="'$(BUILD_SOURCEVERSION)' != ''">$(VersionSuffix)+$(BUILD_SOURCEVERSION)</VersionSuffix>
+    <VersionSuffix Condition="'$(BUILD_SOURCEVERSION)' == '' and '$(GitCommitHash)' != ''">$(VersionSuffix)+$(GitCommitHash)</VersionSuffix>
```
dagood: Arcade normally sets this based on `MajorVersion` `MinorVersion` etc.

eng/publishPackages/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dagood: These shouldn't be necessary anymore, might as well not add them to any of the new projects since it's already gone on `Private.SourceBuilt.Artifacts.csproj`.

eng/publishPackages/publish.proj
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.DotNet.Arcade.Sdk">
+  <PropertyGroup>
+    <ArtifactsPkgDir>$(RepoRoot)artifacts/packages/</ArtifactsPkgDir>
+    <ManifestDir>$(RepoRoot)artifacts/asset-manifests/</ManifestDir>
+    <TarballDir>$(RepoRoot)artifacts/tarball/</TarballDir>
+    <RestoreSources>
+      $(RestoreSources);
+      https://dotnetfeed.blob.core.windows.net/dotnet-tools-internal/index.json;
```
dagood: This might need to be a nuget.config to allow for internal servicing.

.vsts.pipelines/jobs/collect-artifacts.yml
dseefeld: I'm assuming that all of the packages will be the same name/version with the exception of the runtime packages that are rid specific.  I don't think these will use the version tag, so I think it will be OK to leave it.

crummel: Sounds good for now, logged https://github.com/dotnet/source-build/issues/1423.

eng/Versions.props
crummel: Changed this to use proper Arcade versioning.

eng/publishPackages/publish.proj
crummel: Do you know if it will fall back to the root NuGet.config or do we need a separate one for this?

crummel:
Superseded by #1784 and #1785.

PR #1415, opened 2019-12-03
Upgrade SourceLink to 1.0.0 release version.

crummel:
Ran into a couple issues that makes this less straightforward:
- `MS.SourceLink.Vsts.Git` was renamed to `MS.SourceLink.AzureDevOpsServer.Git`, but most repos haven't taken this change yet.  I reverted the name in SourceLink.
- Some repos were not properly uptaking the built version of SourceLink - added `DotNetPackageVersionPropsPath` to a few more repos, and changed around the way we write out the bootstrapping/built `MicrosoftSourceLinkVersion` property.


dir.props
```diff
@@ -206,6 +206,12 @@
     <ExtraPackageVersionPropsPackageInfo Include="VSRedistCommonAspNetCoreTargetingPackx6430PackageVersion" Version="$(aspnetcoreOutputPackageVersion)" />
     <!-- OSX needs the OSX version instead of Linux.  We don't have a lot of flexibility in how we output these properties so we're relying on the previous one being blank if the Linux version of the package is missing.  -->
     <ExtraPackageVersionPropsPackageInfo Include="MicrosoftAspNetCoreAppRuntimePackageVersion" Version="%24(MicrosoftAspNetCoreAppRuntimeOsxX64PackageVersion)" DoNotOverwrite="true" />
+
+    <!-- property used by Arcade to determine what version of SourceLink to use -->
+    <ExtraPackageVersionPropsPackageInfo Include="MicrosoftSourceLinkVersion" Version="%24(MicrosoftSourceLinkCommonPackageVersion)" />
+    <!-- if the built version is available overwrite it instead -->
+    <ExtraPackageVersionPropsPackageInfo Include="MicrosoftSourceLinkVersion" Version="%24(MicrosoftSourceLinkCommonPackageVersion)" />
```
dseefeld: This line is duplicated.

dir.props
```diff
@@ -206,6 +206,12 @@
...
+    <ExtraPackageVersionPropsPackageInfo Condition="'$(OfflineBuild)' == 'true'" Include="MicrosoftSourceLinkVersion" Version="$(SOURCE_LINK_BOOTSTRAP_VERSION)" DoNotOverwrite="true" />
```
dseefeld: Something seems wrong here.  I'm not sure this is doing anything, because DoNotOverwrite is set to true and it is always set in the line above.  Also, I'm not sure of the intent here.  It appears that the intent is to always set the SourceLink version to the bootstrap version if we're building offline, but that doesn't seem right.

crummel: What I'm going for here is that if we've built SourceLink already, we should always prefer that version, either online or offline.  If we haven't built SourceLink, then `%24(MicrosoftSourceLinkCommonPackageVersion)` will be empty, and we'll use `SOURCE_LINK_BOOTSTRAP_VERSION` instead.  The bootstrap version is actually the previously-source-built version (from [here](https://github.com/dotnet/source-build/blob/cfb90a45c0c996e2b1fac24778586b23aa88bc2d/support/tarball/build.sh#L24)), so we'll have that available to use for offline builds.

crummel: I'll try to comment this better.

dir.props
crummel: Thanks, removed.

dir.props
dseefeld: But, I'm saying I don't think the "OfflineBuild" line will do anything in either case.  In the first case, if SourceLink is not built (which means we're building SourceLink) it gets the version from it's proj file.  In the second case, SourceLink is built and the first line is relevant.

crummel: Oh, I see.  Yeah, let me try getting rid of that, that should work.

PR #1433, opened 2020-01-08
Always build portable runtime in addition to non-portable.

crummel:
This allows somebody to build source-build and create portable projects offline in addition to non-portable.  It also makes life easier for bootstrapping, as the portable bits serve as a fallback for all RIDs.

- Copy `coreclr.proj`, `corefx.proj`, and `core-setup.proj` to create additional portable versions that are included as references in the non-portable Linux builds.  These are just always forced to portable.
- Add extra entries to Version.Details.xml.  These don't create extra source directories because the repo URLs are the same for both entries, so we build the portable versions first, copy any packages with `linux` in the name, then delete `artifacts` and `bin` folders, which seems to work reasonably well to reset the repo.

### DIFF
--- a/repos/core-setup-portable.proj
+++ b/repos/core-setup-portable.proj
@@ -2,50 +2,10 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <SourceDirectory>core-setup</SourceDirectory>
+    <BuildingPortableRuntime>true</BuildingPortableRuntime>
   </PropertyGroup>
 
-  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
-
-  <PropertyGroup>
-    <BuildArguments>$(FlagParameterPrefix)restore $(FlagParameterPrefix)build</BuildArguments>
-    <BuildArguments>$(BuildArguments) $(FlagParameterPrefix)configuration $(Configuration)</BuildArguments>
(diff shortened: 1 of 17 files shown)


repos/core-setup-portable.proj
```diff
@@ -0,0 +1,93 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dseefeld: It seems like most of the code in the portable proj files is copied from the non-portable proj file.  Is there any way to remove the duplication and share it instead?

crummel: I think there's a couple ways we could do this:
1. Rather than including this as a separate project with a dependency, do a bit more build shuffling so we call `<MSBuild Projects=repo.proj />` and add `AdditionalProperties="PortableBuild=true"` sometimes.  This changes a lot outside just these portable projects and gets in to managing MSBuild contexts.
2. Alternatively, change our dependency resolution so we can say "we do want to build this project twice, once with these properties."
3. Inside the `repo.proj` files, re-call the same file with `PortableBuild=true`.  This seems like it would have the same problem with MSBuild contexts but at least wouldn't duplicate any of the command-line and other local property stuff.
4. Inside the `repo.proj` files, have an additional build target that does the portable build when necessary.  No MSBuild context problems, but we would still duplicate some things because we'd have to build up separate portable and non-portable command lines.

Out of these, really only #4 doesn't seem worse than this to me - is there anything else we could do?  Or is one of the other better?

dseefeld: How about putting the props in a common props file and targets in a common targets file and import them?

crummel: Yeah, I think that could work, let me try that.

PR #1445, opened 2020-01-15
Update to Jan 2019 2.1 release.

crummel:


crummel:
Whether OSX passes seems dependent on which machine we end up getting, so I'll retry a couple times and merge without it if we keep having problems.

PR #1448, opened 2020-01-16
core-sdk: Add target rid to AppHostRuntimeIdentifiers

### LINKED ISSUE #1444: [3.0/3.1] GenerateBundledVersions.targets needs to include target rid in AppHostRuntimeIdentifiers
With previous release we had [a patch](https://github.com/dotnet/source-build/pull/1231/files) that got rid of the target rid from the RuntimePackRuntimeIdentifiers.

With the servicing update this is already the case (part of https://github.com/dotnet/core-sdk/commit/0f52aed34c765c54fbb7affad075313ae2390e2f).
The target rid is now missing in the AppHostRuntimeIdentifiers.

cc @nguerrera @dagood @crummel @omajid

tmds:
Fixes #1444 for 3.1

### DIFF
--- a/patches/core-sdk/0008-Add-target-rid-to-AppHostRuntimeIdentifiers.patch
+++ b/patches/core-sdk/0008-Add-target-rid-to-AppHostRuntimeIdentifiers.patch
@@ -1,0 +1,37 @@
+From 62020f2c4fdf2c0389b4106cac9286f1c3a9eb7c Mon Sep 17 00:00:00 2001
+From: Tom Deseyn <tom.deseyn@gmail.com>
+Date: Thu, 16 Jan 2020 08:37:34 +0100
+Subject: [PATCH] Add target rid to AppHostRuntimeIdentifiers
+
+---
+ src/redist/targets/GenerateBundledVersions.targets | 7 ++++++-
+ 1 file changed, 6 insertions(+), 1 deletion(-)
+
+diff --git a/src/redist/targets/GenerateBundledVersions.targets b/src/redist/targets/GenerateBundledVersions.targets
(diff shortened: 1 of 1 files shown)


tmds:
cc @nguerrera @dagood @crummel @omajid @dseefeld

Also verified this one against core-sdk repo.

patches/core-sdk/0008-Add-target-rid-to-AppHostRuntimeIdentifiers.patch
```diff
@@ -0,0 +1,37 @@
...
++
+       <AspNetCore30RuntimePackRids Include="
+         win-x64;
+         win-x86;
+@@ -168,7 +173,7 @@ Copyright (c) .NET Foundation. All rights reserved.
+                       TargetFramework="netcoreapp3.1"
+                       AppHostPackNamePattern="Microsoft.NETCore.App.Host.**RID**"
+                       AppHostPackVersion="$(_NETCoreAppPackageVersion)"
+-                      AppHostRuntimeIdentifiers="@(NetCoreRuntimePackRids, '%3B')"
++                      AppHostRuntimeIdentifiers="@(NetCoreAppHostPackRids, '%3B')"
```
tmds: This is the change for 3.1.

It's possible there is an apphost also for 3.0 (from installing the 3.0 sdk), but we can't add it to the 3.0 `AppHostRuntimeIdentifiers` because that would cause a failure in case it isn't installed.
We could make the sdk look into the packs folder so it can figure out itself what apphosts are available locally. Then we don't have to add them explicitly here either.

crummel:
Thanks Tom!

PR #1455, opened 2020-01-22
Allow users to supply their own SDK for bootstrapping

### LINKED ISSUE #1435: Allow user to bring their own SDK
- Skip downloading bootstrap SDK
- Update global.json?
- Flow correct SDK path through build

crummel:
Fixes #1435.


crummel:
We're quite close to where we could generate the tarball without building at all (https://github.com/dotnet/source-build/issues/831), so that would help get rid of some of this.  I was planning on trying the straightforward version of that and seeing if it works soon.

build-source-tarball.sh
```diff
@@ -10,6 +10,7 @@ usage() {
     echo "  --skip-prebuilt-check              do not confirm that all prebuilt packages used are either reference packages, previously-built, or known extras"
     echo "  --with-ref-packages <dir>          use the specified directory of available reference packages to determine what prebuilts to delete, instead of downloading the most recent version"
     echo "  --with-packages <dir>              use the specified directory of available previously-built packages to determine what prebuilts to delete, instead of downloading the most recent version"
+    echo "  --with-sdk                         use the specified SDK to check out source code.  do not copy it to the tarball.  an external SDK will be required to build from the tarball."
```
dseefeld: Should this be "--with-sdk <dir>" ?

Issue #1452, opened 2020-01-20
[3.0.102/3.1.101] various assemblies no longer r2r compiled since service release


tmds:
@omajid noticed file sizes changed for various assemblies with the servicing release:
```
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Antiforgery.dll: from 89600 to 43520 bytes (-51%)
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Authentication.Cookies.dll: from 87552 to 38400 bytes (-56%)
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Authentication.Core.dll: from 65536 to 31744 bytes (-51%)
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Authentication.OAuth.dll: from 76288 to 36864 bytes (-51%)
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Authentication.dll: from 128512 to 57344 bytes (-55%)
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Authorization.dll: from 68096 to 31232 bytes (-54%)
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Components.Forms.dll: from 43520 to 19968 bytes (-54%)
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Components.Server.dll: from 724992 to 391168 bytes (-46%)
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Components.Web.dll: from 94720 to 49664 bytes (-47%)
  /usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/3.0.{0|2}/Microsoft.AspNetCore.Components.dll: from 355840 to 135680 bytes (-61%)
[...]

PR #1458, opened 2020-01-24
[release/3.0] Fix extensions nupkg, enable aspnet crossgen

dagood:
See https://github.com/dotnet/source-build/issues/1452

> Now I see that 3.0.0 has a patch called [`Fix packing on *nix systems`](https://github.com/dotnet/source-build/blob/v3.0.0-runtime/patches/aspnet-extensions/0002-Fix-packing-on-nix-systems.patch) that was deleted in https://github.com/dotnet/source-build/pull/1260/commits/afdf3d6a69a860f807f4e37251821e89fe4ca5e8. 😕 I'm running a build now with it restored on source-build@v3.0.1-runtime to see if that fixes the `lib` issue and lets crossgen run.

It worked for me in a basic build.


dagood:
I think we should wait for https://github.com/dotnet/source-build/pull/1446 to make sure this doesn't interfere with public tagging of 3.0.2.

dagood:
Closing--this is now incorporated into https://github.com/dotnet/source-build/pull/1446! They conflict, so it would create pointless extra effort to merge here then re-uptake in that PR.

PR #1464, opened 2020-01-28
Update 3.1 to Jan2019 release.

crummel:


dagood:
I've added a commit that ports the change to enable ASP.NET Core crossgen from https://github.com/dotnet/source-build/pull/1446 to this branch. It turns out the patch is necessary in 3.0, but not 3.1, so it would *appear* we can simply re-enable crossgen in the 3.1 branch. Helps fix https://github.com/dotnet/source-build/issues/1452 everywhere, not just 3.0.

PR #1525, opened 2020-03-02
Update Version.Details.xml and supporting tasks for coherency and package name audit.

crummel:
- Correct some package names to match the same ones upstream uses.
- Arcade now exposes its dependency on XLiff - added the coherency parent.
- Core-SDK does not use standard directly - make it coherent with core-setup instead.
- Allow multiple RepoNames per Dependency to fix Maestro issue with multiple entries with the same package name.


tools-local/prebuilt-baseline-online.xml
```diff
@@ -104,6 +104,7 @@
     <Usage Id="Microsoft.Dotnet.WinForms.ProjectTemplates" Version="4.8.1-servicing.19605.5" />
     <Usage Id="Microsoft.DotNet.Wpf.ProjectTemplates" Version="3.0.0" />
     <Usage Id="Microsoft.DotNet.Wpf.ProjectTemplates" Version="3.1.2-servicing.20066.4" />
+    <Usage Id="Microsoft.Extensions.CommandLineUtils" Version="1.0.1" IsDirectDependency="true" />
```
dagood: Does this mean that we shouldn't remove `common`? (Going by the commit desc "Common has been merged to Extensions", it sounds like Extensions isn't correctly producing this package yet?)

crummel: It's used online but not offline- there was no offline baseline change, and strangely I didn't have to add it to the `additional-prebuilts-to-delete.txt` either.  I'll look into this a bit more but just based on that it seems safe.

eng/Version.Details.xml
```diff
@@ -31,37 +31,25 @@
...
       <Sha>e946cebe43a510e8c6476bbc8185d1445df33a1a</Sha>
       <RepoName>corefx</RepoName>
-    </Dependency>
-    <Dependency Name="Microsoft.Private.CoreFx.NETCoreApp" Version="4.7.0-servicing.20067.2" CoherentParentDependency="Microsoft.NETCore.App">
-      <Uri>https://github.com/dotnet/corefx</Uri>
-      <Sha>e946cebe43a510e8c6476bbc8185d1445df33a1a</Sha>
       <RepoName>corefx-portable</RepoName>
     </Dependency>
-    <Dependency Name="Microsoft.NETCore.App" Version="3.1.2-servicing.20067.4" CoherentParentDependency="Microsoft.Dotnet.Sdk.Internal">
+    <Dependency Name="Microsoft.NETCore.App.Internal" Version="3.1.2-servicing.20067.4" CoherentParentDependency="Microsoft.Dotnet.Sdk.Internal">
```
dagood: I would recommend `Microsoft.NETCore.App.Runtime.linux-x64` instead. The internal package isn't *likely* to be removed in 3.1, but it shouldn't be used by anyone directly. Using the runtime pack is also in line with `Microsoft.AspNetCore.App.Runtime.linux-x64` below (the scenario is the same).

crummel:
Bootstrap issue is unrelated - see https://github.com/dotnet/source-build/issues/1553.  Merging.

Issue #1472, opened 2020-02-05
The source-build infra binlog is overwritten by package-source-build/pack.sh


dagood:
Example:
https://dev.azure.com/dnceng/internal/_build/results?buildId=491337&view=results
https://dev.azure.com/dnceng/public/_build/results?buildId=495856&view=results

Artifact `Logs centos71 Production` => `source-build` => `logs` => `artifacts` => `log` => `Release` => `Build.binlog` seems to be from:

https://github.com/dotnet/source-build/blob/3085962b9d1c17a6c0b746837b94238257966d0a/src/package-source-build/pack.sh#L19

I would expect to see the main binlog here, showing details about project orchestration etc.

Simple solution: have `pack.sh` specify a project-specific binlog location.

However, treating package-source-build as a repository and actually calling a build command seems pretty strange, vs. handing it in process through MSBuild. From the code, I don't see a reason it helps for it to call the build script again.

dagood:
This hit me in a `v3.1.103-2-SDK` `linux-arm64` investigation after I already built my prod and tarball builds. I think this is a high priority especially as we have more people/places using the source-build infra.

dagood:
I think I'll grab this as part of https://github.com/dotnet/source-build/issues/1566.

Issue #1566, opened 2020-04-21
Improve infra compatibility with Arcade in preparation to migrate to dotnet/installer


dagood:
In [Experiment with adding source-build infrastructure to a repo with Arcade #1466](https://github.com/dotnet/source-build/issues/1466) I found a few things that I think we should clean up in source-build to make the move into dotnet/installer and dotnet/arcade smoother.

* stuff is running without the Arcade SDK, so some properties aren't available
* artifacts are mostly in `bin`, not `artifacts`, so it's not trivial to redirect [to a location that doesn't conflict with the dev's current "prebuilt build" intermediates/outputs]
* there is a *lot* of old code
* some info is set up in the build scripts rather than in MSBuild logic.

My plan of attack is:

1. Make these improvements in source-build `release/3.1`.
   * Technically completes this issue.
2. *After* dotnet/runtime is source-building, merge the improvements into `master`.
3. Move the infra to dotnet/installer.
   * We may find it prudent to wait to get all of 5.0 building in `master` before migrating. That's not actually necessary though.
4. Retire source-build's `master` branch. 🥂

dagood:
Improvements are done in `release/3.1`, and merged into `master`.

PR #1573, opened 2020-04-28
Use Arcade SDK conventions more consistently, fix 'package-source-build' project

dagood:
Completes a few items from https://github.com/dotnet/source-build/issues/1566 (some remain):

* stuff is running without the Arcade SDK, so some properties aren't available
* artifacts are mostly in `bin`, not `artifacts`, so it's not trivial to redirect [to a location that doesn't conflict with the dev's current "prebuilt build" intermediates/outputs]

I adjusted where the source-build resolver was installed: since we're using the Arcade SDK a little more in the tarball build, we need to override it earlier.

Also fixes https://github.com/dotnet/source-build/issues/1472 for build diagnosability (avoid overwriting primary binlog).

Tested locally with CentOS container production + tarball (online) + smoke-test.


dagood:
Bootstrap smoke-test error is existing issue: https://github.com/dotnet/source-build/issues/1553.

PR #1574, opened 2020-04-29
Remove src/package-source-build, replace with target: fix binlog overwrite

dagood:
Fixes https://github.com/dotnet/source-build/issues/1472

Splitting from https://github.com/dotnet/source-build/pull/1573: https://github.com/dotnet/source-build/pull/1573 significantly increases build time, but since successful PR validation builds don't produce useful binlogs, I couldn't easily compare against another CI build to see why build time regressed. Getting this change in will let me compare CI build vs. CI build. (Until then, I'm trying to repro locally.)


dagood:
Huh... this PR validation run has even worse perf on the production build step (> 1h50m vs. ~1h20m of https://github.com/dotnet/source-build/pull/1573, vs ~50m I was seeing in earlier runs I saw by browsing AzDO). I bet something's up with the agents. I'm going to run this locally and rely on that measurement.

dagood:
On my machine, this change ran in 56min and https://github.com/dotnet/source-build/pull/1573 ran in 43min. I think I'm seeing noise, I'm just surprised there's this much of it. I'm not going to block these PRs on perf.

dagood:
Failure in bootstrap smoke-test is existing, tracked by https://github.com/dotnet/source-build/issues/1553.

PR #1575, opened 2020-04-30
Trim some unused infrastructure code

dagood:
For part of https://github.com/dotnet/source-build/issues/1566:

* there is a *lot* of old code

I overstated it there--I only got rid of a few things that are fairly obviously only for 2.1. A lot of stuff appeared unused when I was working on the prototype and wasn't in tune with the tarball build.


dagood:
Bootstrap smoke test failure is known issue: https://github.com/dotnet/source-build/issues/1553

Issue #1487, opened 2020-02-18
CoreFX tests sometimes time out during PR validation builds


dagood:
* https://dev.azure.com/dnceng/public/_build/results?buildId=522079&view=logs&j=935c00b8-5de8-597d-a783-42e7f079746c&t=a2eb024e-cab8-5a87-0588-93d2706bcaf2&l=266
* https://dev.azure.com/dnceng/public/_build/results?buildId=522713&view=logs&j=935c00b8-5de8-597d-a783-42e7f079746c&t=a2eb024e-cab8-5a87-0588-93d2706bcaf2&l=267

This seems to be flaky. /cc @crummel @adaggarwal

adaggarwal:
I have also observed the same in CI. However, I don't see that whenever I try it locally.

adaggarwal:
I saw this in one of the timeout cases

```
Resolving SDK 'Microsoft.NET.Sdk.IL'...

[SourceBuiltSdkResolver] Looking for SDK Microsoft.NET.Sdk.IL. Detected config(s) in env:
'ARCADE_CONFIGURATION' for 'Microsoft.DotNet.Build.Tasks.Configuration/1.0.0-beta.20113.5' at '/src/Tools/source-built/Microsoft.DotNet.Build.Tasks.Configuration/'
'IL' for 'Microsoft.NET.Sdk.IL/1.0.0-source-built' at '/src/Tools/source-built/Microsoft.NET.Sdk.IL/'
'ARCADE_COREFX_TESTING' for 'Microsoft.DotNet.CoreFxTesting/1.0.0-beta.20113.5' at '/src/Tools/source-built/Microsoft.DotNet.CoreFxTesting/'
'ARCADE' for 'Microsoft.DotNet.Arcade.Sdk/1.0.0-beta.20113.5' at '/src/Tools/source-built/Microsoft.DotNet.Arcade.Sdk/'

[SourceBuiltSdkResolver] Overriding Microsoft.NET.Sdk.IL with 'IL'

Error when reading the file: System.IO.EndOfStreamException: Unable to read beyond the end of the stream.
   at System.IO.__Error.EndOfFile()
   at System.IO.BinaryReader.ReadString()
   at Microsoft.Build.Logging.StructuredLogger.BuildEventArgsReader.ReadBuildEventArgsFields() in C:\MSBuildStructuredLog\src\StructuredLogger\BinaryLogger\BuildEventArgsReader.cs:line 628
[...]

PR #1583, opened 2020-05-12
Move most of the tarball build into MSBuild

dagood:
* Modify build-source-tarball.sh as an entry point into the build. Trim some now-unused code (now shared with rest of the build) and pass info into build.sh.
* Refactor Darc acquisition as a project, using this pattern for MSBuild concurrency handling. This also means it won't happen twice in one MSBuild call, but will happen if it hasn't already.

For https://github.com/dotnet/source-build/issues/1566, "some info is set up in the build scripts rather than in MSBuild logic."


dagood:
* Bootstrap smoke-test error is known issue: https://github.com/dotnet/source-build/issues/1553
* Bootstrap tarball copy warning (`bin` doesn't exist) was introduced by https://github.com/dotnet/source-build/pull/1573 but didn't notice it until now. Filed https://github.com/dotnet/source-build/issues/1584 to track.
* "Run unit tests" timeouts are known issue: https://github.com/dotnet/source-build/issues/1487

eng/SourceBuild.Tarball.targets
```diff
@@ -0,0 +1,375 @@
...
+      <TarballCopyFile
+        Include="$(ProjectDir)support\tarball\PackageVersions.props"
+        RelativeRoot="artifacts/obj/$(Platform)/Release/" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="AddTarballPackages">
+    <!-- Remove reference-packages from tarball prebuilts...-->
+    <ItemGroup Condition="'$(CustomRefPackagesDir)' != ''">
+      <ExcludePrebuiltPackageFile Include="$(CustomRefPackagesDir)**\*.nupkg" />
```
dseefeld, line 164: I don't understand where `ExcludePrebuiltPackageFile` is used.  I only see it here and on line 177, but I don't see where it's used anyplace else.

eng/SourceBuild.Tarball.targets
```diff
@@ -0,0 +1,375 @@
...
+      <Output TaskParameter="ConsoleOutput" ItemName="PreviouslySourceBuiltPackageFile" />
+    </Exec>
+
+    <!-- Add packages to the tarball. -->
+    <ItemGroup>
+      <!-- Set up prebuilt nupkg filenames from tarballs to exclude. -->
+      <SourceBuildReferencePackageFilename Include="@(SourceBuildReferencePackageFile -> '%(Filename)')" />
+      <PreviouslySourceBuiltPackageFilename Include="@(PreviouslySourceBuiltPackageFile -> '%(Filename)')" />
+
+      <!-- Remove source-built packages from tarball prebuilts... -->
```
dseefeld, line 194: This comment confused me.  It's not really removing these packages from the tarball prebuilts, it's just collecting the names so it can remove them later.  I think the remove comment should be later when they're removed.

eng/SourceBuild.Tarball.targets
dagood: Good catch, this name is leftover from a refactor. It's meant to be the same item list as the task output. Fixed.

eng/SourceBuild.Tarball.targets
dagood: These are just copied from the tarball script. Reworked to "*Find* ... packages *to exclude* from tarball prebuilts", since we don't really remove them anymore.

dagood:
Bootstrap smoke test failure is known issue: https://github.com/dotnet/source-build/issues/1553

Issue #1190, opened 2019-08-22
Source-build support for Windows is suspended.


dseefeld:
This has never worked beyond core-setup. 

- [ ] Update readme
- [ ] Remove Windows build scripts / legs

neeraj9:
The README or relevant documentation for building this repository should be updated to reflect this stand.

MichaelSimons:
[Triage] source-build for windows is tracked in https://github.com/dotnet/source-build/issues/2910.

PR #1588, opened 2020-05-20
Update Windows build instruction

martincostello:
The Windows instruction says to run `build.ps1`, but there's no such file.


dagood:
Hi @martincostello, Windows isn't currently supported in this repo: https://github.com/dotnet/source-build/issues/1190. You'll see that even build.cmd turns around and tries to call the non-existent powershell script:

https://github.com/dotnet/source-build/blob/71e0adabbcc3a25b1f5aeb77e12eded220621a0c/build.cmd#L2

Also, our main branch is `release/3.1`, if you'd like to help us update the readme there.

martincostello:
Fair enough - I only did a cursory read of the repo after seeing it in the 5.0 preview 4 release blog post. If it doesn't work at all for Windows, I'll just leave it.

dagood:
Yeah, wish we had some time to set aside to make sure our docs and contrib guidance was clear before that link. This repo is quite unusual compared to the ones that make up the main components. Sorry for the confusion.

PR #1589, opened 2020-05-20
Update to May 2.1.18 release.

crummel:


crummel:
CLI commit should be merged now.
/azp run

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks.LeakDetection/MarkAndCatalogPackages.cs
```diff
@@ -53,10 +53,17 @@ public class MarkAndCatalogPackages : Task
         [Required]
         public ITaskItem[] PackagesToMark { get; set; }
 
+        /// <summary>
+        /// Use this directory instead of the system temp directory for staging.
+        /// Intended for Linux systems with limited /tmp space, like Azure VMs.
+        /// This directory must exist.
```
dagood, line 59: Why not handle this, to avoid surprises?

tools-local/init-build.proj
```diff
@@ -130,8 +130,8 @@
       <PrebuiltPackages Include="$(PrebuiltPackagesPath)**/*.nupkg" />
       <PrebuiltSourceBuiltPackages Include="$(PrebuiltSourceBuiltPackagesPath)**/*.nupkg" />
     </ItemGroup>
-    <MarkAndCatalogPackages PackagesToMark="@(PrebuiltPackages)" CatalogOutputFilePath="$(PoisonReportDataFile)" MarkerFileName="$(PoisonMarkerFile)" />
-    <MarkAndCatalogPackages PackagesToMark="@(PrebuiltSourceBuiltPackages)" CatalogOutputFilePath="$(SourceBuiltPoisonReportDataFile)" MarkerFileName="$(SourceBuiltPoisonMarkerFile)" />
+    <MarkAndCatalogPackages PackagesToMark="@(PrebuiltPackages)" CatalogOutputFilePath="$(PoisonReportDataFile)" MarkerFileName="$(PoisonMarkerFile)" OverrideTempPath="$(LeakDetectionTempPath)" />
```
dagood, line 133: Where is `LeakDetectionTempPath` set? I'm not seeing it in the repo.

crummel: The VSO build machines actually have enough /tmp space without this.  I added it because the standard DevTestLab machines don't, so I had to set it to debug this issue.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks.LeakDetection/MarkAndCatalogPackages.cs
crummel: Good point, added handling.

tools-local/init-build.proj
dagood: Ah, I expected the opposite usage, only in CI. Makes sense. :+1:

PR #1624, opened 2020-06-15
Updating owner for vstest

AbhitejJohn:
Updating with the current owner of vstest.


AbhitejJohn:
Tagging @nohwnd on this PR as well.

PR #1628, opened 2020-06-17
Add self-contained and framework-dependent publish tests.

crummel:
Fixes https://github.com/dotnet/source-build/issues/1580.

### DIFF
--- a/smoke-test.sh
+++ b/smoke-test.sh
@@ -205,6 +205,17 @@ function doCommand() {
             $killCommand
             wait $!
             echo "    terminated with exit code $?" | tee -a "$logFile"
+        elif [ "$1" == "publish" ]; then
+            binlogPrefix=$(echo "$testingDir/${lang}_${proj}_${binlogOnlinePart}_${binlogHttpsPart}_" | sed 's/--/_/g ; s/ /_/g')
+            if [ "$projectOutput" == "true" ]; then
+                "${dotnetCmd}" publish --self-contained false /bl:"${binlogPrefix}publish-fx-dep.binlog" | tee -a "$logFile"
+                "${dotnetCmd}" publish --self-contained true -r $targetRid /bl:"${binlogPrefix}publish-self-contained-${targetRid}.binlog" | tee -a "$logFile"
+                "${dotnetCmd}" publish --self-contained true -r linux-x64 /bl:"${binlogPrefix}publish-self-contained-portable.binlog" | tee -a "$logFile"
+            else
+                "${dotnetCmd}" publish --self-contained false /bl:"${binlogPrefix}publish-fx-dep.binlog" >> "$logFile" 2>&1
+                "${dotnetCmd}" publish --self-contained true -r $targetRid /bl:"${binlogPrefix}publish-self-contained-${targetRid}.binlog" >> "$logFile" 2>&1
+                "${dotnetCmd}" publish --self-contained true -r linux-x64 /bl:"${binlogPrefix}publish-self-contained-portable.binlog" >> "$logFile" 2>&1
+            fi
         else
             if [ "$projectOutput" == "true" ]; then
                 "${dotnetCmd}" $1 /bl:"$binlog" | tee -a "$logFile"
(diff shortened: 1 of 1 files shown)


smoke-test.sh
```diff
@@ -236,18 +243,18 @@ function setupDevCerts() {
 function runAllTests() {
     # Run tests for each language and template
     if [ "$excludeNonWebTests" == "false" ]; then
-        doCommand C# console new restore build run
-        doCommand C# classlib new restore build
+        doCommand C# console new restore build run "publish --self-contained false" "publish --self-contained true -r linux-x64" "publish --self-contained true -r $targetRid"
+        doCommand C# classlib new restore build "publish --self-contained false" "publish --self-contained true -r linux-x64" "publish --self-contained true -r $targetRid"
```
dagood, line 247: These are already just aliases rather than the full set of commands we run per test case, so could we just call it `publish` here and trifurcate in a conditional branch in `doCommand`?

Should get rid of the need to specially handle spaces and dashes?

smoke-test.sh
```diff
@@ -168,7 +174,8 @@ function doCommand() {
             binlogHttpsPart="https"
         fi
 
-        binlog="$testingDir/${lang}_${proj}_${binlogOnlinePart}_${binlogHttpsPart}_$1.binlog"
+        # double-hyphen and space in the binlog path screw up MSBuild argument parsing
+        binlog=$(echo "$testingDir/${lang}_${proj}_${binlogOnlinePart}_${binlogHttpsPart}_$1.binlog" | sed 's/--/_/g ; s/ /_/g')
```
dagood, line 178: Are there any cases where this is needed?

smoke-test.sh
```diff
@@ -198,6 +205,17 @@ function doCommand() {
             $killCommand
             wait $!
             echo "    terminated with exit code $?" | tee -a "$logFile"
+        elif [ "$1" == "publish" ]; then
+            binlogPrefix=$(echo "$testingDir/${lang}_${proj}_${binlogOnlinePart}_${binlogHttpsPart}_" | sed 's/--/_/g ; s/ /_/g')
```
dagood, line 209: Instead of duplicating, change the earlier declaration to support this?

```sh
binlogPrefix="$testingDir/${lang}_${proj}_${binlogOnlinePart}_${binlogHttpsPart}_"
binlog="${binlogPrefix}${1}binlog.binlog"
```

smoke-test.sh
```diff
@@ -198,6 +205,17 @@ function doCommand() {
...
+            if [ "$projectOutput" == "true" ]; then
+                "${dotnetCmd}" publish --self-contained false /bl:"${binlogPrefix}publish-fx-dep.binlog" | tee -a "$logFile"
+                "${dotnetCmd}" publish --self-contained true -r $targetRid /bl:"${binlogPrefix}publish-self-contained-${targetRid}.binlog" | tee -a "$logFile"
+                "${dotnetCmd}" publish --self-contained true -r linux-x64 /bl:"${binlogPrefix}publish-self-contained-portable.binlog" | tee -a "$logFile"
+            else
+                "${dotnetCmd}" publish --self-contained false /bl:"${binlogPrefix}publish-fx-dep.binlog" >> "$logFile" 2>&1
+                "${dotnetCmd}" publish --self-contained true -r $targetRid /bl:"${binlogPrefix}publish-self-contained-${targetRid}.binlog" >> "$logFile" 2>&1
+                "${dotnetCmd}" publish --self-contained true -r linux-x64 /bl:"${binlogPrefix}publish-self-contained-portable.binlog" >> "$logFile" 2>&1
```
dagood, line 217: Define a quick function locally?

```sh
runPublishScenarios() {
  "${dotnetCmd}" publish --self-contained false /bl:"${binlogPrefix}publish-fx-dep.binlog"
  "${dotnetCmd}" publish --self-contained true -r $targetRid /bl:"${binlogPrefix}publish-self-contained-${targetRid}.binlog"
  "${dotnetCmd}" publish --self-contained true -r linux-x64 /bl:"${binlogPrefix}publish-self-contained-portable.binlog"
}
if [ "$projectOutput" == "true" ]; then
  runPublishScenarios | tee -a "$logFile"
else
  runPublishScenarios >> "$logFile" 2>&1
fi
```

smoke-test.sh
crummel: No, shouldn't be any more.  Removed.

smoke-test.sh
crummel: Thanks, changed as suggested.

smoke-test.sh
crummel: Cool, I did not know about that.  Changed as suggested.

Issue #1676, opened 2020-07-21
Automatically create merged previously-source-built tar.gz in CI builds


dagood:
The build should create `Private.SourceBuild.Artifacts.{version}.tar.gz` automatically. Currently it's created manually.

Then, we may be able to upload it automatically: https://github.com/dotnet/source-build/issues/1321.

Issue #1638, opened 2020-06-26
Fix tarball usage of PackageVersions.props


dseefeld:
When building the tarball, the PVP is populated from 2 sources: 1) the PVP from the source-built-packages archive and 2) the PVP that gets generated as we build repos.  A PVP file that combines these two gets copied to `artifacts/obj/x64/Release/PackageVersions.props` when the tarball is created.  

This should be improved in a couple of ways:

1. The PVP from the source-built-packages archive should get generated when the packages are unpacked, rather than having a PVP that needs to match in the archive.  This has caused issues in the past.  See https://github.com/dotnet/source-build/issues/1629 
2. The PVP file that combines the two should be redesigned and generated on build, not copied to the tarball.  This requires the tarball to be shipped with an existing `artifacts` directory which is a bit odd.

mthalman:
This issue targets the old standalone source-build tarball's `PackageVersions.props` handling, which no longer exists. Source-build moved into the VMR (dotnet/dotnet) starting in .NET 8.

PR #1679, opened 2020-07-22
Add prepare artifacts job; fixup AzDO fetch script

dagood:
Add a simple prepare-artifacts job to take over manual work preparing previously-source-built tarballs.

Fix up the AzDO fetch script: simplify and document for the scenario it's actually used for: downloading onto fresh machines for manual testing.

---

I worked on this to avoid manual steps for 3.1.106, and polished it a little more for this PR for next time.

Fixes https://github.com/dotnet/source-build/issues/1676

### DIFF
--- a/.vsts.pipelines/builds/matrix.yml
+++ b/.vsts.pipelines/builds/matrix.yml
@@ -72,3 +72,14 @@ jobs:
     scriptPrefix: ./
     scriptSuffix: .sh
     setupMac: true
+
+- template: ../jobs/prepare-artifacts.yml
+  parameters:
+    # Gather from all jobs that have offline in the matrix. This also defines dependencies.
+    gatherJobs:
+    - job: centos71
+    - job: centos8
+    - job: debian9
+      artifactName: Tarball debian9 Online
+    - job: fedora30
+    gatherPortableJob: centos71
--- a/.vsts.pipelines/builds/matrix.yml
+++ b/.vsts.pipelines/builds/matrix.yml
@@ -73,13 +73,15 @@ jobs:
     scriptSuffix: .sh
     setupMac: true
 
-- template: ../jobs/prepare-artifacts.yml
-  parameters:
-    # Gather from all jobs that have offline in the matrix. This also defines dependencies.
-    gatherJobs:
-    - job: centos71
-    - job: centos8
-    - job: debian9
-      artifactName: Tarball debian9 Online
-    - job: fedora30
-    gatherPortableJob: centos71
+# Prepare artifacts in all situations but public PR validation.
+- ${{ if not(and(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest'))) }}:
+  - template: ../jobs/prepare-artifacts.yml
+    parameters:
+      # Gather from all jobs that have offline in the matrix. This also defines dependencies.
+      gatherJobs:
+      - job: centos71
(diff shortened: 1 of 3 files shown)


.vsts.pipelines/builds/matrix.yml
```diff
@@ -72,3 +72,14 @@ jobs:
     scriptPrefix: ./
     scriptSuffix: .sh
     setupMac: true
+
+- template: ../jobs/prepare-artifacts.yml
```
dagood, line 76: Should probably condition this to not run on public PR validation actually. (In all other scenarios we may take the previously-source-tarball, but not this.)

.vsts.pipelines/jobs/prepare-artifacts.yml
```diff
@@ -0,0 +1,84 @@
...
+      set -xeuo pipefail
+      mkdir -p "$(allSourceBuiltStageDir)"
+
+      # Extract all source-built assets into a single place. Overlap and overwrites are expected.
+      # What matters is the portable ones are copied last and ultimately win.
+      find \
+        "$(nonportableSourceBuiltStageDir)" \
+        "$(portableSourceBuiltStageDir)" \
+        -iname 'Private.SourceBuilt.Artifacts.*.tar.gz' \
+        -exec tar -xf {} -C "$(allSourceBuiltStageDir)" \;
```
dseefeld, line 72: What about PackageVersions.props file?  If last one in wins, it will not reflect the combined files.  It seems like this needs to be coupled with https://github.com/dotnet/source-build/issues/1638

dagood: This is how @crummel made the previously-source-built tarball we're using now, and I don't see any PVPs, so I assume it worked out... I'm not sure.

crummel: Yeah, it's not strictly correct, but it works out because we don't use runtime package versions for anything right now.  If we're productizing it, it might be worth looking into a bit.  There's probably two things we care about:
- getting all package versions into PackageVersions.props: I think something like adding unique lines from each individual PackageVersions.props to the output PackageVersions.props would work decently well here and probably shouldn't screw up too much.  This currently doesn't matter but would cover us if we ever needed to depend on, e.g., a runtime package's version.
- not having any conflicting package versions: in theory, this shouldn't happen.  Package names and versions should be consistent across RIDs, and I don't think we've seen any real-life cases where this has gone wrong.  I can't think offhand of a cheap way to detect this other than a bunch of `sed` nonsense, but I think it would be worth filing an issue for a future improvement.

I don't think either of these are blockers and #1638 should obviate them anyway.

crummel:
Do you have the tarball from before you disabled for pull requests?  I checked for the build but couldn't find it, I was just going to make sure it looked reasonable.

dagood:
I don't (I don't think the build completed), but this is the output for 3.1.106:
https://dotnetcli.blob.core.windows.net/source-built-artifacts/assets/Private.SourceBuilt.Artifacts.0.1.0-3.1.106.tar.gz
from:
https://dev.azure.com/dnceng/internal/_build/results?buildId=719842&view=results

Thanks for checking.

dagood:
osx job failure (libicu not found in smoke-test) tracked by https://github.com/dotnet/source-build/issues/1685, seems impossible to be related to this PR's changes. Ignoring for merge.

PR #1700, opened 2020-08-10
Add incremental build

dseefeld:
Add the following items to accomplish incremental builds in source-build:

* Add defined semaphore files for each target that participates
  * Semaphores in Directory.Build.props apply to top-level targets and projects, i.e. Directory.Build.targets, Build.proj, tools-local/init-build.proj, etc.
  * Semaphores in repos/Directory.Build.props apply to all repo level targets and projects.
* For each participating target, add Inputs (generally something larger and more static, like the project file) and assign the semaphore file as the output.
* For each participating target, create the semaphore file when the target completes.
* Add Timestamps to build binlogs to ensure that older binlogs don't get overwritten when build is restarted.

Accomplishes item #6 in https://github.com/dotnet/source-build/issues/928 with the exception of clearing semaphore files for downstream repos when a repo starts building.


Directory.Build.targets
```diff
@@ -1,4 +1,8 @@
-<Project>
+<Project InitialTargets="SetupSemaphorePath">
+
+  <Target Name="SetupSemaphorePath">
+    <MakeDir Directories="$(CompletedSemaphorePath)" />
```
dagood, line 4: `<WriteLinesToFile File="a/b/c/d.txt" />` seems to write the dirs fine, is it possible to switch to that and remove the pre-create targets?

Directory.Build.props
```diff
@@ -202,10 +202,28 @@
...
+    <FetchSourcesSuccessSemaphore>$(CompletedSemaphorePath)FetchSources.complete</FetchSourcesSuccessSemaphore>
+    <WriteSourceLinkMetadataSuccessSemaphore>$(CompletedSemaphorePath)WriteSourceLinkMetadata.complete</WriteSourceLinkMetadataSuccessSemaphore>
+    <CopyAndDisassembleReferenceOnlyPackagesSuccessSemaphore>$(CompletedSemaphorePath)CopyAndDisassembleReferenceOnlyPackages.complete</CopyAndDisassembleReferenceOnlyPackagesSuccessSemaphore>
+    <DownloadSourceBuildReferencePackagesSuccessSemaphore>$(CompletedSemaphorePath)DownloadSourceBuildReferencePackages.complete</DownloadSourceBuildReferencePackagesSuccessSemaphore>
+    <DownloadSourceBuiltArtifactsSuccessSemaphore>$(CompletedSemaphorePath)DownloadSourceBuiltArtifacts.complete</DownloadSourceBuiltArtifactsSuccessSemaphore>
+    <ReportPoisonUsageSuccessSemaphore>$(CompletedSemaphorePath)ReportPoisonUsage.complete</ReportPoisonUsageSuccessSemaphore>
+    <GeneratePrebuiltBurndownDataSuccessSemaphore>$(CompletedSemaphorePath)GeneratePrebuiltBurndownData.complete</GeneratePrebuiltBurndownDataSuccessSemaphore>
+  </PropertyGroup>
```
dagood, line 226: I'm not convinced about the maintainability of having props for these... it seems like this is basically:

```xml
<PropertyGroup>
  <GeneratePrebuiltBurndownDataSuccessSemaphore>$(CompletedSemaphorePath)GeneratePrebuiltBurndownData.complete</GeneratePrebuiltBurndownDataSuccessSemaphore>
</PropertyGroup>
<Target Name="GeneratePrebuiltBurndownData"
        Inputs="$(MSBuildProjectFullPath)" 
        Outputs="$(GeneratePrebuiltBurndownDataSuccessSemaphore)">
  ...
  <Touch Files="$(GeneratePrebuiltBurndownDataSuccessSemaphore)" AlwaysCreate="true" />
</Target>
```

vs.

```xml
<Target Name="GeneratePrebuiltBurndownData"
        Inputs="$(MSBuildProjectFullPath)" 
        Outputs="$(CompletedSemaphorePath)GeneratePrebuiltBurndownData.complete">
  ...
  <Touch Files="$(CompletedSemaphorePath)GeneratePrebuiltBurndownData.complete" AlwaysCreate="true" />
</Target>
```

I prefer the second by quite a bit:
* I can double click the target name and paste it into the paths.
  * It's hard to rename a target with the non-delimited names currently.
* The info isn't spread out across files.
  * In theory, you can put a property group just above the target like I did above, to keep them together but still use the current approach. This would be non-standard MSBuild though.

Are there benefits to the current way I'm not seeing?

build.sh
```diff
@@ -123,5 +123,5 @@ scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
 if [ "$alternateTarget" == "true" ]; then
   "$CLIPATH/dotnet" $SDKPATH/MSBuild.dll "$scriptroot/build.proj" /bl:source-build-test.binlog /flp:v=diag /clp:v=m "$@"
 else
-  "$scriptroot/eng/common/build.sh" --restore --build -c Release --warnaserror false -bl /flp:v=diag "$@"
+  "$scriptroot/eng/common/build.sh" --restore --build -c Release --warnaserror false /bl:Build_$(date +"%m%d%H%M%S").binlog /flp:v=diag "$@"
```
dagood, line 126: I think we should move this to the standard `artifacts/log/Debug` location. Having one binlog in the repo root is not great (in theory source dirs should not be modified) but having them build up makes it worse--harder to navigate the repo since it's harder to see the "real" files.

Good idea to preserve them though!

Directory.Build.targets
dagood: Ah, that doesn't change the modify date if the file already exists. This does though:

```
<WriteLinesToFile File="a/b/c/d.txt" Overwrite="true" />
```

dseefeld: Thanks.  I'll change this.

Directory.Build.props
dseefeld: My thinking for adding them all in props was to make sure that there weren't duplicate names of semaphore files, because this could cause hard to diagnose problems.  Thinking more about it, I like your suggestion and paring it down to less code to implement.  grep can ensure that duplicates don't crop up.

build.sh
dseefeld: Agreed.

Directory.Build.props
dagood: That's a good point, I think that the extra maintenance is probably more risky than that though, and it's somewhat mitigated by MSBuild target names typically having to be unique anyway (or worse things will happen).

Could consider writing a build task that parses our MSBuild files and looks for unique and matching values to go even beyond. (I think the MSBuild libraries have a parser we could use. Seemed nice when I saw it but completely forget the context.)

repos/Directory.Build.targets
```diff
@@ -78,11 +82,15 @@
     <Exec Command="$(PatchCommand) %(PatchesToApply.Identity)"
           WorkingDirectory="$(ProjectDirectory)"
           Condition="'@(PatchesToApply)' != ''" />
+
+    <WriteLinesToFile File="$(RepoCompletedSemaphorePath)ApplyPatches.complete" Lines="complete" Overwrite="true" />
```
dagood, line 86: Why `Lines="complete"` rather than an empty file?

dagood:
> Accomplishes item #6 in #928

It doesn't have the clearing part of it though, right?

> When a repo starts building, all downstream repo semaphores should be removed so the new build's dependencies flow through to the end without question. This will help when making changes to e.g. CoreFX and expecting them show up in the Runtime without fully cleaning the repo first.

repos/Directory.Build.targets
dseefeld: I was under the impression that Lines was required.  I see now that it's not.  I'll update.

dseefeld:
> > Accomplishes item #6 in #928
> 
> It doesn't have the clearing part of it though, right?
> 
> > When a repo starts building, all downstream repo semaphores should be removed so the new build's dependencies flow through to the end without question. This will help when making changes to e.g. CoreFX and expecting them show up in the Runtime without fully cleaning the repo first.

Right. I missed that part.  I'll update.

dseefeld:
OSX issue is https://github.com/dotnet/source-build/issues/1685

PR #1727, opened 2020-09-02
Update to 5.0 Preview8

crummel:


repos/aspnetcore.proj
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <SourceDirectory>AspNetCore</SourceDirectory>
+    <SourceDirectory>aspnetcore</SourceDirectory>
```
dagood, line 4: Same as Version.Details.xml, remove?

repos/roslyn.proj
```diff
@@ -10,11 +10,14 @@
...
     <!-- Versioning args. -->
     <BuildCommandArgs>$(BuildCommandArgs) /p:OfficialBuild=true</BuildCommandArgs>
-    <BuildCommandArgs>$(BuildCommandArgs) /p:EnableNgenOptimization=false</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:ApplyPartialNgenOptimization=false</BuildCommandArgs>
+    <BuildCommandArgs>$(BuildCommandArgs) /p:EnablePartialNgenOptimization=false</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) /p:PublishWindowsPdb=false</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) /p:DotNetPackageVersionPropsPath=$(PackageVersionPropsPath)</BuildCommandArgs>
     <BuildCommandArgs>$(BuildCommandArgs) /p:DotNetBuildFromSource=true</BuildCommandArgs>
-    <BuildCommandArgs>$(BuildCommandArgs) /p:UsingToolMicrosoftNetCompilers=false</BuildCommandArgs>
+    <!-- UstingToolMSNetCompilers=false introduces a prebuilt so we'll have to turn back on after our bootstrap SDK has native integer support -->
```
dagood, line 18:
```suggestion
    <!-- UsingToolMicrosoftNetCompilers=false introduces a prebuilt so we'll have to turn back on after our bootstrap SDK has native integer support -->
```

repos/runtime.common.props
```diff
@@ -72,6 +74,8 @@
 
     <CommonBuildArguments>$(FlagParameterPrefix)arch $(Platform)</CommonBuildArguments>
     <CommonBuildArguments>$(CommonBuildArguments) $(FlagParameterPrefix)configuration $(Configuration)</CommonBuildArguments>
+    <CommonBuildArguments>$(CommonBuildArguments) /p:RuntimeConfiguration=$(Configuration)</CommonBuildArguments>
+    <CommonBuildArguments>$(CommonBuildArguments) /p:BypassFrameworkInstallChecks=true</CommonBuildArguments>
```
dagood, line 78: What's `BypassFrameworkInstallChecks` about?

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat/PublishCoreSetupBinaries.cs
```diff
@@ -28,7 +29,21 @@ public class PublishCoreSetupBinaries : Task
         //   dotnet-runtime-rhel.7-x64.2.0.0-preview2-25401-9.tar.gz
         //   dotnet-runtime-2.0.0-preview2-25401-9-rhel.7-x64.tar.gz
         // the "semver" capture would be 2.0.0-preview2-25401-9 in this case.
-        protected virtual string VersionMatchRegex => @"(\.|-)(?'semver'[0-9]+\.[0-9]+\.[0-9]+(-[A-Za-z0-9]+(-|\.)[0-9]+(-|\.)[0-9]+((-|\.)[0-9]+)?)?)";
+        protected virtual string VersionMatchRegex => @"(\.|-)(?'semver'[0-9]+\.[0-9]+\.[0-9]+(([-.])[A-Za-z0-9]+)*)";
+
+        // The regular expression is even more of a mess when we try to account for every RID and suffix that may exist.
+        // This is a list of bad stuff we should remove that's never part of a version number.  If adding to it, it
+        // should include the delimiter immediately beofre the RID, arch, or extension.
```
dagood, line 36:
```suggestion
        // should include the delimiter immediately before the RID, arch, or extension.
```

patches/arcade/0012-Disable-ILRewrite-step-this-isn-t-built-in-source-bu.patch
```diff
@@ -0,0 +1,26 @@
+From 893bccaed963504f49bd186f80f70b43d748a2ef Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Sun, 16 Aug 2020 18:29:04 -0500
+Subject: [PATCH 12/12] Disable ILRewrite step - this isn't built in
+ source-build so it fails.
```
dagood, line 5: Any idea what ILRewrite does, is our SDK defective? Issue tracking?

patches/runtime/0010-Remove-duplicate-property.patch
```diff
@@ -0,0 +1,31 @@
+From af031b51511a210b4a496107a1229ae3a4598c46 Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Tue, 1 Sep 2020 15:44:55 -0500
+Subject: [PATCH 10/10] Remove duplicate property
```
dagood, line 4: Not clear what the duplicate property is here, what problem this causes. Somewhere else in the file?

patches/runtime/0013-testing-tfm-fix.patch
```diff
@@ -0,0 +1,25 @@
+From 2cc5df80750de41544cd2a784ee1cf812e94295d Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Wed, 2 Sep 2020 08:24:20 -0500
+Subject: [PATCH 13/13] testing tfm fix
```
dagood, line 4: "testing"? Looks like a src project?

dagood (commented):
Migrating comments from AzDO:

repos/aspnetcore-tooling.proj
```diff
@@ -1,9 +1,4 @@
 <Project>
-
-  <PropertyGroup>
-    <SourceDirectory>AspNetCore-Tooling</SourceDirectory>
```
dagood, line 4: Remove this project entirely? It no longer contributes to the .NET SDK in 5.0, it's just VS and VS Code stuff. https://github.com/dotnet/versions/pull/636#issuecomment-684083154 (and I saw this elsewhere too, e.g. it's not in the arcade-powered source-build plan).

patches/aspnetcore/0001-Use-non-portable-NETCoreAppRuntime-for-crossgen.patch
```diff
@@ -0,0 +1,50 @@
...
+
+diff --git a/eng/Dependencies.props b/eng/Dependencies.props
+index 8fc6f8364f..8107003c09 100644
+--- a/eng/Dependencies.props
++++ b/eng/Dependencies.props
+@@ -93,6 +93,7 @@ and are generated based on the last package release.
+     <LatestPackageReference Include="Microsoft.NETCore.App.Runtime.linux-arm64" />
+     <LatestPackageReference Include="Microsoft.NETCore.App.Runtime.linux-musl-x64" />
+     <LatestPackageReference Include="Microsoft.NETCore.App.Runtime.linux-musl-arm64" />
++    <LatestPackageReference Condition="'$(TargetOsName)' != 'osx'" Include="microsoft.netcore.app.runtime.$(SourceBuildRuntimeIdentifier)" Version="$(MicrosoftNETCoreAppRuntimeVersion)" />
```
dagood, line 19: Package name case is different, and custom Version seems unnecessary, should be assigned below:

```xml
    <LatestPackageReference Update="@(LatestPackageReference->WithMetadataValue('Version', ''))">
      <Version
          Condition=" $([System.String]::new('%(Identity)').StartsWith('Microsoft.NETCore.App.Runtime.')) ">$(MicrosoftNETCoreAppRuntimeVersion)</Version>
```

Probably won't hurt, but would be nice to have this mergeable. On that note, could also slap a `'$(DotNetBuildFromSource)' == 'true'` on it.

repos/aspnetcore.proj
crummel: Removed.

repos/roslyn.proj
crummel: Changed as suggested.

repos/runtime.common.props
crummel: When compiling for a full framework, the SDK's targets file includes a check to make sure the framework is installed - this is never true for us because Linux doesn't have these and we use the ReferenceAssemblies packages to compile for full-fx instead.  That being said, I think I just removed the full fx compile so this probably isn't needed right now, I'll double-check.

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat/PublishCoreSetupBinaries.cs
crummel: Thanks, fixed.

patches/arcade/0012-Disable-ILRewrite-step-this-isn-t-built-in-source-bu.patch
crummel: Added an issue https://github.com/dotnet/source-build/issues/1728.  Note that this is just preserving the status quo - we've never built ILRewrite as far as I can tell but previously this target didn't get called.  Now that it was being called, we were failing without having those tasks available.

repos/aspnetcore-tooling.proj
crummel: Removed.

patches/runtime/0013-testing-tfm-fix.patch
crummel: Fixed the patch name and added comment.

patches/runtime/0010-Remove-duplicate-property.patch
crummel: Added a comment in the patch.

repos/runtime.common.props
crummel: Yup, removing the full fx TFM obsoleted this.  Removed.

dseefeld (approved):

patches/runtime/0017-Skip-some-test-projects.patch
```diff
@@ -0,0 +1,37 @@
...
+
+diff --git a/eng/Subsets.props b/eng/Subsets.props
+index fa72f142f0c..17103c8c611 100644
+--- a/eng/Subsets.props
++++ b/eng/Subsets.props
+@@ -59,7 +59,7 @@
+     <DefaultMonoSubsets Condition="'$(MonoEnableLLVM)' == 'true' and '$(MonoLLVMDir)' == ''">mono.llvm+</DefaultMonoSubsets>
+     <DefaultMonoSubsets>$(DefaultMonoSubsets)mono.runtime+mono.corelib</DefaultMonoSubsets>
+ 
+-    <DefaultLibrariesSubsets>libs.depprojs+libs.native+libs.ref+libs.src+libs.pretest+libs.packages</DefaultLibrariesSubsets>
```
dseefeld, line 18: How does this impact the source-build build?  Doesn't source-build specify it's own subsets in the patched int sb-specific build script?  Are the defaults used?

crummel: No, for libraries the default (`libs`) is used (see https://github.com/dotnet/source-build/blob/092d4d5b76a1365add295de3630d5e27ee13f027/patches/runtime/0003-Add-source-build-specific-build-script.patch#L99).

crummel:
Merging with CI failures; OSX timeout and other transient network issues.

PR #1734, opened 2020-09-09
[release/3.1] Fix macOS PAT flow for internal non-authenticated builds

dagood:
I wanted to take previously source-built artifacts from https://dev.azure.com/dnceng/internal/_build/results?buildId=807228&view=results, but the macOS leg failed due to auth issues. The auth json looks like this, literally, and gets a 401:

```json
...
{
  "endpoint":"https://pkgs.dev.azure.com/dnceng/_packaging/dotnet3.1-internal-transport/nuget/v3/index.json",
  "username":"optional",
  "password":"$(dn-bot-dnceng-artifact-feeds-rw)"
}
...
```

This is because `dn-bot-dnceng-artifact-feeds-rw` is not defined in that particular build definition. AzDO leaves the entire variable substitution expression when it doesn't match any varaible, then we pass it in directly:

```yaml
    InternalPackageFeedPatEnv: $(dn-bot-dnceng-artifact-feeds-rw)
...
    - bash: ...
      env:
        internalPackageFeedPat: $(InternalPackageFeedPatEnv)
```

On Linux, we actually inject this value into a shell script rather than using the task `env` setting. This means that bash is evaluating it, so the `$()` syntax means it's a subshell. The subshell fails to find a command `dn-bot-dnceng-artifact-feeds-rw` and gives no output. This assigns an empty string value to the PAT variable, which disables auth:

```
-e internalPackageFeedPat=$(dn-bot-dnceng-artifact-feeds-rw)
```

This is convoluted and probably only worked by coincidence in the first place, but it's trivial to change the macOS side to match the Linux side, and that's a very localized fix, so I figure it's reasonable to just do that. I added a comment to the yaml to explain what's going on here for the future.

I have a build running with this change here: https://dev.azure.com/dnceng/internal/_build/results?buildId=807484&view=results

### DIFF
--- a/.vsts.pipelines/jobs/ci-local.yml
+++ b/.vsts.pipelines/jobs/ci-local.yml
@@ -36,9 +36,13 @@ jobs:
     type: Production
 
     ${{ if not(and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'))) }}:
-      InternalPackageFeedPatEnv: ''
+      SetInternalPackageFeedPatBashCommand: ''
     ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
-      InternalPackageFeedPatEnv: $(dn-bot-dnceng-artifact-feeds-rw)
+      # In the internal build, we have one definition that provides dn-bot-dnceng-artifact-feeds-rw,
+      # and another that doesn't (to be more like a public build). For the definition that doesn't
+      # define dn-bot-dnceng-artifact-feeds-rw, this line relies on the subshell to fail when it
+      # tries to find a command by that name and set internalPackageFeedPat to nothing.
+      SetInternalPackageFeedPatBashCommand: 'export internalPackageFeedPat=$(dn-bot-dnceng-artifact-feeds-rw)'
 
   steps:
   - checkout: self
@@ -68,9 +72,11 @@ jobs:
   - template: ../steps/check-space-powershell.yml
 
   # Build source-build.
-  - script: ${{ format('{0}build{1} $(args.build)', parameters.scriptPrefix, parameters.scriptSuffix) }}
-    env:
-      internalPackageFeedPat: $(InternalPackageFeedPatEnv)
+  - script: |
+      set -x
+      $(SetInternalPackageFeedPatBashCommand)
+      set -e
+      ${{ format('{0}build{1} $(args.build)', parameters.scriptPrefix, parameters.scriptSuffix) }}
     displayName: Build source-build
     timeoutInMinutes: 150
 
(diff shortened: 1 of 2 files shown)


dagood:
Some flakiness in tooling setup. Hasn't impacted any of the macOS jobs, both are green now.

Mock official build CentOS 7 Offline:

```
+ /src/eng/common/build.sh ...
Downloading 'https://dot.net/v1/dotnet-install.sh'
curl: (35) Network file descriptor is not connected
```

Mock official build Fedora 30 Online:

```
+ /src/eng/common/build.sh ...
Downloading 'https://dot.net/v1/dotnet-install.sh'
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to dot.net:443
```

PR validation, Fedora 30 Offline Portable:

```
+ /src/eng/common/build.sh ...
Downloading 'https://dot.net/v1/dotnet-install.sh'
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to dot.net:443
```

Similar to the errors in https://github.com/dotnet/runtime/issues/34015, but this is occurring even before `dotnet-install.sh` is downloaded. In https://github.com/dotnet/runtime/issues/34015, retries seem like the only solution, so maybe we need them at the Arcade script level here.

.vsts.pipelines/jobs/ci-local.yml
```diff
@@ -68,19 +72,23 @@ jobs:
   - template: ../steps/check-space-powershell.yml
 
   # Build source-build.
-  - script: ${{ format('{0}build{1} $(args.build)', parameters.scriptPrefix, parameters.scriptSuffix) }}
-    env:
-      internalPackageFeedPat: $(InternalPackageFeedPatEnv)
+  - script: |
+      set -x
+      $(SetInternalPackageFeedPatBashCommand)
```
crummel, line 77: Does `set -e` cause a failure for an empty command?  Otherwise could do our standard `exuo pipefail` here.

dseefeld (approved):

crummel (approved):
LGTM, just one small note on the bash syntax.

.vsts.pipelines/jobs/ci-local.yml
dagood: `$()` exits 0, but we don't actually have an empty command here--we basically have `x=$(asdf)`. `asdf` can't be found, so the subshell command exits nonzero.

dagood:
Merging past CI: the affected jobs (just macOS) already completed successfully. I want to get started on an internal non-authenticated build to get the previously source-built artifacts to finish off the 3.1.108 release, where this change will be running in a build that goes to completion, so I'll catch any potential ultra-strange problems there.

dagood:
> I want to get started on an internal non-authenticated build to get the previously source-built artifacts to finish off the 3.1.108 release.

I looked again, and it turns out that the macOS job isn't used to produce previously source-built artifacts, so this was unnecessary. Nice to have the build running green for the future though, I suppose. 😄

PR #1741, opened 2020-09-11
Fix a couple CI issues leftover from preview8

crummel:
Portable build: Patch runtime to not include extra package icon in nupkgs.
OSX: Patch installer to not generate extra packages that require the aspnetcore-targeting-pack.tar.gz that aren't produced in source-build.  I checked and neither 3.1 nor 5.0 preview4 produced these, so this is a new requirement and the OSX build passes smoke-tests locally without it.

### DIFF
--- a/patches/installer/0005-Don-t-do-this-in-source-build-aspnetcore-targeting-p.patch
+++ b/patches/installer/0005-Don-t-do-this-in-source-build-aspnetcore-targeting-p.patch
@@ -1,0 +1,35 @@
+From 5f34f12928157f5ed1c351dc217bf4bf5b62befe Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Mon, 14 Sep 2020 09:47:04 -0500
+Subject: [PATCH 5/5] Don't do this in source-build -
+ aspnetcore-targeting-pack.tar.gz not available.
+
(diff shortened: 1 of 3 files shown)


patches/installer/0005-Don-t-do-this-in-source-build-aspnetcore-targeting-p.patch
```diff
@@ -0,0 +1,35 @@
+From 5f34f12928157f5ed1c351dc217bf4bf5b62befe Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Mon, 14 Sep 2020 09:47:04 -0500
+Subject: [PATCH 5/5] Don't do this in source-build -
+ aspnetcore-targeting-pack.tar.gz not available.
```
dagood, line 5: Is there an issue tracking this or a discussion about what this is/why source-build doesn't need it? Should loop in dotnet/installer folks if haven't been already.

crummel: Logged https://github.com/dotnet/source-build/issues/1747 and I'll loop in installer.

dagood (approved):

dseefeld (approved):

PR #1750, opened 2020-09-14
[5.0] Fix runtime pack RID list: omit source-built RID

dagood:
Fixes https://github.com/dotnet/source-build/issues/1746

Running a local build now to verify manually.


dagood:
I confirmed locally that this fixes it: `dotnet restore -r centos.7-x64` in my CentOS 7 build downloads the `linux-x64` runtime packs.

dseefeld (approved):

PR #1766, opened 2020-09-24
Patch Roslyn and installer to remove Humanizer and dotnet-deb-tool prebuilts.

crummel:


patches/roslyn/0007-Don-t-use-Humanizer-in-source-build.patch
```diff
@@ -0,0 +1,92 @@
+From f44e01f25b4d45a6a66a739f8ba7b89d7c52d63e Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Thu, 24 Sep 2020 13:45:50 -0500
+Subject: [PATCH 7/7] Don't use Humanizer in source-build.
```
dagood, line 4: Do you know what the package does? Are there any inhuman results on the outputs? 😄

patches/roslyn/0007-Don-t-use-Humanizer-in-source-build.patch
```diff
@@ -0,0 +1,92 @@
...
+   </ItemGroup>
+   <ItemGroup>
+     <InternalsVisibleTo Include="Microsoft.CodeAnalysis.CSharp.EditorFeatures" />
+@@ -60,4 +60,4 @@
+   </ItemGroup>
+   <Import Project="..\..\..\Workspaces\SharedUtilitiesAndExtensions\Compiler\CSharp\CSharpCompilerExtensions.projitems" Label="Shared" />
+   <Import Project="..\..\..\Workspaces\SharedUtilitiesAndExtensions\Workspace\CSharp\CSharpWorkspaceExtensions.projitems" Label="Shared" />
+-</Project>
+\ No newline at end of file
++</Project>
```
dagood, line 89: I think we should probably avoid doing this to keep the diff as minimal as possible. (Avoid conflicts + keep diff small when applying to repos in arcade-powered source-build.)

dagood (approved):

patches/roslyn/0007-Don-t-use-Humanizer-in-source-build.patch
crummel: It appears to be used to help make friendly names of things, for instance expressions in the debugger.  Some of these will end up being grammatically incorrect (plurals) or harder to read (CamelCaseWithNoSpaces), but there should be no real functional impact.

patches/roslyn/0007-Don-t-use-Humanizer-in-source-build.patch
crummel: Fixed.

patches/roslyn/0007-Don-t-use-Humanizer-in-source-build.patch
dagood: Interesting... it sounds like this might be important for new users in particular though--reduce the number of weird things they need to understand to investigate issues they're seeing. I think we should track this as a product experience bug/diff. Looks like it's in .NET Foundation at least, to add it at some point: https://github.com/Humanizr/Humanizer

crummel: Logged https://github.com/dotnet/source-build/issues/1767 to track it.

dseefeld (approved):

PR #1773, opened 2020-09-25
SourceLink patch debugging

crummel:
I can't actually repro either of these issue locally so seeing what breaks in CI.


crummel:
Figured this out and https://github.com/dotnet/sourcelink/pull/653 is the resulting better patch.

PR #1774, opened 2020-09-28
More complete release scripts, fix checklist ordering and reference to scripts

dagood:
The step ordering previously suggested downloading + renaming the tarballs even though they won't be used yet. This slows down the process because it takes a while, and people could be doing validation. This is just a small bug: the intent was clear before, but just a little wrong if followed step by step.

I also added some scripts to the repo that should help with the tarball munging process. I moved the inner dir renaming function over to this file from a personal repo for easier usage here. These scripts are still unpolished and just enough to get the release together.


dseefeld (approved):

dagood:
Going ahead with merge: these files aren't touched by the build.

PR #1784, opened 2020-10-01
[release/3.1] Use one stage per job in CI

dagood:
Use one stage per job. This is a workaround that allows us to retry a job at any time, even if other jobs are still running. This means that if a particular job fails early (e.g. macOS due to a temporary homebrew dependency install failure), we can restart it immediately, rather than having to wait until the entire build completes *then* queueing a fresh macOS build to also wait on for a while longer.

This means no more built-in `matrix` job strategy. That was only able to split a job into several jobs within a single stage. Instead, use a template with `${{ each ... }}` to create multiple stages.

This also makes the dependencies between builds visualized now. It's odd--job dependencies don't show up anywhere in the AzDO UI at all, but stage dependencies get an attempt at a dependency graph. (Our stage graph works out cleanly here, but when you get intersecting edges, the UI makes them ambiguous.)

Split into two commits to make `prepare-artifacts.yml` more reviewable--there are only slight changes to that template. (Also enable "no whitespace" mode.)


dseefeld (approved):

crummel (approved):

dagood:
I appreciate AzDO keeping my life interesting by giving me puzzles to solve every day, like "Why is my build still running after leaving it overnight--oh, it's a second attempt? Why is that running?" (More seriously, thanks to someone for presumably hitting retry.)

---

Noticed this potential flakiness in the first attempt of `centos71_Offline_Portable`:

```
LOCAL RESTORE SOURCE - ALL TESTS PASSED!
curl: (60) Peer's Certificate has expired.
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.

##[debug]Exit code 60 received from tool '/bin/bash'
##[debug]STDIO streams have closed for tool '/bin/bash'
##[error]Bash exited with code '60'.
```

https://dev.azure.com/dnceng/public/_build/results?buildId=839463&view=logs&j=8c35c666-1cc6-56eb-620e-0f6e7d3caf0a&s=149ab9f9-2746-51c3-9bed-64d13cb4f094&t=b30ab823-f7fe-562e-90b7-7ff3e51be1be&l=1420

Filed https://github.com/dotnet/source-build/issues/1788 for that, assuming it's flakiness.

crummel:
Sorry, yup, that was me, it looked like flakiness this morning so I kicked it.

PR #1785, opened 2020-10-01
Download only Private.SourceBuilt artifacts

dseefeld:
Update prepare-artifacts.yml to only update Private.SourceBuilt.Artifacts.*.tar.gz rather than all artifacts in the tarball folder.  The job is failing because the agent is running out of disk space.


dagood:
Have an internal testing build running on this? (Since the job doesn't run in public PR.)

dagood (approved):

crummel (approved):

dagood:
Seems stale (no longer necessary?) and the PR branch is on the official repo rather than a fork--closing for now to clean up.

PR #1782, opened 2020-09-30
Update to RC1

dseefeld:

PR #1790, opened 2020-10-05
[master] Use one stage per job in CI

dagood:
(cherry picked from commit f08ac8a5aaffd3936f5b7c64e925118d9190dfc3)

```
# Conflicts:
#	.vsts.pipelines/builds/matrix.yml
#	.vsts.pipelines/jobs/prepare-artifacts.yml
```


dagood:
macOS first failure seems to be:

>    Please install openssl before running this script, see https://github.com/dotnet/runtime/blob/master/docs/workflow/requirements/macos-requirements.md

This is already broken on the branch, and will be resolved by the macOS patch in https://github.com/dotnet/source-build/pull/1782. I think because this PR only changes some high-level wrappers in CI, we can go ahead with the merge regardless. (Then we can start using this feature in https://github.com/dotnet/source-build/pull/1782 after a re-merge.)

crummel (approved):

PR #1797, opened 2020-10-12
Patch installer to not restore WPF and WinForms templates.

crummel:
These are already not included in non-Windows builds but are still restored.

### DIFF
--- a/patches/installer/0007-Don-t-restore-WinForms-and-WPF-templates-for-source-.patch
+++ b/patches/installer/0007-Don-t-restore-WinForms-and-WPF-templates-for-source-.patch
@@ -1,0 +1,74 @@
+From b8b4aecabbd046eff325395899410ba427ab8ede Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Mon, 12 Oct 2020 14:19:49 -0500
+Subject: [PATCH 7/7] Don't restore WinForms and WPF templates for
+ source-build.
+
+---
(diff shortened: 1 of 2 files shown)


patches/installer/0007-Don-t-restore-WinForms-and-WPF-templates-for-source-.patch
```diff
@@ -0,0 +1,74 @@
...
+   </ItemGroup>
+ 
+   <Target Name="LayoutTemplates"
+@@ -173,10 +155,6 @@
+ 
+   <Target Name="LayoutTemplatesForSDK"
+           DependsOnTargets="SetupBundledComponents;CalculateTemplatesVersions">
+-    <ItemGroup Condition="!$(ProductMonikerRid.StartsWith('win'))">
+-      <Bundled50Templates Remove="Microsoft.Dotnet.Wpf.ProjectTemplates" />
+-      <Bundled50Templates Remove="Microsoft.Dotnet.WinForms.ProjectTemplates" />
```
dagood, line 67: To make this patch more mergeable, we could have a similar condition on the packagedownloads. To save time now, this seems fine though.

dagood (approved):

dseefeld (approved):

patches/installer/0007-Don-t-restore-WinForms-and-WPF-templates-for-source-.patch
crummel: Logged https://github.com/dotnet/source-build/issues/1802 for this.

PR #1805, opened 2020-10-16
Update ASP.NET to not exclude templates in source-build

crummel:


crummel:
This was rolled into another PR.

PR #1806, opened 2020-10-16
[release/3.1] Fix prepare-artifacts portable dependency

dagood:
Prepare Artifacts downloads artifacts from a portable job, but doesn't have a stage dependency on it. The timing's been working out by chance so far but we should fix this.


crummel (approved):

dagood:
Since this has no effect on PR validation, I'm going to merge since it has approval and to see how it works in the rolling build.

dagood:
:+1:

image

PR #1808, opened 2020-10-19
runtime: System.Private.CoreLib: disable analyzers

tmds:
should eliminate StyleCop.Analyzers, StyleCop.Analyzers.Unstable

@crummel ptal

cc @omajid


dagood (approved):

tmds:
It also eliminates `"Microsoft.CodeAnalysis.NetAnalyzers" Version="3.3.0-beta1.20355.1"`.

crummel (approved):

PR #1815, opened 2020-10-20
[master] Try to diagnose macOS failures during brew update

dagood:


dagood:
That workaround seems to have worked on the "Install native dependencies", where it counts. (I had the build intentionally fail later to avoid wasting time.) Incorporating that workaround in https://github.com/dotnet/source-build/pull/1817.

PR #1817, opened 2020-10-20
[master] Add macOS brew image/env workaround

dagood:
Based on https://github.com/actions/virtual-environments/issues/1811.

I'm also porting this to `release/3.1`.

### DIFF
--- a/.vsts.pipelines/steps/setup-macos-native.yml
+++ b/.vsts.pipelines/steps/setup-macos-native.yml
@@ -8,6 +8,11 @@ steps:
       )
     }
     set -ex
+    # Work around image/environment bug: https://github.com/actions/virtual-environments/issues/1811
+    brew uninstall openssl@1.0.2t
+    brew uninstall python@2.7.17
+    brew untap local/openssl
+    brew untap local/python2
     # Work around brew bug: https://github.com/dotnet/coreclr/pull/21913
     brew update
     # Install macOS native dependencies not present in hosted pool.


dagood:
Aaand there are general AzDO degradations showing up now. https://status.dev.azure.com

.vsts.pipelines/steps/setup-macos-native.yml
```diff
@@ -8,6 +8,11 @@ steps:
       )
     }
     set -ex
+    # Work around image/environment bug: https://github.com/actions/virtual-environments/issues/1811
+    brew uninstall openssl@1.0.2t
+    brew uninstall python@2.7.17
+    brew untap local/openssl
```
dagood, line 14: Well, I guess this is to be expected:

```
  [21:17:11.11] Building 'runtime'
...
    Commencing CoreCLR Repo build
    __DistroRid: osx-x64
    __RuntimeId: osx-x64
    Setting up directories for build
    Checking prerequisites...
    /usr/local/bin/pkg-config
    Please install openssl before running this script, see https://github.com/dotnet/runtime/blob/master/docs/workflow/requirements/macos-requirements.md
  /Users/runner/work/1/s/artifacts/src/runtime.bf456654f9a4f9a86c15d9d50095ff29cde5f0a4/src/coreclr/runtime.proj(33,5):
    error MSB3073: The command ""/Users/runner/work/1/s/artifacts/src/runtime.bf456654f9a4f9a86c15d9d50095ff29cde5f0a4/src/coreclr/build-runtime.sh" -x64 -release -ci -keepnativesymbols -os OSX /p:OfficialBuildId=20200807.11 -nopgooptimize" exited with code 1.
```

Once it's practical to work on this (vs. AzDO general failures) we've got to figure out where we actually want to get openssl from. Maybe after the `brew update` we can install it again? I haven't looked very far into why exactly we need to uninstall it in the first place. (My impression is that the macOS image comes with a busted `brew` installation that we have to un-break before we can run `brew update`, but this might not be accurate.)

Issue #1816, opened 2020-10-20
Build R2RDump from source


dagood:
Context: R2RDump's code is in dotnet/runtime for 5.0. It brings in a prebuilt dependency from a repo that isn't in source-build.

Currently R2RDump is not included in the Runtime and SDK so we plan to skip building it from source: https://github.com/dotnet/runtime/issues/43599. The project's dependencies don't look trivial to source-build. (Native code, doesn't use Arcade and dependency flow as of writing, code maintenance in flux between dotnet/llilc and dotnet/jitutils. More details in issue.)

But should R2RDump be something that we *do* provide as part of the source-built .NET ecosystem?

I'm not too familiar with it, but it [sounds like](https://github.com/dotnet/runtime/tree/dcc49112aa4b03a2347409fb95e837767e2ad49f/src/coreclr/src/tools/r2rdump) it might be interesting:

> Parses and outputs the contents of a ReadyToRun image

This issue tracks inclusion of R2RDump in source-build. If it's something we will never care about, then we can close, but I think it's worth tracking as it's something that exists in dotnet/runtime but we won't be source-building for now.

PR #1820, opened 2020-10-20
[master] Exclude R2RDump from source-build

dagood:
This should remove the `Microsoft.NETCore.CoreDisTools/1.0.1-prerelease-00005` prebuilt dependency.

Context at https://github.com/dotnet/source-build/issues/1816 for why we can exclude R2RDump.


dagood:
Confirmed in local build that it's removed from production + tarball build.

```xml
<BaselineComparison>
  <Removed>
...
    <PackageIdentity Id="Microsoft.NETCore.CoreDisTools" Version="1.0.1-prerelease-00005" />
...
    <PackageIdentity Id="runtime.placeholder-rid.Microsoft.NETCore.CoreDisTools" Version="1.0.1-prerelease-00005" />
...
  </Removed>
</BaselineComparison>
```

Forgot to make the change to the baselines to verify this in PR validation, but I'm going to avoid making that change here so I don't reset CI. (Unless I also need to make some other change.)

dseefeld (approved):

dseefeld:
@dagood Can this be closed in favor of #1822?

dagood:
Yep, this package shows up as removed in the prebuilt data for https://github.com/dotnet/source-build/pull/1822.

Issue #1819, opened 2020-10-20
macOS CI failure: "brew update" doesn't work in CI image; workaround causes openssl to be missing


dagood:
These occur in the `Install native dependencies` step in our macOS CI:

```
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
```

```
Error: Fetching /usr/local/Homebrew/Library/Taps/local/homebrew-openssl failed!
Fetching /usr/local/Homebrew/Library/Taps/local/homebrew-python2 failed!
fatal: invalid upstream 'origin/master'
fatal: invalid upstream 'origin/master'
```

```
##[error]Bash exited with code '1'.
```

This is caused by a macOS CI image bug: https://github.com/actions/virtual-environments/issues/1811

---

There is a workaround, but applying it causes the build to be unable to find `openssl`. More info in https://github.com/dotnet/source-build/pull/1817 (master) and https://github.com/dotnet/source-build/pull/1818 (release/3.1).

dagood:
The CI at https://github.com/dotnet/source-build/pull/1925 failed later into the build than brew configuration. Maybe we can look at turning this back on.

PR #1827, opened 2020-10-21
Disable macOS CI temporarily

dagood:
This makes it easier to tell when PR validation succeeds.

(It also removes the risk of someone doing a bad PR merge if a real error is mistaken for the known macOS failure, but this isn't IMO very significant for our small team vs. a repo like dotnet/runtime where it matters much more.)

Fixing is tracked by https://github.com/dotnet/source-build/issues/1819


dagood:
@dseefeld @crummel ok to merge without full CI? macOS isn't involved in the gathering/bootstrap legs and this removal is very targeted. It would be nice to get this in ASAP to remove clutter from other PRs.

crummel (approved):
Yes, that sounds fine to me.

dseefeld (approved):

PR #1843, opened 2020-10-28
Update runtime nuget version to SBRP version

dseefeld:
Update RefOnlyNuget*Versions to use 5.1.0 rather than 4.9.4, since 5.1.0 already exists as a reference package in source-build-reference-packages


dagood (approved):
IIRC these dotnet/runtime NuGet dependencies are only used to build the local build tasks, so I think this will work out without complicating things downstream, which is nice.

Issue #1825, opened 2020-10-21
Investigate ExcludeFromSourceBuild not working in dotnet/runtime


dagood:
https://github.com/dotnet/source-build/pull/1820 tried to exclude R2RDump from source-build, but it still got restored and its dependencies were still prebuilts.

I suspect that the new static restore might not be evaluating the projects the same way as they used to, preventing the `ExcludeFromSourceBuild` property from working. I haven't looked very far into it, but I wonder if maybe the problem is that it isn't checking `DotNetBuildFromSource` in the environment and we need to pass it in as a command, since `RestoreTaskEx` is called something like this (newlines for clarity):

```
"/src/.dotnet/dotnet"
"/src/.dotnet/sdk/5.0.100-preview.8.20417.9/NuGet.Build.Tasks.Console.dll"
"CleanupAssetsForUnsupportedProjects=True;
  NoCache=True;
  Recursive=True"
"/src/.dotnet/sdk/5.0.100-preview.8.20417.9/MSBuild.dll"
"/src/artifacts/src/runtime.bf456654f9a4f9a86c15d9d50095ff29cde5f0a4/Build.proj"
"ContinuousIntegrationBuild=true;
  Configuration=Release;
  RepoRoot=/src/artifacts/src/runtime.bf456654f9a4f9a86c15d9d50095ff29cde5f0a4/;
  Rebuild=false;
  Test=false;
  Pack=false;
  IntegrationTest=false;
  PerformanceTest=false;
  RuntimeConfiguration=Release;
  Subset=clr.tools+clr.runtime+clr.corelib+clr.nativecorelib+clr.packages;
  TargetArchitecture=x64;
  CMakeArgs=;
  MicrosoftNetFrameworkReferenceAssembliesVersion=1.0.0;
  PackagesDir=/src/packages/restored/;
  PackageRid=centos.7-x64;
  NoPgoOptimize=true;
  KeepNativeSymbols=true;
  RuntimeOS=centos.7;
  DotNetPackageVersionPropsPath=/src/artifacts/obj/x64/Release/PackageVersions.props;
  DotNetRestoreSourcePropsPath=/src/artifacts/obj/x64/Release/RestoreSources.props;
  VersionsPropsPath=/src/artifacts/src/runtime.bf456654f9a4f9a86c15d9d50095ff29cde5f0a4/eng/Versions.props;
  DisableWarnForInvalidRestoreProjects=true;
  __BuildPhase=SolutionRestore;
[...]

Issue #1763, opened 2020-09-23
Document why we upgrade/delete targetframeworks throughout repos


dagood:
As a repo owner, I'd expect this to be fine for source-build:

```xml
<TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
```

`net461` should be built using a .NET Framework targeting pack, and `netcoreapp2.0` using I think `Microsoft.NETCore.App/2.0.9` as a targeting pack. The result can then run on any later runtimes.

But, source-build tends to have patches that remove `net461` and upgrade `netcoreapp2.0` to a later version. These targetframework upgrades are somewhat viral: downstream repos have to react, and we do this with more source-build patches. In some cases, we even have source code patches to handle the API differences in new targetframeworks.

I think we need well-documented justification for this, since we're asking repo owners to take our patches for arcade-powered source-build. Or better, we should remove these changes.

/cc @dseefeld @crummel @dleeapho 

---

This also figures into *[Establish a policy for tooling repos with arcade-powered source-build: how does one build work for 5.0 and 6.0? #1743](https://github.com/dotnet/source-build/issues/1743)*. The 6.0 SDK may need to build the 5.0 repos, which it should be able to do. If we require SDK-version-specific patching (upgrading `net5.0` to `net6.0`), we need to account for how to do that in a plan.

tmds:
My 2 cents... may be too simplistic.

As a source-build user, I expect to build a .NET runtime, and SDK that works on that runtime.
Being able to set that runtime as a target framework, is a good way to build only necessary bits and eliminate prebuilts.

For cases like OmniSharp-vscode, I hope we may be getting to a place where mono can use the net5.0+ assemblies instead of requiring separate net4* targets.

cc @omajid

dagood:
Yeah, I agree it seems good as far as building the SDK and Runtime. I don't know any reason any of the DLLs in either one of those must be compatible with a lower framework version than the current Runtime version. (Other than for OmniSharp (https://github.com/dotnet/source-build/issues/1650).)

[...]

Issue #1627, opened 2020-06-16
UseStableVersions property should be eliminated


crummel:
We should be able to use checked-in metadata to determine whether to build a given repo as stabilized or unstabilized.  Previously when I tried this I ran into issues with [WriteSourceRepoProperties](https://github.com/dotnet/source-build/blob/a6c33a189e30588ebe0ad2c7e0b88545184f6bbd/tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat/WriteSourceRepoProperties.cs) not getting the stable/unstable metadata correctly.

dagood:
> We should be able to use checked-in metadata to determine

Since ProdCon v2/Arcade/Darc (3.0+), I think we shouldn't even need to determine it... the value is checked in, so using a default build command should create the right stabilization.

https://github.com/dotnet/corefx/blob/c68bc274ce95cb5eda143223942d6fd86a547a0d/eng/Versions.props#L14
https://github.com/dotnet/installer/blob/v3.1.201/src/redist/targets/Versions.targets#L11

In ProdCon v1, we had to know it because an orchestrator passed in a value, which could be changed without any source code changes. It was literally impossible to know based on the source code. 😛 I expect we'll always need to pass this in manually in `release/2.1`.

In ProdCon v2, there's no more orchestrator. As far as this kind of build param goes, it's deterministic.

dagood:
Triage:
@dseefeld ran into stabilization property issues in 5.0, and I think we should probably do this there. We don't think it's necessary in 3.1 because things should be stable there.

mthalman:
No longer relevant.

PR #1845, opened 2020-10-28
Update to 5.0 GA

dseefeld:


eng/Version.Details.xml
```diff
@@ -75,22 +75,22 @@
       <Uri>https://github.com/dotnet/xliff-tasks</Uri>
       <Sha>a52f3d7fb58470749ee4035fbbcb7e63c78b0459</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.AspNetCore.App.Ref" Version="5.0.0-rc.1.20451.17" CoherentParentDependency="Microsoft.Dotnet.Sdk.Internal">
+    <Dependency Name="Microsoft.AspNetCore.App.Ref" Version="5.0.0" CoherentParentDependency="Microsoft.Dotnet.Sdk.Internal">
```
dagood, line 78: This is the wrong package, should be `Microsoft.AspNetCore.App.Runtime.win-x64` to avoid missing servicing updates.

patches/installer/0008-Update-blob-versions-to-use-ref-pkg-versions-to-trac.patch
```diff
@@ -0,0 +1,28 @@
...
+--- a/src/redist/targets/GenerateLayout.targets
++++ b/src/redist/targets/GenerateLayout.targets
+@@ -11,8 +11,8 @@
+ 
+   <PropertyGroup>
+     <!-- Blob storage directories are not stabilized, so these must refer to a package that does not stabilize -->
+-    <AspNetCoreBlobVersion>$(VSRedistCommonAspNetCoreSharedFrameworkx6450PackageVersion)</AspNetCoreBlobVersion>
+-    <CoreSetupBlobVersion>$(MicrosoftNETCoreAppInternalPackageVersion)</CoreSetupBlobVersion>
++    <AspNetCoreBlobVersion>$(MicrosoftAspNetCoreAppRefPackageVersion)</AspNetCoreBlobVersion>
++    <CoreSetupBlobVersion>$(MicrosoftNETCoreAppRefPackageVersion)</CoreSetupBlobVersion>
```
dagood, line 22: This will break in servicing, the ref package doesn't update with every release but the runtime version will. I believe the infra we have putting them in a stable location is what's wrong here (or rather, not matching what the modern infra does anymore). The blob storage virtual folder should always be unstable and only the filename version gets stable.

repos/linker.proj
```diff
@@ -20,7 +20,7 @@
     <BuildPackagesCommandArgs>$(BuildPackagesCommandArgs) $(ProjectDirectory)src/ILLink.Tasks/ILLink.Tasks.csproj</BuildPackagesCommandArgs>
     <BuildPackagesCommand>$(DotNetCliToolDir)dotnet $(BuildPackagesCommandArgs)</BuildPackagesCommand>
 
-    <PackagesOutput>$(ProjectDirectory)artifacts/packages/illink_$(Configuration)/Shipping/</PackagesOutput>
+    <PackagesOutput>$(ProjectDirectory)artifacts/packages/illink_$(Configuration)/NonShipping/</PackagesOutput>
```
dagood, line 23: Let both through with this bit I just noticed in other projects?

```xml
  <Target Name="SetOutputList" AfterTargets="Package" BeforeTargets="GatherBuiltPackages">
    <ItemGroup>
      <PackagesOutputList Include="$(ShippingPackagesOutput)" />
      <PackagesOutputList Include="$(NonShippingPackagesOutput)" />
    </ItemGroup>
  </Target>
```

repos/runtime.proj
```diff
@@ -11,5 +11,12 @@
       <RepositoryReference Include="runtime-portable" Condition="'$(TargetOS)' == 'Linux' and '$(BuildPortableRuntime)' == 'true'" />
   </ItemGroup>
 
+    <!-- Additional Targets -->
+  <Target Name="InstallJustBuiltRuntime" AfterTargets="RemoveBuiltPackagesFromCache">
+    <!-- Install the runtime that was just built to be used by downstream repos, namely, aspnetcore -->
+    <Exec Command="tar -xvf $(SourceBuiltAssetsDir)dotnet-runtime-$(runtimeOutputPackageVersion)-$(TargetRid).tar.gz -C $(DOTNET_INSTALL_DIR)" />
```
dagood, line 17: What happens on a rebuild? What happens in bootstrapping when we build N using N? This seems pretty difficult to get right. Maybe it doesn't need to block merging this PR (so we can keep working on prebuilts in parallel ASAP) but I think we should find a less overwritey way immediately.

repos/linker.proj
dseefeld: I'm not sure adding both is worthwhile, because later in the project, we're installing the version we built, which needs to use `PackagesOutput` to locate the package.

eng/Version.Details.xml
dseefeld: Updated.

patches/installer/0008-Update-blob-versions-to-use-ref-pkg-versions-to-trac.patch
dseefeld: Filed issue https://github.com/dotnet/source-build/issues/1853

dseefeld:
@dagood @crummel  Any more comments on this?  I think it will be ready to merge as soon as it goes green.

repos/runtime.proj
dagood: Filed https://github.com/dotnet/source-build/issues/1854 for this.

dagood:
Filed an issue for my remaining comment from earlier. Going through it again now.

Directory.Build.props
```diff
@@ -20,7 +20,7 @@
     <Platform Condition="'$(Platform)' == '' AND '$(BuildArchitecture)' == 'arm64'">$(BuildArchitecture)</Platform>
     <Platform Condition="'$(Platform)' == ''">x64</Platform>
 
-    <UseStableVersions Condition="'$(UseStableVersions)' == ''">false</UseStableVersions>
+    <UseStableVersions Condition="'$(UseStableVersions)' == ''">true</UseStableVersions>
```
dagood, line 23: Can we delete this line? #1627

tools-local/prebuilt-baseline-offline.xml
```diff
@@ -78,90 +84,39 @@
...
+    <PackageIdentity Id="System.Text.Encodings.Web" Version="4.7.1" />
     <PackageIdentity Id="System.Xml.XPath.XmlDocument" Version="4.0.1" />
     <PackageIdentity Id="XliffTasks" Version="1.0.0-beta.20420.1" />
   </NeverRestoredTarballPrebuilts>
   <Usages>
-  <!-- portable-build only prebuilts -->
-    <Usage Id="runtime.any.System.Collections" Version="4.3.0" Rid="any" />
-    <Usage Id="runtime.any.System.Diagnostics.Tools" Version="4.3.0" Rid="any" />
-    <Usage Id="runtime.any.System.Diagnostics.Tracing" Version="4.3.0" Rid="any" />
-    <Usage Id="runtime.any.System.Globalization" Version="4.3.0" Rid="any" />
```
dagood, line 95: Not sure if it'll cause a problem to remove this manually maintained section, but pointing it out.

dagood (approved):

patches/aspnetcore/0005-Fix-version-number.patch
```diff
@@ -1,24 +0,0 @@
-From 366568840602e3d7b99e5a3076c24dd9d53dada3 Mon Sep 17 00:00:00 2001
-From: dseefeld <dseefeld@microsoft.com>
-Date: Wed, 30 Sep 2020 14:13:13 +0000
-Subject: [PATCH] Fix version number
-
-This file is actually a very long minified JS file with a version number at the
-end. The file is treated as binary, making this patch less clear, but it's easy
-to update/reapply it with a text editor.
----
- .../Web.JS/dist/Release/blazor.server.js           | Bin 245791 -> 245793 bytes
```
crummel, line 10: Is this just not needed anymore or did the version number move somewhere else?

crummel (approved):
LGTM, just one question about the ASP.NET version.

patches/aspnetcore/0005-Fix-version-number.patch
dseefeld: The version number was 5.0.0.

Directory.Build.props
dseefeld: No.  Without it, Roslyn wasn't building stable vesions.

tools-local/prebuilt-baseline-offline.xml
dseefeld: Augh.  Missed that one.  Since the portable build is almost done, I'll wait and see if any of the portable prebuilts change and update it if so.

Directory.Build.props
dagood: > Roslyn wasn't building stable versions.

Why is that a goal? The Microsoft Roslyn build didn't produce stable versions.
https://maestro-prod.westus2.cloudapp.azure.com/1297/https:%2F%2Fgithub.com%2Fdotnet%2Froslyn/68310/graph

> image

And in Version.Details.xml, `<Dependency Name="Microsoft.NETCore.Compilers" Version="3.8.0-5.20519.18" ...`

vs.
```
  [14:25:41.87] Building 'roslyn'...done
  New NuGet package(s) after building roslyn:
    -> Microsoft.CodeAnalysis.Build.Tasks 3.8.0
    -> Microsoft.CodeAnalysis.Common 3.8.0
```

dseefeld: Hmm...  A later repo was looking for a stable version and wasn't finding it.  I was mistaken in my fix for that.

dagood: I think it's fine to leave https://github.com/dotnet/source-build/issues/1627 tracking it for followup but wanted to dig into any issues to make it easier to tackle later.

patches/aspnetcore/0005-Fix-version-number.patch
dagood: I take it you mean that the version number already present in the file is 5.0.0, so we don't need to patch it to 5.0.0? Maybe we should open an issue to track adding this patch back in 5.0.1 when the version gets desynchronized again, so we're less likely to forget/rediscover it in the future?

dseefeld: Yes.  It didn't need to be updated.  I added https://github.com/dotnet/source-build/issues/1855

Directory.Build.props
dseefeld: Without the patch, the production build produces `Microsoft.NETCore.Compilers.3.8.0-dev.20519.18.nupkg`.  The "dev" version doesn't seem right.

dagood: That sounds like a missing `-ci` or `OfficialBuildId` or something along those lines.

dseefeld: I didn't change roslyn.proj.  It has `OfficialBuild=true` here: https://github.com/dotnet/source-build/blob/92d2aefc755e365d09f661fbaf9e0c3755f2179d/repos/roslyn.proj#L12

dseefeld: Adding `--ci` to roslyn.proj caused it to generate the desired version.  `Microsoft.Net.Compilers.Toolset.3.8.0-5.20519.18.nupkg`  I'll fix this once this PR is merged under #1627

patches/runtime/0026-Pin-System.Security.Cryptography.Cng-to-single-versi.patch
```diff
@@ -0,0 +1,30 @@
...
+From: dseefeld <dseefeld@microsoft.com>
+Date: Fri, 30 Oct 2020 16:25:21 +0000
+Subject: [PATCH] Pin System.Security.Cryptography.Cng to single version
+
+If System.Security.Cryptography.Cng is restored using the PVP
+version (5.0.0), this causes conflicts later in the restore which
+results in an error:
+   An item with the same key has already been added.
+   Key: (System.Security.Cryptography.Cng, 5.0.0)
+---
```
tmds, line 11: @dseefeld @dagood does this mean something doesn't allow us to consume `System.Security.Cryptography.Cng` `5.0.0` to produce `System.Security.Cryptography.Cng` `5.0.0`?

dseefeld: Yes, that's what I was seeing.  Here's the error I was seeing without this patch:
```
3000 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error : An item with the same key has already been added. Key: (System.Security.Cryptography.Cng, 5.0.0) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3001 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3002 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at System.Linq.Enumerable.ToDictionary[TSource,TKey](List`1 source, Func`2 keySelector, IEqualityComparer`1 comparer) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3003 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at NuGet.Commands.LockFileBuilder.CreateLockFile(LockFile previousLockFile, PackageSpec project, IEnumerable`1 targetGraphs, IReadOnlyList`1 localRepositories, RemoteWalkContext context) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3004 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at NuGet.Commands.RestoreCommand.ExecuteAsync(CancellationToken token) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3005 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at NuGet.Commands.RestoreRunner.ExecuteAsync(RestoreSummaryRequest summaryRequest, CancellationToken token) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3006 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at NuGet.Commands.RestoreRunner.ExecuteAndCommitAsync(RestoreSummaryRequest summaryRequest, CancellationToken token) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3007 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at NuGet.Commands.RestoreRunner.CompleteTaskAsync(List`1 restoreTasks) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3008 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at NuGet.Commands.RestoreRunner.RunAsync(IEnumerable`1 restoreRequests, RestoreArgs restoreContext, CancellationToken token) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3009 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at NuGet.Commands.RestoreRunner.RunAsync(RestoreArgs restoreContext, CancellationToken token) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3010 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at NuGet.Build.Tasks.BuildTasksUtility.RestoreAsync(DependencyGraphSpec dependencyGraphSpec, Boolean interactive, Boolean recursive, Boolean noCache, Boolean ignoreFailedSources, Boolean disableParallel, Boolean force, Boolean forceEvaluate, Boolean hideWarningsAndErrors, Boolean restorePC, Boolean cleanupAssetsForUnsupportedProjects, ILogger log, CancellationToken canc     ellationToken) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]
3011 /src/tb-5.0GA/.dotnet/sdk/5.0.100-rc.1.20452.10/NuGet.RestoreEx.targets(10,5): error :    at NuGet.Build.Tasks.Console.MSBuildStaticGraphRestore.RestoreAsync(String entryProjectFilePath, IDictionary`2 globalProperties, IReadOnlyDictionary`2 options) [/src/tb-5.0GA/src/runtime.cf258a14b70ad9069470a108f13765e0e5988f51/Build.proj]

```

tmds: Doesn't this happen for a lot of packages since we're consuming 5.0.0 prebuilt packages to build 5.0.0 source-built packages? Or is issue somehow specific to the Cng package?

dseefeld: I don't know why this issue popped up for this specific package.  We could certainly try to remove the patch and see if the issue repros.  It may have been a "point in time" issue.

dagood: Here's some more info from our old chat about it. dotnet/runtime seems to be doing something specific that aggravates some NuGet bug. It won't be a point in time issue, I doubt the NuGet issue will be fixed in our timeline. I was able to put together a simplified repro, but we dropped it because we probably don't actually need a fix here. (And we always want to cut things/work because time is not something we have much of.)

> dagood: [in response to a binlog] Yeah, good old RestoreTaskEx... another issue we have with static restore is https://github.com/dotnet/source-build/issues/1825 FWIW, but unrelated.
> 
> ...
>
> Seems like we might be running into https://github.com/NuGet/Home/issues/6795 ? But not sure why we would but the Microsoft build wouldn't...

> [10/30 9:51 AM] Davis Goodin
>     Hmm, noticed a few things earlier in those RestoreTaskEx logs:
> 
> ```
> Could not parse version  for LatestPackage System.Security.Cryptography.Cng, will use latest stable.
> ...
> Installing System.Security.Cryptography.Cng 5.0.0.
> ```
> 
> (And actually the lines do seem scrambled around that installing line, so not sure which project it's coming from.)

> ​[10/30 9:52 AM] Davis Goodin
>     I guess some other project is referring to the Cng package. Not totally sure that's important though.

> ​[10/30 9:55 AM] Davis Goodin
>     Hmm. Wonder if it's choking on this, where in the prod build we don't have a PVP value but in the tarball build we're overwriting it?
> https://github.com/dotnet/runtime/blob/7ef6d50b312217d2f7c17b9697891fa8ab98a19d/src/libraries/System.Security.Cryptography.Pkcs/src/System.Security.Cryptography.Pkcs.csproj#L639-L642
> 
> ```
>  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard2.0'))">
>  <PackageReference Include="System.Buffers" Version="$(SystemBuffersVersion)" />
>  <PackageReference Include="System.Memory" Version="$(SystemMemoryVersion)" />
>  <PackageReference Include="System.Security.Cryptography.Cng" Version="$(SystemSecurityCryptographyCngVersion)" />
> ```
> 

> ​[10/30 9:57 AM] Davis Goodin
>     For a sanity check... does the PVP being passed in in your local build have SystemSecurityCryptographyCngVersion defined?

> [10/30 9:58 AM] Dan Seefeldt
>     Yes.
> ​[10/30 9:59 AM] Dan Seefeldt
>     The difference between production and tarball is that production build has it defined as 4.7.0 while the tarball build is 5.0.0.

> [10/30 10:40 AM] Davis Goodin
>   Got it, looks like TargetFrameworks is part of the key for NuGet to mash them together and fail: https://github.com/dagood/repro/tree/repro-static-restore-ref-mashup
> 
> Oh huh. Static restore is not necessary to repro.
>
> (I was wondering if we could also fix it by disabling static restore.)
>
> After mulling over what Tom's suggesting (and Wes suggested in the past) I think the real long term fix is to stop building all TFMs except net5.0... (so that we don't have this PackageReference to begin with) [referring to #1763]

> dseefeld: BTW, forcing System.Security.Cryptography.Cng to 4.7.0 fixed the issue.

4.7.0 isn't a problem because we should be able to build it in SBRP, so we should be able to leave it at that.

This info probably should have made it to the patch, or an issue linked to from the patch. (But we're rushing.)

tmds: @dagood since you have a small reproducer, it seems worth filing a bug in https://github.com/NuGet/Home (even if it won't be fixed for our 5.0).

dagood: Agreed, but I don't feel I have time to make it make enough sense for discussion. Filed https://github.com/dotnet/source-build/issues/1888 to track it though.

dagood: FYI, I've filed the issue for this: https://github.com/NuGet/Home/issues/10368.

PR #1856, opened 2020-11-03
Fix Roslyn version

dseefeld:
As part of the 5.0GA update, the roslyn repo was building with a "-dev" version.  To fix this, the `UseStableVersions` propery was set to true, but this caused it to not build the version that ships with the Microsoft build.  Fix this issue by removing the `UseStableVersions` property and add `--ci` to the roslyn build to get it to build the correct version.


dagood:
For https://github.com/dotnet/source-build/issues/1627

dagood:
For what it's worth, IMO this should have a pretty low priority compared to prebuilt reduction because we haven't identified any product breaks happening due to this. (Not sure whether you're investigating the failure in CI.)

crummel:
@dseefeld to follow up.

dseefeld:
Closing.  This is not high priority at this point.

PR #1863, opened 2020-11-05
Dynamically inject prev-sb nuget into arcade to remove NuGet prebuilts

dagood:


dseefeld (approved):

repos/arcade.proj
```diff
@@ -51,5 +51,26 @@
...
+      Text="Unable to find any previously source built NuGet.Packaging version to use in Arcade SDK tooling."
+      Condition="'@(PreviouslySourceBuiltNuGetPackageFile)' == ''" />
+
+    <ReadNuGetPackageInfos PackagePaths="@(PreviouslySourceBuiltNuGetPackageFile)">
+      <Output TaskParameter="PackageInfoItems" ItemName="PreviouslySourceBuiltNuGetPackageInfo" />
+    </ReadNuGetPackageInfos>
+
+    <PropertyGroup>
+      <PreviouslySourceBuiltNuGetVersion>%(PreviouslySourceBuiltNuGetPackageInfo.PackageVersion)</PreviouslySourceBuiltNuGetVersion>
+      <BuildCommand>$(BuildCommand) /p:NuGetVersion=$(PreviouslySourceBuiltNuGetVersion)</BuildCommand>
```
dagood, line 71: I didn't consider the `PackageVersions.props` stored within the previously source-built dir when I wrote this--I don't think that makes this approach invalid (so we should still merge it for now), but I was running into issues with this in a future prebiult removal and will focus on sharing that infra properly in the future.

PR #1871, opened 2020-11-09
Use our Microsoft.Diagnostics.NETCore.Client, not prebuilt

dagood:
We build `Microsoft.Diagnostics.NETCore.Client/0.2.0` thanks to https://github.com/dotnet/source-build/pull/1824. Add a patch to dotnet/vstest to use the version from the PVP file rather than the hard-coded `0.2.0-preview.20419.2`.


dagood:
I think the reason this regressed (it worked initially: https://github.com/dotnet/source-build/pull/1824#issuecomment-716088622) is that the 5.0 GA upgrade PR enabled a flag that forces all repos to build a stable version. That bumped us off the hard-coded version.

dseefeld (approved):

PR #1875, opened 2020-11-12
vstest: don't include an apphost for vstest.console

tmds:
By targetting .NET 5 instead of .NET Core 2.1 an apphost is included by default.
This updates the patch to not include the apphost.

Fixes https://github.com/dotnet/source-build/issues/1874

cc @omajid @dagood


dagood (approved):

omajid (approved):
Thanks! I can confirm this patch fixes the issue for me:

```
$ tar tf ./artifacts/x64/Release/dotnet-sdk-5.0.100-fedora.32-x64.tar.gz | grep vstest.console
./sdk/5.0.100/vstest.console.dll.config
./sdk/5.0.100/vstest.console.runtimeconfig.json
./sdk/5.0.100/vstest.console.deps.json
./sdk/5.0.100/vstest.console.dll
```

PR #1904, opened 2020-11-20
runtime: Use substitute SHA-1 impl in wasm

dagood:
Port https://github.com/dotnet/runtime/pull/44982 from dotnet/runtime `master` to 5.0.0 GA as a source-build patch. (Applies cleanly.)

For https://github.com/dotnet/runtime/issues/44884.


omajid (approved):
I haven't read the code, but given that the change was verified in runtime and patch applies cleanly: :+1:

crummel (approved):

PR #1908, opened 2020-11-20
linker: Remove apphost prebuilt usage

dagood:
For https://github.com/dotnet/source-build/issues/1905.

Running a local build now to make sure I can see if it's working, not sure yet.


dagood:
> Running a local build now to make sure I can see if it's working

Actually, now that I know which `project.assets.json` files to look for, I can download them and look at them directly. No need for a local build--local files are just needed for initial discovery.

dagood:
Confirmed that `src/linker.e792d61a0ead2f6422b4c9af651c1128d978d454/artifacts/obj/analyzer` no longer exists, and `src/linker.e792d61a0ead2f6422b4c9af651c1128d978d454/artifacts/obj/Mono.Linker/project.assets.json` no longer uses the apphost package.

dseefeld (approved):

PR #1910, opened 2020-11-21
fsharp: Use net5.0 to avoid 3.1 apphost prebuilt

dagood:
`fsc` and `fsi` apphosts end up in the SDK, so I think they need to be `net5.0`.

`fslex`, `fsyacc`, and `AssemblyCheck` seem to be build-time tooling, so removing apphost generation might be an option. But we might as well move all these projects to `net5.0` consistently if possible.

For https://github.com/dotnet/source-build/issues/1905.

---

Adds dependency on `System.Configuration.ConfigurationManager` because it seems necessary now in my local build. I assume this is because of `net5.0`, but I'm not sure.

---

Also fixes a missing repo dependency that `fsharp` has on `msbuild`. It needs the new APIs available in new `msbuild` packages. This was preventing `/p:RootRepo=fsharp` from working:

```
.../src/fsharp/LegacyMSBuildReferenceResolver.fs(88,59): error FS0039: The type 'TargetDotNetFrameworkVersion' does not define the field, constructor or member 'Version472'. [.../src/fsharp/FSharp.Compiler.Private/FSharp.Compiler.Private.fsproj]
.../src/fsharp/LegacyMSBuildReferenceResolver.fs(89,58): error FS0039: The type 'TargetDotNetFrameworkVersion' does not define the field, constructor or member 'Version48'. [.../src/fsharp/FSharp.Compiler.Private/FSharp.Compiler.Private.fsproj]
.../src/fsharp/LegacyMSBuildReferenceResolver.fs(113,98): error FS0039: The type 'TargetDotNetFrameworkVersion' does not define the field, constructor or member 'Version48'. [.../src/fsharp/FSharp.Compiler.Private/FSharp.Compiler.Private.fsproj]
.../src/fsharp/LegacyMSBuildReferenceResolver.fs(114,100): error FS0039: The type 'TargetDotNetFrameworkVersion' does not define the field, constructor or member 'Version472'. [.../src/fsharp/FSharp.Compiler.Private/FSharp.Compiler.Private.fsproj]
.../src/fsharp/SimulatedMSBuildReferenceResolver.fs(69,59): error FS0039: The type 'TargetDotNetFrameworkVersion' does not define the field, constructor or member 'Version472'. [.../src/fsharp/FSharp.Compiler.Private/FSharp.Compiler.Private.fsproj]
.../src/fsharp/SimulatedMSBuildReferenceResolver.fs(70,58): error FS0039: The type 'TargetDotNetFrameworkVersion' does not define the field, constructor or member 'Version48'. [.../src/fsharp/FSharp.Compiler.Private/FSharp.Compiler.Private.fsproj]
```

### DIFF
--- a/patches/fsharp/0006-Use-net5.0-TFM-to-avoid-3.1-apphost-prebuilt.patch
+++ b/patches/fsharp/0006-Use-net5.0-TFM-to-avoid-3.1-apphost-prebuilt.patch
@@ -1,21 +1,37 @@
-From 793325d73d4a84e73bf78e7c0d0f8b6b22c8a66f Mon Sep 17 00:00:00 2001
+From c13763b3a2deba6cb48afd4e08157ae647d59a5a Mon Sep 17 00:00:00 2001
 From: Davis Goodin <dagood@microsoft.com>
-Date: Fri, 20 Nov 2020 18:28:36 -0600
+Date: Fri, 20 Nov 2020 19:13:35 -0600
 Subject: [PATCH] Use net5.0 TFM to avoid 3.1 apphost prebuilt
 
 See https://github.com/dotnet/source-build/issues/1905
 ---
- eng/Versions.props                                |  1 +
- eng/build.sh                                      | 12 +++++++-----
- proto.proj                                        |  6 +++---
- src/buildtools/AssemblyCheck/AssemblyCheck.fsproj |  2 +-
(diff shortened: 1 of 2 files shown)


dagood:
Looking good: the dependency disappeared from the `project.assets.json` files, but the output nupkg still has the apphosts in it, which should mean that it's coming from the 5.0 package in the SDK's `packs/` dir.

dseefeld (approved):

PR #1912, opened 2020-11-23
runtime: Keep Microsoft.Win32.SystemEvents pkg ref in System.Drawing.Common

dagood:
For https://github.com/dotnet/source-build/issues/1889.

This package ref is what brings Microsoft.Win32.SystemEvents into the aspnetcore shared framework. We need it, or else the source-built Runtime ends up missing the assembly.


dagood:
Didn't work, `System.Drawing.Common/5.0.0` is still missing its ref on `Microsoft.Win32.SystemEvents/5.0.0`.

PR #1927, opened 2020-11-26
Fix patch name

tmds:
@dagood changes patch prefix from 00015- to 0015-


dseefeld (approved):

dseefeld:
centos71_Offline succeeded on builds and smoke-tests, but failed uploading assets.  This is OK to merge.

PR #1933, opened 2020-11-30
Add fix from runtime for - ilasm doesn't preserve "where T:struct"...

dseefeld:
Fix is for https://github.com/dotnet/runtime/issues/45183


omajid (approved):
LGTM!

Issue #1943, opened 2020-12-08
dotnet/runtime portable build doesn't preserve coreclr binlogs


dagood:
The portable build is missing `Build.binlog`, and the portable build's `coreclrBuild.binlog` is actually the portable installer build binlog:

> ![image](https://user-images.githubusercontent.com/12819531/101419167-18213e80-38b5-11eb-9a26-624f8a0fc46a.png)

The last line of this patch should change `Build.binlog` to `installerBuild.binlog` and we wouldn't hit this problem: (when the non-portable next build comes along, it does the rename to the portable build's file, overwriting the old coreclr binlog!)

https://github.com/dotnet/source-build/blob/bd97d04be7f54d63c414429bade767ec3222a6af/patches/runtime/0003-Add-source-build-specific-build-script.patch#L92-L101

But applying this patch doesn't apply the last line. 😕

dagood:
Looks like the line count in the patch went bad somehow.

dagood:
Fixed by https://github.com/dotnet/source-build/pull/1944

Issue #1937, opened 2020-12-03
Portable build contains prebuilts


dseefeld:
The portable build for source-build contains prebuilts when the other builds don't.  Here are the prebuilts detected when building portable:

```
/src/eng/SourceBuild.Tarball.targets(364,5): error : Detected forbidden prebuilts. They must be removed, or explicitly allowed (see target for details): [/src/packages/restored/microsoft.dotnet.arcade.sdk/5.0.0-beta.20426.4/tools/Build.proj]
/src/eng/SourceBuild.Tarball.targets(364,5): error : /tb/tarball_907431/packages/prebuilt/runtime.any.system.collections.4.3.0.nupkg [/src/packages/restored/microsoft.dotnet.arcade.sdk/5.0.0-beta.20426.4/tools/Build.proj]
/src/eng/SourceBuild.Tarball.targets(364,5): error : /tb/tarball_907431/packages/prebuilt/runtime.any.system.diagnostics.tools.4.3.0.nupkg [/src/packages/restored/microsoft.dotnet.arcade.sdk/5.0.0-beta.20426.4/tools/Build.proj]
/src/eng/SourceBuild.Tarball.targets(364,5): error : /tb/tarball_907431/packages/prebuilt/runtime.any.system.diagnostics.tracing.4.3.0.nupkg [/src/packages/restored/microsoft.dotnet.arcade.sdk/5.0.0-beta.20426.4/tools/Build.proj]
/src/eng/SourceBuild.Tarball.targets(364,5): error : /tb/tarball_907431/packages/prebuilt/runtime.any.system.globalization.4.3.0.nupkg [/src/packages/restored/microsoft.dotnet.arcade.sdk/5.0.0-beta.20426.4/tools/Build.proj]
/src/eng/SourceBuild.Tarball.targets(364,5): error : /tb/tarball_907431/packages/prebuilt/runtime.any.system.globalization.calendars.4.3.0.nupkg [/src/packages/restored/microsoft.dotnet.arcade.sdk/5.0.0-beta.20426.4/tools/Build.proj]
/src/eng/SourceBuild.Tarball.targets(364,5): error : /tb/tarball_907431/packages/prebuilt/runtime.any.system.io.4.3.0.nupkg [/src/packages/restored/microsoft.dotnet.arcade.sdk/5.0.0-beta.20426.4/tools/Build.proj]
[...]

PR #1944, opened 2020-12-08
Fix coreclr binlog: corrupted patch

dagood:
Recreate the patch adding the dotnet/runtime `build-source-build.sh`: it wasn't adding the last line. This ends up overwriting the coreclr binlog in the portable build.

Fixes https://github.com/dotnet/source-build/issues/1943.

### DIFF
--- a/patches/runtime/0003-Add-source-build-specific-build-script.patch
+++ b/patches/runtime/0003-Add-source-build-specific-build-script.patch
@@ -1,19 +1,19 @@
 From acdac361d8c2e101f401c0e71a51d0018cfb76f9 Mon Sep 17 00:00:00 2001
 From: dseefeld <dseefeld@microsoft.com>
 Date: Wed, 15 Jan 2020 15:14:54 +0000
-Subject: [PATCH 03/10] Add source-build specific build script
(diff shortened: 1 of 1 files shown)


crummel (approved):
Is there some kind of Git issue with this patch?  I remember this happening before.

dagood:
It looks like with https://github.com/dotnet/source-build/commit/26c298f8acd841a8cc9e495c1fdcf6c437541018#diff-c72afe4ad8f6530dfdc0d7bb0a852e671bd0e57d4bf2dda5af448d9a1e38aebe, the patch was changed without updating the line count.

Within the commits of the PR that was squashed to make that commit, I see this tighter diff which only adds one line and makes the patch wrong: https://github.com/dotnet/source-build/commit/d0caaeace71ba5da8b6f6f2a94f295f2d93d1aff#diff-c72afe4ad8f6530dfdc0d7bb0a852e671bd0e57d4bf2dda5af448d9a1e38aebe.

Maybe the patch text was manually modified without updating the rest of the patch properly? Seems like the simplest answer to me.

dagood:
Merging past https://github.com/dotnet/source-build/issues/1937.

PR #1949, opened 2020-12-10
Re enable bootstrap build job in 5.0 CI

dseefeld:


crummel (approved):

patches/roslyn-analyzers/0010-Build-roslyn-analyzer-tools-with-net5.0-TFM.patch
```diff
@@ -0,0 +1,64 @@
...
+   <Target Name="BuildGenerateAnalyzerNuspecFile"
+diff --git a/src/Tools/GenerateAnalyzerNuspec/GenerateAnalyzerNuspec.csproj b/src/Tools/GenerateAnalyzerNuspec/GenerateAnalyzerNuspec.csproj
+index ba0a961..3a16ac4 100644
+--- a/src/Tools/GenerateAnalyzerNuspec/GenerateAnalyzerNuspec.csproj
++++ b/src/Tools/GenerateAnalyzerNuspec/GenerateAnalyzerNuspec.csproj
+@@ -2,6 +2,7 @@
+   <PropertyGroup>
+     <OutputType>Exe</OutputType>
+     <TargetFramework>netcoreapp3.1</TargetFramework>
++    <TargetFramework Condition="'$(DotNetBuildFromSource)' == 'true'">net5.0</TargetFramework>
```
dagood, line 34: I'm curious why this wouldn't work as `netcoreapp3.1`... I would think if executed, these would roll forward. Why isn't this a problem with the Microsoft build too? (Since there's also only a .NET 5 runtime installed there.)

Building with `net5.0` is now source-build standard operating procedure, so this is a fine change as far I'm concerned. 😄 I just don't quite understand why it failed without it and think it would be worthwhile to capture that.

dagood (approved):

patches/roslyn-analyzers/0010-Build-roslyn-analyzer-tools-with-net5.0-TFM.patch
dseefeld: It's not a problem with the stage1 build because source-build installs a 3.1 runtime to use with darc.  https://github.com/dotnet/source-build/blob/be50c2bb920fb623c83423771f5417e54eac9434/build.sh#L108-L109  
We don't do that in the final-sdk build. 
Also, in the Microsoft build for aspnetcore, they're installing a 3.1.4 runtime.

PR #1965, opened 2021-01-08
Build MSBuild.SourceBuild.slnf (new solution filter)

ladipro:
https://github.com/dotnet/msbuild/pull/6010 replaced `MSBuild.SourceBuild.sln` with `MSBuild.SourceBuild.slnf`. The subset of projects it's building stays the same.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -52,7 +52,7 @@
     </Dependency>
     <Dependency Name="Microsoft.Build" Version="16.8.0" CoherentParentDependency="Microsoft.Dotnet.Sdk.Internal">
       <Uri>https://github.com/dotnet/msbuild</Uri>
-      <Sha>126527ff107ae93fed10af675506c56d046aa5a3</Sha>
+      <Sha>c6ed2fef9d1cb5504eb51c6f6a9d529e4efd9054</Sha>
     </Dependency>
     <Dependency Name="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.0" CoherentParentDependency="Microsoft.NET.Sdk">
       <Uri>https://github.com/dotnet/roslyn-analyzers</Uri>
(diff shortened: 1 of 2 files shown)


crummel:
I updated so this PR is actually getting the version with the slnf file, but that might not be all it takes to fix it.

eng/Version.Details.xml
```diff
@@ -52,7 +52,7 @@
     </Dependency>
     <Dependency Name="Microsoft.Build" Version="16.8.0" CoherentParentDependency="Microsoft.Dotnet.Sdk.Internal">
       <Uri>https://github.com/dotnet/msbuild</Uri>
-      <Sha>126527ff107ae93fed10af675506c56d046aa5a3</Sha>
+      <Sha>c6ed2fef9d1cb5504eb51c6f6a9d529e4efd9054</Sha>
```
dagood, line 55: Probably shouldn't update our 5.0 release branch (source-build `master`) to a 6.0 commit (msbuild `master`). 😛

dagood (changes requested):
We don't have a branch for 6.0, so reacting to the slnf change isn't actually possible yet. Can let CI keep going to see if this brings up any forward-looking issues that need to be addressed though.

eng/Version.Details.xml
rainersigwald: MSBuild master is 16.9 is 5.0.200.

dagood: Oh, interesting. We do only build 5.0.1xx, so the logic still applies, but thanks for the correction.

crummel:
@dseefeld will be looking into this in relation to the 6.0 preview1 PR: https://github.com/dotnet/source-build/pull/2044.

crummel:
Triaged: superseded by https://github.com/dotnet/source-build/pull/2044.

Issue #1571, opened 2020-04-27
Quality of life "darc clone --depth 0" improvements


dagood:
* Add parallelism.
* Reduce log noise.
  * The length right now exceeds my console scroll buffer size. 😋 
  * Trim fat. We don't care about dependencies in `--depth 0` mode, so `Abandoning dependency` and similar are clutter.
  * Perhaps log details to file, and by default send only enough to stdout to assure the user that a long-running process is happening and progressing.
* Remove the "master copy" of the repo. Spawn worktrees from the bare git repo in `.git` instead.
* Shorten commit hash length to make paths easier to read.

Some of this may be better to track in dotnet/arcade, but wanted to write them down.

I implemented some of this in https://github.com/dotnet/arcade-services/pull/1112, but they aren't easy to extract because that PR was basically an overhaul.

/cc @crummel

dagood:
I think we should consider removing our dependency on `darc` and move to a build task maintained in dotnet/arcade for arcade-powered source-build.

* Getting `darc` set up is complicated. There's an active dotnet/source-build bug where our infra tries to put it in a weird place: https://github.com/dotnet/source-build/issues/1711. With arcade-powered source-build, I expect more people to start running similar scenarios, and it would be great if this tooling was simply part of the Arcade SDK.
* There's less need for code sharing. Since we only use `--depth 0` now, we don't need to do any of the complex coherency parsing and analysis. I think we only share code for Version.Details.xml parsing and `git clone/fetch/checkout` commands now, and I'm not sure that's worth it.
* It's (subjectively) easier for us to maintain code in dotnet/arcade than dotnet/arcade-services. dotnet/arcade-services doesn't branch for new product versions. (For good reason, but it raises the bar and back-compat needs for changes.) Also, this single `darc` command is the only thing source-build folks maintain in that repo, so we're not very familiar with it. (E.g. I'm not sure how to add new tests, and I hacked up my build to work with it last time because it has some specific VS workload requirements.)
* It's (also subjectively) easier to test changes in dotnet/arcade. It builds easily on Linux, fits into the MSBuild nupkg ecosystem that I'm already familiar with, and can be accompanied by other changes in the Arcade SDK I need to do at the same time as my other testing.
[...]

PR #1975, opened 2021-01-12
Update to 5.0.102

dseefeld:


repos/aspnetcore.proj
```diff
@@ -1,4 +1,7 @@
 <Project>
+  <PropertyGroup>
+    <SourceDirectory>aspnetcore</SourceDirectory>
```
dagood, line 3: Not necessary when the same as the project name. Not worth respinning to remove it though, IMO. Will be nice when we can fix up cloning. (https://github.com/dotnet/source-build/issues/1534 / #1571)

dagood (approved):

crummel (approved):

Issue #1977, opened 2021-01-13
OSX build machine issues


crummel:
See https://github.com/dotnet/source-build/pull/1963

MichaelSimons:
[Triage] OSX support is tracked with #2909

Issue #1794, opened 2020-10-09
macOS Big Sur, during CoreFX 3.1: "EXEC : error : include location '/usr/local/include' is unsafe for cross-compilation"


hrytskivr:
Hey, trying to build on macOS and have this error, no clue what to do.

```
...
    /Users/metal/Downloads/source-build/artifacts/src/corefx.3009a1fc20d9cb47d117399040d31c23f7b23e02/src/Native/build-native.sh x64 Release OSX outconfig netcoreapp-OSX-Release-x64 -portable
    Setting up directories for build
    Checking pre-requisites...
    Commencing build of corefx native components for OSX.x64.Release
    Invoking "/Users/metal/Downloads/source-build/artifacts/src/corefx.3009a1fc20d9cb47d117399040d31c23f7b23e02/src/Native/Unix/gen-buildsys-clang.sh" "/Users/metal/Downloads/source-build/artifacts/src/corefx.3009a1fc20d9cb47d117399040d31c23f7b23e02/src/Native/Unix" "3" "9" "x64" "RELEASE" " -DFEATURE_DISTRO_AGNOSTIC_SSL=0 -DCMAKE_STATIC_LIB_LINK=0"
    -- Configuring done
    -- Generating done
    -- Build files have been written to: /Users/metal/Downloads/source-build/artifacts/src/corefx.3009a1fc20d9cb47d117399040d31c23f7b23e02/artifacts/obj/native/netcoreapp-OSX-Release-x64
    Executing make install -j 1
    [ 14%] Built target System.IO.Compression.Native-Static
    [ 28%] Built target System.IO.Compression.Native
    [ 30%] Built target System.IO.Ports.Native
    [ 32%] Built target System.IO.Ports.Native-Static
    [ 33%] Building C object System.Native/CMakeFiles/System.Native-Static.dir/pal_console.c.o
  EXEC : error : include location '/usr/local/include' is unsafe for cross-compilation [-Werror,-Wpoison-system-directories] [/Users/metal/Downloads/source-build/artifacts/src/corefx.3009a1fc20d9cb47d117399040d31c23f7b23e02/src/Native/build-native.proj]
  ##vso[task.logissue type=error;sourcepath=EXEC;linenumber=0;columnnumber=0;code=;]include location '/usr/local/include' is unsafe for cross-compilation [-Werror,-Wpoison-system-directories%5D
    1 error generated.
    Failed to build corefx native components.
[...]

PR #1980, opened 2021-01-13
Update to Jan 2021 3.1.11 release.

crummel:


dseefeld (approved):

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat/RemoveInternetSourcesFromNuGetConfig.cs
```diff
@@ -67,6 +68,9 @@ public override bool Execute()
 
             packageSourcesElement.ReplaceNodes(local.ToArray());
 
+            // Remove disabledPackageSources element so if any internal packages remain, they are used in source-build
+            disabledPackageSourcesElement?.ReplaceNodes(new XElement("clear"));
```
dagood, line 72: Hmm... do you know why this works for non-internal builds? I would expect it to try to use internal sources in the production build and fail, if we remove the disables unconditionally. It obviously didn't fail, but curious why.

dagood (approved):

tools-local/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat/RemoveInternetSourcesFromNuGetConfig.cs
crummel: I don't.  My guess would be it's hitting the public feeds first, we've seen some nondeterministic behavior with which NuGet feed gets hit first before.  Think it's worth removing this?

dagood: I think we should go ahead with and merge the PR, but also file an issue to look into it.

dagood: Ahh right, poked around a bit, and we actually remove all internal sources all the time unless specifically included by specific projects when they need an internal prebuilt dependency like templates:

https://github.com/dotnet/source-build/blob/8cad9c79cb726997804c415782eaac07168e40d8/repos/core-sdk.proj#L77-L84

That only happens when auth is provided, so it works fine externally. Case closed. 😄

crummel: Cool, thanks for looking into it!

crummel:
OSX failures are known issues: https://github.com/dotnet/source-build/issues/1977, https://github.com/dotnet/source-build/issues/1794.

PR #1995, opened 2021-01-21
Let Microsoft.DotNet.Build.Tasks.Packaging set default icon

dagood:
I'd like to see why this patch is necessary. (Or, hopefully, see that it isn't. 😄) There isn't a clear way to integrate it into dotnet/arcade and it would be nice to just delete it. It was originally added for 5.0.0-rc1: https://github.com/dotnet/source-build/pull/1782.


crummel (approved):

crummel:
Yeah, I think I remember that they worked around this issue a different way.

dseefeld (approved):

dseefeld:
It was originally added in 3.1 and brought forward to 5.0.  https://github.com/dotnet/source-build/blob/release/3.1/patches/arcade/0014-Don-t-attempt-to-update-icon.patch  I agree with getting rid of it.

PR #2002, opened 2021-01-26
[master] Set up arcade-powered source-build CI

dagood:


dagood:
Issue with the inner clone: 😢
1. The repo builds in the standard managed-only arcade-specified source-build container.
   * `mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-3e800f1-20190501005343`, `git version 1.8.3.1`
1. The Checkout CI step makes a shallow clone (depth 3).
   * It looks like this hasn't been the case in other repos we've added CI to so far, so we haven't encountered it yet.
1. The inner clone script tries to clone the outer repo into a subdirectory.
   * Git fails because it doesn't allow cloning from a shallow repo into another place.
   * `fatal: attempt to fetch/clone from a shallow repository`

Using a new distro (1) would make this work. This limitation of `git clone` was actually lifted in more recent versions with https://github.com/git/git/commit/92251b1b5b5e53ac9de890105e2a2bd9d15e2ecb. (It works for me on Fedora with `git version 2.25.4`.) Unfortunately https://github.com/git/git/commit/92251b1b5b5e53ac9de890105e2a2bd9d15e2ecb is only 7 years old, so it isn't available everywhere. 😛 We don't have much freedom to use a newer distro, because the point is to make sure everything's buildable on the oldest distro .NET can support.

Fixing at (2) is possible: we can probably set it up in our yaml template to override the CI pipeline and always do a full clone. Or, have the inner clone script unshallow it (https://stackoverflow.com/a/6802238). This makes CI slower, though.

We can try a hacky fix for (3): just before the clone, delete `.git/shallow`, and put it back afterwards. https://stackoverflow.com/q/4770532. This lets us keep everything shallow.

dagood:
Mock official build: https://dev.azure.com/dnceng/internal/_build/results?buildId=976047&view=results

crummel (approved):

PR #2008, opened 2021-01-29
[release/5.0] Build aspnetcore for net46, to provide OmniSharp web project dependency

dagood:
This PR simply removes the patch to try to see in CI why the patch is necessary (if it is). For https://github.com/dotnet/source-build/issues/2006


michaelbarkdoll:
Thats a lot of fails for a lot of OS.  How will it be corrected?

dagood:
That just means that there was a reason for the patch. Someone will need to look into the error and figure out what changes are necessary--it's very open-ended.

Based on this error:

>   /src/artifacts/src/aspnetcore.2670c128d522473e146ff9f8159bfffdfe694cd9/src/Razor/Microsoft.NET.Sdk.Razor/src/Microsoft.NET.Sdk.Razor.csproj : error NU1202: Package Microsoft.Build.Framework 16.8.3 is not compatible with net46 (.NETFramework,Version=v4.6). Package Microsoft.Build.Framework 16.8.3 supports: netstandard2.0 (.NETStandard,Version=v2.0) [/src/.dotnet/sdk/5.0.100/NuGet.targets]

I would guess we need to get rid of this `DotNetBuildFromSource` condition:

https://github.com/dotnet/msbuild/blob/39993bd9d02917993b6147a9973dc3aa6e9b00c7/src/Directory.Build.props#L29-L30

```xml
    <LibraryTargetFrameworks>$(FullFrameworkTFM);netstandard2.0</LibraryTargetFrameworks>
    <LibraryTargetFrameworks Condition="'$(DotNetBuildFromSource)' == 'true'">netstandard2.0</LibraryTargetFrameworks>
```

PR #2019, opened 2021-02-03
[master] Clarify ArPow initial setup requirements, binlog

dagood:
Add a quick overview of the initial setup and specify where the binlogs are. Should help out with the first steps if we get more people working from this doc in the future.


crummel (approved):

omajid (approved):
:+1: on the changes. This clarifies things that I found confusing on my first run.

PR #2020, opened 2021-02-03
[master] Update ArPow onboarding docs: include patches, onboarding stage ownership

dagood:
/cc @omajid @crummel @dseefeld


omajid (approved):
This is very helpful for a dummy like me!

PR #2032, opened 2021-02-09
[release/2.1] NuGet feed update

crummel:


smoke-testNuGet.Config
```diff
@@ -4,9 +4,6 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="source-built-packages" value="SOURCE_BUILT_PACKAGES" />
-    <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
-    <add key="dotnet/feed/dotnet-core" value="https://dotnetmygetlegacy.blob.core.windows.net/dotnet-core/index.json" />
-    <add key="ProdCon" value="PRODUCT_CONTRUCTION_PACKAGES" />
-    <add key="smoke-test feed" value="SMOKE_TEST_PACKAGE_FEED" />
```
dagood, line 10: Are these "template" feeds needed for some smoke-test scenarios? Maybe track along with the ProdCon feed?

dagood (approved):

smoke-testNuGet.Config
dagood: It looks like this change is the same across the branches. Curious what will happen in CI.

ProdConFeed.txt
```diff
@@ -1 +1 @@
-https://dotnetfeed.blob.core.windows.net/orchestrated-release-2-1/20201118-01/final/index.json
+https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public%40Local/nuget/v3/index.json
```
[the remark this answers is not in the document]
crummel: Logged https://github.com/dotnet/source-build/issues/2033.

smoke-testNuGet.Config
crummel: It worked internally, I think dotnet-public includes all the feeds we need for right now.  I will add a note that this may need to be changed to that same issue.

PR #2045, opened 2021-02-09
Update to 3.1.112

dseefeld:


smoke-testNuGet.Config
```diff
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
-    <clear />
     <add key="source-built-packages" value="SOURCE_BUILT_PACKAGES" />
```
dagood, line 4: Is this intentional? Should keep the clear if possible.

dagood (approved):

smoke-testNuGet.Config
dseefeld: Unintentional.  Thanks for catching it.

crummel (approved):

Issue #2034, opened 2021-02-09
ArPow: Handle official build output that is too large to publish to Azure DevOps feeds


dagood:
After getting dotnet/runtime to build with ArPow infra, the intermediate nupkg was 1.092 GB. This is significantly more than the <500&nbsp;MB that I found a while ago when I took an official dotnet/runtime build and put all its outputs in a zip.

The big difference is symbol stripping. I got a 422 MB intermediate nupkg by enabling symbol stripping (normally disabled for source-build, so the distro can strip the symbols into debuginfo packages later on). Removing `.symbols.nupkg` (sympkgs) that ArPow doesn't need helps, but not a whole lot. Data:

  | Remove sympkg | Keep sympkg
-- | --: | --:
Strip symbols (native) | 182 MB | 422 MB
Keep symbols | 741 MB | 1,092 MB

The AzDO feed limit is 500 MB. https://docs.microsoft.com/en-us/azure/devops/artifacts/reference/limits?view=azure-devops. But when I tried publishing the 422 MB file to a test feed, I wasn't able to get it to go through--I got a 503 error instead. 😕 I don't know if I'm hitting a bug, or if the true upper limit is somewhere lower than 500 MB.

The [planning doc says](https://github.com/dotnet/source-build/blob/master/Documentation/planning/arcade-powered-source-build/intermediate-nupkg.md#too-large) we should just split it up. I don't see any problem with this yet. The ArPow infra in Arcade will probably need to be adjusted to allow it.

Stripping symbols and removing sympkgs to get down to 182 MB is an option. As long as downstream repos don't change their behavior when they get stripped vs. unstripped binaries, it don't affect anything. This is a regression risk, however.

dagood:
There's a broader perf concern we should pay attention to here. (Thanks @MichaelSimons for pointing this out.) With large intermediate nupkgs, it'll take time to publish, but the impact on the downstream repos that have to download these files as part of CI is IMO more important.

Splitting the intermediate nupkg up helps there: repos could download only what they need. In theory, we can crank granularity all the way up (1 per output file), and waste as little as possible. Getting a reasonable design on the downstream side for how these packages are addressed is important.

Also, maybe we should go ahead and strip symbols, if the perf concern outweighs the risk of regression.

PR #2053, opened 2021-02-11
ArPow: Add plan for supplemental intermediate nupkgs

dagood:
I started writing out what I was considering in https://github.com/dotnet/source-build/issues/2034, but I wrote enough that it fits better in the docs. Also, line review comments are always nice, rather than just one issue thread.

So far I'd only expect these to be needed for dotnet/runtime -> dotnet/aspnetcore -> dotnet/installer, perhaps dotnet/sdk -> dotnet/installer. (Also dotnet/runtime -> repos that need platform extensions libraries.)

/cc @MichaelSimons @omajid

### DIFF
--- a/Documentation/planning/arcade-powered-source-build/intermediate-nupkg.md
+++ b/Documentation/planning/arcade-powered-source-build/intermediate-nupkg.md
@@ -98,16 +98,16 @@ download sizes:
 
 Name | Download size (approx)
 -- | --
-Microsoft.NETCore.App.Ref | 1 MB
-Microsoft.NETCore.App.Ref.archive | 1 MB
-Microsoft.NETCore.App.Runtime | 107 MB
-Microsoft.NETCore.App.Runtime.archive | 93 MB
-Microsoft.NETCore.App.Host | 47 MB
-Microsoft.NETCore.App.Host.archive | 47 MB
-Microsoft.NETCore.App.Crossgen2 | 167 MB
-Microsoft.NETCore.App.Crossgen2.archive | 167 MB
-libraries | 6 MB
-coreclr (ilasm/ildasm/testhost) | 6 MB
+RefPack | 1 MB
+RefPack.Archive | 1 MB
+RuntimePack | 107 MB
+RuntimePack.Archive | 93 MB
+HostPack | 47 MB
+HostPack.Archive | 47 MB
+Crossgen2Pack | 167 MB
+Crossgen2Pack.Archive | 167 MB
+Libraries | 6 MB
+CoreCLR (ilasm/ildasm/testhost) | 6 MB
 
 For context, some other repositories' intermediate nupkg total download size
 that do not need any supplemental nupkgs:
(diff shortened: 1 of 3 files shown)


Documentation/planning/arcade-powered-source-build/intermediate-nupkg.md
```diff
@@ -35,14 +35,90 @@ source-built intermediates.
...
+* `Microsoft.SourceBuild.Intermediate.runtime.libraries.linux.x64/6.0.0-foo`
+  * Contains the dotnet/runtime platform extensions libraries.
+* `Microsoft.SourceBuild.Intermediate.runtime.Microsoft.NETCore.App.Crossgen2.linux.x64/6.0.0-foo`
+  * Contains the crossgen2 pack nupkg.
+  * `Microsoft.NETCore.App.Crossgen2` is the name of the crossgen2 pack without
+    its RID, so use that as the name by convention.
+* `Microsoft.SourceBuild.Intermediate.runtime.Microsoft.NETCore.App.Crossgen2.archive.linux.x64/6.0.0-foo`
+  * This stores the `tar.gz` version of the crossgen2 pack.
+  * Crossgen2 is huge (~167 MB), so we need one supplemental nupkg for each
+    artifact--the nupkg and the archive.
```
dseefeld, line 55: When will the archive be used vs. the nupkg?  When will both be produced instead of just the nupkg?

Documentation/planning/arcade-powered-source-build/intermediate-nupkg.md
```diff
@@ -35,14 +35,90 @@ source-built intermediates.
...
+  * The name should apply to any RID, so I use `archive` rather than `tarball`
+    to allow Windows support in the future.
+
+Each nupkg should either be < 20 MB, or contain only one artifact. This is a
+heuristic to use to decide when to produce supplemental nupkgs and how to split
+them up.
+
+The intermediate nupkg we produce now (non-supplemental) contains:
+
```
dseefeld, line 64: Supplemental nupkgs will just contain artifacts?

Documentation/planning/arcade-powered-source-build/intermediate-nupkg.md
dagood: They're always both produced--this is talking about the difference between e.g. `dotnet-runtime-6.0.0-ci-centos.7-x64.tar.gz` and `Microsoft.NETCore.App.Runtime.centos.7-x64.6.0.0-ci.nupkg`. On the usage side, I haven't quite kept track, but I think they're both used for various reasons when piecing together the SDK. The nupkg version might be more likely to be used when the SDK targets are involved, because officially, the only way to get these is from nuget.org.

(I'm going to add the artifact filenames here in case there was just a missing connection.)

Documentation/planning/arcade-powered-source-build/intermediate-nupkg.md
dagood: Yep. Since the "main" package and all supplementals are built at the same time inside an official build, there's no data that's really unique to the supplemental package except the artifact itself. The "main" package is always restored and carries the reports (and list of supplemental packages).

dseefeld (approved):

PR #2057, opened 2021-02-16
Fix broken Arcade Powered Source-build link

MichaelSimons:
The original `[Arcade-powered source-build](Documentation\planning\arcade-powered-source-build)` link produces a 405 because it gets translated to `https://github.com/dotnet/source-build/blob/master/Documentation%5Cplanning%5Carcade-powered-source-build` in GitHub.   Other tooling handles the back slashes just fine like VsCode.


dagood (approved):
Good catch. I think I used "copy relative path" in VS Code on Windows. (And didn't account for the platform difference, and didn't double-check it on GitHub.)

It looks like the leading `./` and trailing `/` aren't necessary, are they functional, or for clarity?

MichaelSimons:
I am happy to remove the leading './' and '/'.  They came from referencing other source-build links 😸.

PR #2093, opened 2021-03-16
[release/5.0] Add missing TRUE and FALSE definitions for OSX

crummel:
Local OSX doesn't repro, using this to figure out the issue.

### DIFF
--- a/patches/runtime/0030-Add-TRUE-and-FALSE-definitions.patch
+++ b/patches/runtime/0030-Add-TRUE-and-FALSE-definitions.patch
@@ -1,0 +1,191 @@
+From 15f59106c7f6cb99945d55e9ac280565e28e22bc Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Wed, 17 Mar 2021 01:32:33 -0500
+Subject: [PATCH 29/29] Add missing TRUE and FALSE definitions.
+
+---
+ .../Unix/System.Globalization.Native/pal_calendarData.c   | 8 ++++++++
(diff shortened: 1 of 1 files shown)


patches/runtime/0030-Add-TRUE-and-FALSE-definitions.patch
```diff
@@ -0,0 +1,191 @@
+From 15f59106c7f6cb99945d55e9ac280565e28e22bc Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Wed, 17 Mar 2021 01:32:33 -0500
+Subject: [PATCH 29/29] Add missing TRUE and FALSE definitions.
```
dseefeld, line 4: What is the runtime repo doing about this?  Are they encountering the same issue?  I'm assuming this is a short-lived patch and will be removed in the next servicing release?

dseefeld (approved):

patches/runtime/0030-Add-TRUE-and-FALSE-definitions.patch
crummel: runtime appears to use a specific OSX image.  I've logged https://github.com/dotnet/source-build/issues/2100 for us to look into this, that seems like it would really help us out.

crummel: And yes, this is intended to be removed as soon as the lab machines are updated to Big Sur (or we switch to a frozen image).

Bo98:
This is the relevant runtime issue: https://github.com/dotnet/runtime/issues/47346

PR #2136, opened 2021-04-21
Correct specification of itemPattern

dseefeld:


.vsts.pipelines/stages/prepare-artifacts.yml
```diff
@@ -57,7 +58,8 @@ stages:
         downloadType: single
         artifactName: 'Tarball ${{ parameters.gatherPortableJob }}'
         downloadPath: $(portableSourceBuiltStageDir)
-        itemPattern: /*/Private.SourceBuilt.Artifacts.*.tar.gz
+        itemPattern: |
```
MichaelSimons, line 61: Curious what the key change was or if all are necessary?

1. Removal of leading /
2. Replacing * with **
3. Multi-line declaration e.g. |

crummel (approved):

.vsts.pipelines/stages/prepare-artifacts.yml
dseefeld: Yes, yes and yes.  I needed to remove the leading /.  I needed ** to match all prior directories.  For some reason, yaml doesn't like a pattern starting with ** on the same line as the tag, so I needed the multi-line declaration.

PR #2175, opened 2021-05-20
[release/5.0] Add x64 RHEL7 and 8 builds.

crummel:
Internal build: https://dev.azure.com/dnceng/internal/_build/results?buildId=1180681&view=results.  This won't be visible in the PR because the RHEL builds are internal-only.

### DIFF
--- a/.vsts.pipelines/builds/matrix.yml
+++ b/.vsts.pipelines/builds/matrix.yml
@@ -54,6 +54,7 @@ stages:
     name: debian9
     jobParameters:
       imageName: mcr.microsoft.com/dotnet-buildtools/prereqs:debian-stretch-d61254f-20190807161114
+      cleanAptGet: true
     matrix:
       Production: {}
       Online: { type: Online }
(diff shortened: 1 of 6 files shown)


.vsts.pipelines/builds/matrix.yml
```diff
@@ -98,6 +98,38 @@ stages:
       scriptSuffix: .sh
       setupMac: true
 
+# RHEL builds only internal
+- ${{ if not(and(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest'))) }}:
```
MichaelSimons, line 102: Why the condition on PullRequests only?

MichaelSimons (commented):
Has any work been done to capture how these new RHEL machines are configured (e.g. a setup script or Dockerfile)?  It would be great to have this in case we need to configure additional build agents in the future.  It would also be valuable for local dev scenarios.

Issue #2358, opened 2021-08-05
Update Newtonsoft.Json to latest version


dseefeld:
The source-build repo build 2 versions of Newtonsoft.Json (9.0.1 and 12.0.2).  There has been a concerted effort to move repos to the latest version (13.0.1).  The 12.0.2 version should be updated to 13.0.1 in source-build to support these new dependencies.  Once this is added, other repos that may still use the 12.0.2 version should have issues to use the 13.0.1 version.

For reference, here are the PRs for updating installer and sdk:
- installer - https://github.com/dotnet/installer/pull/11365
- sdk - https://github.com/dotnet/sdk/pull/19440

PR #2363, opened 2021-08-06
Update Newtonsoft.Json to 13.0.1

lbussell:
This PR updates the Newtonsoft.Json submodule to the latest stable release, 13.0.1 (https://github.com/JamesNK/Newtonsoft.Json/tree/13.0.1). I also had to re-make our patches, basing them them off of 13.0.1.

https://github.com/dotnet/source-build/issues/2358


MichaelSimons (approved):
Changes look good to me.  I wonder though if we should be leaving https://github.com/dotnet/source-build/issues/2358 open until the new version is consumed in source-build.  Perhaps this work should be tracked w/repos specific issues.

lbussell:
> Changes look good to me. I wonder though if we should be leaving #2358 open until the new version is consumed in source-build. Perhaps this work should be tracked w/repos specific issues.

Good call, let's make sure to leave it open.

dseefeld (approved):

PR #2383, opened 2021-08-17
[release/3.1] Exclude locally-built packages from smoke-test-prereqs.

crummel:
This makes two changes to prevent our locally-built packages from making it into the smoke-test-prereqs tarball:
- Remove any packages cached in the system cache after we copy it to our local NuGet cache (so we get the credential plugin).
- Don't copy packages to the prereqs directory after the local tests (which can restore locally-built packages).

This accomplishes two things:
- Smoke-testing will be more true to the partner and user experience (using the official packages that a developer would restore when actually using the product).
- The official packages will include all TFMs and runtimes so the smoke-test-prereqs tarball will be universal across all of our builds.


crummel:
Like the 5.0 version of this (#2382), I verified that the various legs of prereqs are identical and signed, so this appears to solve the problem.

dseefeld (approved):

PR #2615, opened 2021-11-12
Update to November 2021 3.1.21 / 3.1.415 Servicing Release

lbussell:
This is also a feature band update from 3.1.1xx to 3.1.4xx.


lbussell:
It looks like we're currently failing on a corefx/sourcelink issue. It looks similar to [this issue](https://github.com/dotnet/sourcelink/issues/458) but we already have that fix in our version of sourcelink, so it isn't exactly the same. @tmat does this issue look familiar to you?

MichaelSimons:
[Triage] Closing as this is getting rolled into the next servicing.

PR #2646, opened 2021-11-29
[release/3.1] Build base tarball

dseefeld:
Add generation of base tarball in release/3.1 branch. 

This is one that has no artifacts for reference-packages, previously source-built packages and dotnet SDK. Add prep.sh script to allow prepping the tarball to be run on a given platform. Also add generation of base tarball in source-build CI.


dseefeld:
osx issue is known.  OK to merge.

PR #2681, opened 2021-12-15
Update to December 3.1.416 release

crummel:


patches/core-sdk/0009-Experimental-NuGet-feed-patch.patch
```diff
@@ -0,0 +1,24 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Chris Rummel <crummel@microsoft.com>
+Date: Wed, 8 Dec 2021 00:44:07 -0600
+Subject: [PATCH 9/9] Experimental NuGet feed patch
```
MichaelSimons, line 4: Can you explain this patch in more detail?  Why is it necessary?  This subject makes it seems like this shouldn't be checked in.

MichaelSimons (approved):
One comment about a patch.

patches/core-sdk/0009-Experimental-NuGet-feed-patch.patch
crummel: This was only needed because it was an internal build, removed.

lbussell:
MacOS CI failure is due to https://github.com/dotnet/source-build/issues/2687

lbussell (approved):

PR #2691, opened 2021-12-16
[release/5.0] Add CI triggers for release/5.0 branches

lbussell:
This should make it possible to turn off pipeline trigger overrides in our internal CI.


.vsts.pipelines/builds/ci-internal.yml
```diff
@@ -1,3 +1,11 @@
+trigger:
```
MichaelSimons, line 1: I think we should consider removing the ci-internal.yml and corresponding build definitions if possible.  Same comment applies to the 3.1 PR.

MichaelSimons (approved):

lbussell:
The OSX CI failure is new but out of the scope of this PR. I filed this issue for it: https://github.com/dotnet/source-build/issues/2700

PR #2698, opened 2021-12-17
[release/5.0] Bump Darc version to allow for use of dotnet-runtime 5.0

ayakael:
Default version of Darc needs dotnet 3.1 runtimes for builds of dotnet 5.0. For the sake of keeping dotnet5 build decoupled from needing 3.1 runtimes, why not push Darc past merge dotnet/arcade#5412 so Darc uses dotnet-runtime 5.0? Can't push it too much, though, as latest version breaks with application-insights as it doesn't include a eng/Version.Details.xml file.

Made as part of Alpine Linux dotnet31 / dotnet5 packaging project, see dotnet/source-build#2695


crummel (approved):

crummel:
Thanks for making this change!

PR #2787, opened 2022-03-08
Update to .NET 5.0.15 / 5.0.212 March 2022 servicing release

lbussell:
See https://github.com/dotnet/announcements/issues/210


patches/aspnetcore/0019-Try-to-get-templates-to-build.patch
```diff
@@ -1,7 +1,7 @@
-From 42a4cb1b80128e87e014d91bbb349a8057e308e7 Mon Sep 17 00:00:00 2001
```
crummel, line 1: I usually try to avoid no-content changes to patches like this but since it's setting up the zero-commit and no-git-version stuff for later I think it's fine in this case.

crummel (approved):
One nitpick but LGTM.

Issue #2652, opened 2021-12-02
Automate the poison report generation


MichaelSimons:
Dependent on:

https://github.com/dotnet/source-build/issues/2633
https://github.com/dotnet/source-build/issues/2568

crummel:
Notes:
This is a two-part issue - in 3.1/5.0 the process is tooling build -> poison tooling builds if enabled -> poisoning runs -> source-build runs -> poison check runs.  So poison automation needs to be disabled for a dev/default build and CI will enable a flag to build and run poison tooling and checks.

PR #2812, opened 2022-04-04
Remove manual poison check in release checklist

MichaelSimons:
Poison check is now automated with https://github.com/dotnet/source-build/issues/2652 therefore the manual step is no longer needed in the release checklist.


crummel (approved):
🎉🎉🎉

Issue #2568, opened 2021-10-25
Document poison report format


crummel:
Also including common patterns to look for, how to add exceptions, types of poison hits

PR #2826, opened 2022-04-11
Add poisoning documentation.

crummel:
Fixes https://github.com/dotnet/source-build/issues/2568.

### DIFF
--- a/Documentation/leak-detection.md
+++ b/Documentation/leak-detection.md
@@ -8,7 +8,7 @@ Source-build includes a mechanism for *poisoning* its input files, and then for 
 
 Before the build, the MarkAndCatalogFiles task runs.  This does a few things:
 
-- Record the hash of every file in the source-build binary input directories (prebuilt and previously-source-built.  future expansion: reference-packages).  If the file is a zip, tarball, or nupkg, unpack it and do the same thing recursively.
+- Record the hash of every file in the source-build binary input directories (prebuilts and previously-source-built.  future expansion: [reference-packages](https://github.com/dotnet/source-build/issues/2817)).  If the file is a zip, tarball, or nupkg, unpack it and do the same thing recursively.
 - For managed DLLs, either bare or in an archive, add a custom attribute that marks the file as poisoned.
 - For nupkgs, drop a `.poisoned` file.
 - Repack the poisoned assemblies and extra files, removing signatures so they don't fail verification.
@@ -21,4 +21,4 @@ There's no change in source-build operation in poisoning mode during the build.
 
 ## After the build
 
-After the build, the CheckForPoison task is run on the source-build output directory.  It again unpacks any archives and packages recursively, and checks for all three kinds of markers injected before the build.  It then writes out a [report](poison-report-format.md) that details everything that was found.  This report is compared to the poison baseline to determine if the build should be marked with a warning or not. 
+After the build, the CheckForPoison task is run on the source-build output directory.  It again unpacks any archives and packages recursively, and checks for all three kinds of markers injected before the build.  It then writes out a [report](poison-report-format.md) that details everything that was found.
--- a/Documentation/leak-detection.md
+++ b/Documentation/leak-detection.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-Source-build includes a mechanism for *poisoning* its input files, and then for checking for that poison after the build is complete.  This allows us to ensure that all output files were built during the build rather than copied directly from the input, which would be an illegal prebuilt usage.
+Source-build includes a mechanism for *poisoning* its input files, and then for checking for that poison in the build output.  This allows us to ensure that all output files were built during the build rather than copied directly from the input, which would be an illegal prebuilt usage.
 
 ## Before the build
 
@@ -11,7 +11,7 @@ Before the build, the MarkAndCatalogFiles task runs.  This does a few things:
 - Record the hash of every file in the source-build binary input directories (prebuilts and previously-source-built.  future expansion: [reference-packages](https://github.com/dotnet/source-build/issues/2817)).  If the file is a zip, tarball, or nupkg, unpack it and do the same thing recursively.
 - For managed DLLs, either bare or in an archive, add a custom attribute that marks the file as poisoned.
 - For nupkgs, drop a `.poisoned` file.
-- Repack the poisoned assemblies and extra files, removing signatures so they don't fail verification.
+- Repack the poisoned assemblies and extra files, removing nupkg signatures so they don't fail verification.
 - Replace the binary inputs with these new packages and archives.
 - Record the hash of each file again, to make sure we will be able to track output files whether they were used poisoned or unpoisoned.
 
--- a/Documentation/leak-detection.md
+++ b/Documentation/leak-detection.md
@@ -1,0 +1,24 @@
+# Leak detection (poisoning)
+
+## Overview
+
+Source-build includes a mechanism for *poisoning* its input files, and then for checking for that poison after the build is complete.  This allows us to ensure that all output files were built during the build rather than copied directly from the input, which would be an illegal prebuilt usage.
+
+## Before the build
+
+Before the build, the MarkAndCatalogFiles task runs.  This does a few things:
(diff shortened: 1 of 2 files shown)


Documentation/leak-detection.md
```diff
@@ -0,0 +1,24 @@
+# Leak detection (poisoning)
+
+## Overview
+
+Source-build includes a mechanism for *poisoning* its input files, and then for checking for that poison after the build is complete.  This allows us to ensure that all output files were built during the build rather than copied directly from the input, which would be an illegal prebuilt usage.
```
MichaelSimons, line 5:
```suggestion
Source-build includes a mechanism for *poisoning* its input files, and then for checking for that poison int the build output.  This allows us to ensure that all output files were built during the build rather than copied directly from the input, which would be an illegal prebuilt usage.
```

Documentation/leak-detection.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Before the build
+
+Before the build, the MarkAndCatalogFiles task runs.  This does a few things:
+
+- Record the hash of every file in the source-build binary input directories (prebuilt and previously-source-built.  future expansion: reference-packages).  If the file is a zip, tarball, or nupkg, unpack it and do the same thing recursively.
```
MichaelSimons, line 11:
```suggestion
- Record the hash of every file in the source-build binary input directories (prebuilts and previously-source-built.  future expansion: [reference-packages](https://github.com/dotnet/source-build/issues/2817)).  If the file is a zip, tarball, or nupkg, unpack it and do the same thing recursively.
```

Documentation/leak-detection.md
```diff
@@ -0,0 +1,24 @@
...
+- For managed DLLs, either bare or in an archive, add a custom attribute that marks the file as poisoned.
+- For nupkgs, drop a `.poisoned` file.
+- Repack the poisoned assemblies and extra files, removing signatures so they don't fail verification.
```
MichaelSimons, line 14: Is "removing signatures" specific to the NuGet packages or does this apply to assemblies as well?

Documentation/leak-detection.md
```diff
@@ -0,0 +1,24 @@
...
+- Replace the binary inputs with these new packages and archives.
+- Record the hash of each file again, to make sure we will be able to track output files whether they were used poisoned or unpoisoned.
+
+## During the build
+
+There's no change in source-build operation in poisoning mode during the build.
+
+## After the build
+
+After the build, the CheckForPoison task is run on the source-build output directory.  It again unpacks any archives and packages recursively, and checks for all three kinds of markers injected before the build.  It then writes out a [report](poison-report-format.md) that details everything that was found.  This report is compared to the poison baseline to determine if the build should be marked with a warning or not.
```
MichaelSimons, line 24:
```suggestion
After the build, the CheckForPoison task is run on the source-build output directory.  It again unpacks any archives and packages recursively, and checks for all three kinds of markers injected before the build.  It then writes out a [report](poison-report-format.md) that details everything that was found.
```

I would shy away from the errors as warnings as that is a point in time decision.

This document should mention how to enable poisoning and that there is a baseline test.

Documentation/poison-report-format.md
```diff
@@ -0,0 +1,63 @@
...
+</PrebuiltLeakReport>
+```
+
+The first thing to notice is the different `Type`s.  These can help you track down where the leak came from.  One match can have multiple types, e.g. `<Type>AssemblyAttribute,Hash</Type>`.
+
+## Failure types
+
+### AssemblyAttribute
+
+This means that we detected a custom attribute that source-build adds to each prebuilt and previously-source-built assembly before the build.  This is a *definite* leak - there will not be any false positives.  However, this may not be cause for failure - some categories of assemblies may leak without causing problems:
```
MichaelSimons, line 29: Isn't it possible we would need to ship something from previously-source-built  (e.g. n-1) because of build ordering?

Documentation/poison-report-format.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Text files (licenses, templates, etc)
+- Empty files (\_.\_)
+- Anything else not processed by the build.
+
+However, if this is a DLL this is probably a legitimate leak.
+
+### NupkgFile
+
+This type of failure means that a nupkg was copied wholesale to the final output - we add a `.poisoned` file to each nupkg before the build and one was detected in the output.  This is a definite leak, but very uncommon and likely easy to trace in the binlogs.
```
MichaelSimons, line 47:
```suggestion
This type of failure means that a nupkg was copied wholesale to the final output.  This is a definite leak, but very uncommon and likely easy to trace in the binlogs.
```

Consider simplifying as is was covered earlier.

Documentation/poison-report-format.md
```diff
@@ -0,0 +1,63 @@
+# Interpreting the poison report
+
+The poison report looks something like this:
```
MichaelSimons, line 3: Suggestion: cross reference the leak-detection document.

Documentation/poison-report-format.md
```diff
@@ -0,0 +1,63 @@
...
+
+This means that we detected a custom attribute that source-build adds to each prebuilt and previously-source-built assembly before the build.  This is a *definite* leak - there will not be any false positives.  However, this may not be cause for failure - some categories of assemblies may leak without causing problems:
+
+- Localization assemblies, e.g. .../langcode/Assembly.Resources.dll
+- Reference assemblies (these should not ship, but also will very rarely cause problems).
+- Test or resource assemblies - these should be removed.
+
+### Hash
+
+Hash failures mean that a file matches the catalog of hashes that we take of existing files before the build.  These are commonly false positives due to:
```
MichaelSimons, line 37:
```suggestion
Hash failures mean that a file matches the catalog of hashes that we take of existing files before the build.  These are commonly false positives due to shipping static files not produced by the build such as:
```

I find it more meaningful to explain the high level view before listing examples and then the general explanation at the end.

Documentation/poison-report-format.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Text files (licenses, templates, etc)
+- Empty files (\_.\_)
+- Anything else not processed by the build.
```
MichaelSimons, line 41:
```suggestion
```

MichaelSimons (commented):
Thanks for writing this.

Documentation/leak-detection.md
crummel: Changed as suggested.

Documentation/leak-detection.md
crummel: Specified that this is the nupkgs only.

Documentation/poison-report-format.md
crummel: Yes, that's true.  Reworked this a bit.

Documentation/poison-report-format.md
crummel: Added, thanks.

Documentation/leak-detection.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Overview
+
+Source-build includes a mechanism for *poisoning* its input files, and then for checking for that poison in the build output.  This allows us to ensure that all output files were built during the build rather than copied directly from the input, which would be an illegal prebuilt usage.
+
+## Before the build
+
+Before the build, the MarkAndCatalogFiles task runs.  This does a few things:
+
+- Record the hash of every file in the source-build binary input directories (prebuilts and previously-source-built.  future expansion: [reference-packages](https://github.com/dotnet/source-build/issues/2817)).  If the file is a zip, tarball, or nupkg, unpack it and do the same thing recursively.
```
MichaelSimons, line 11:
```suggestion
- Record the hash of every file in the source-build binary input directories (prebuilts and previously-source-built, in the future [reference-packages](https://github.com/dotnet/source-build/issues/2817)).  If the file is a zip, tarball, or nupkg, unpack it and do the same thing recursively.
```

MichaelSimons (approved):
LGTM - I just had one minor formatting suggestion.

Issue #2994, opened 2022-08-17
internal/release/6.0.1xx branch fails to build after latest dependency flow


MichaelSimons:
See the latest [internal Microsoft build](https://dev.azure.com/dnceng/internal/_build/results?buildId=1947344&view=results)

```
    /tarball/src/sdk/artifacts/source-build/self/src/src/Cli/dotnet/Telemetry/Telemetry.cs(105,25): error CS1061: 'TelemetryClient' does not contain a definition for 'TelemetryConfiguration' and no accessible extension method 'TelemetryConfiguration' accepting a first argument of type 'TelemetryClient' could be found (are you missing a using directive or an assembly reference?) [/tarball/src/sdk/artifacts/source-build/self/src/src/Cli/dotnet/dotnet.csproj]
##[error]/tarball/src/sdk/artifacts/source-build/self/src/src/Cli/dotnet/Telemetry/Telemetry.cs(105,25): error CS1061: (NETCORE_ENGINEERING_TELEMETRY=Build) 'TelemetryClient' does not contain a definition for 'TelemetryConfiguration' and no accessible extension method 'TelemetryConfiguration' accepting a first argument of type 'TelemetryClient' could be found (are you missing a using directive or an assembly reference?)
 ```

PR #2996, opened 2022-08-18
[release/6.0] Update application-insights to 2.19.0

lbussell:
This should fix https://github.com/dotnet/source-build/issues/2994.

Also see
* https://github.com/dotnet/sdk/pull/22974

### DIFF
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -1,5 +1,5 @@
 <UsageData>
   <IgnorePatterns>
-    <UsagePattern IdentityGlob="*/*" />
+    <UsagePattern IdentityGlob="Microsoft.SourceBuild.Intermediate.*/*" />
   </IgnorePatterns>
 </UsageData>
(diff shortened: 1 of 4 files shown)


MichaelSimons (approved):

MichaelSimons:
@lbussell - Are these changes also needed?  https://github.com/dotnet/source-build-externals/pull/33

lbussell:
> Are these changes also needed?

Trying to figure that out now - will add them if needed.

MichaelSimons:
IIRC this is only an issue in the tarball where it removes the online feeds from the known nuget.configs.

MichaelSimons:
...The result is the potential for prebuilts because NuGet non-determinism.

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -1,5 +1,5 @@
 <UsageData>
   <IgnorePatterns>
-    <UsagePattern IdentityGlob="*/*" />
+    <UsagePattern IdentityGlob="Microsoft.SourceBuild.Intermediate.*/*" />
```
MichaelSimons, line 3: I don't think we should be enabling prebuilt detection with these changes.  There are likely other changes needed to get the repo level prebuilts clean.  We don't want to block this PR on them.

Issue #2710, opened 2022-01-07
Automate the publishing process of source tarballs for servicing releases


MichaelSimons:
This involves taking the source ~~and smoke-test-prereqs~~ tarball and publishing to the internal blob storage account with the correctly versioned names.

lbussell:
Providing some more clarity, the steps required to complete this work are:

- Read installer build information from our internal release (non-source-build) pipelines
- Upload dotnet source tarball to azure storage (dotnetcli)
- Upload source-build artifacts to azure storage (dotnetclimsrc)

Reference: https://github.com/dotnet/source-build/blob/3c8f23d08475e6f59a5ca889e0719b61e18db476/.github/ISSUE_TEMPLATE/releases/release_checklist.md

lbussell:
Smoke test prereqs will be handled by: https://github.com/dotnet/source-build/issues/3068

Issue #3065, opened 2022-10-13
Automate source-build release process for 6.0 and 7.0


lbussell:
### Describe the Problem

The source-build release process takes up a lot of time. Also, many of the release tasks are completed manually which leaves lots of room for individual error.

### Describe the Solution

The current release process is documented here: https://github.com/dotnet/source-build/blob/3c8f23d08475e6f59a5ca889e0719b61e18db476/.github/ISSUE_TEMPLATE/releases/release_checklist.md

The large majority of those release tasks should be automated. We can break this up into pipelines that can be run during different parts of the release cycle. I am proposing that we expand the current [source-build-release pipeline](https://github.com/dotnet/source-build/blob/3c8f23d08475e6f59a5ca889e0719b61e18db476/eng/source-build-release.yml) and create a new pipeline, **source-build-pre-release**.

`source-build-pre-release` (in order of execution, not priority):
- https://github.com/dotnet/source-build/issues/2710
- https://github.com/dotnet/source-build/issues/3068
- https://github.com/dotnet/source-build/issues/3064
- https://github.com/dotnet/source-build/issues/3069
- https://github.com/dotnet/source-build/issues/2904

In between `pre-release` and `release`, after security-partners-dotnet PRs are merged:
- https://github.com/dotnet/source-build/issues/3147
- https://github.com/dotnet/source-build/issues/3179
- https://github.com/dotnet/source-build/issues/3196

`source-build-release`:

- https://github.com/dotnet/source-build/issues/3080
- https://github.com/dotnet/source-build/issues/3062
- https://github.com/dotnet/source-build/issues/2888
- https://github.com/dotnet/source-build/issues/3058

lbussell:
In terms of priority, I see the following as most important:

- #3064: Updating the VMR-lite is done manually and prone to individual error
- #2904: We currently only manually test the VMR-lite

Everything else is fairly well defined, just manual work.

MichaelSimons:
@lbussell - where does #2921 fit in here?  Is it a duplicate of 3096?

premun:
The remaining open items are still relevant and I am not sure I'll have time to work on those so this might make sense to keep for tracking?

PR #3079, opened 2022-10-20
[main] Add source-build-pre-release pipeline yaml

lbussell:
Fixes https://github.com/dotnet/source-build/issues/2710. Part of https://github.com/dotnet/source-build/issues/3065

This is a new pipeline that:

1. Reads installer build information from our internal release (non-source-build) pipelines
2. Uploads the dotnet source tarball to azure storage

### DIFF
--- a/.github/ISSUE_TEMPLATE/releases/release_checklist.md
+++ b/.github/ISSUE_TEMPLATE/releases/release_checklist.md
@@ -15,13 +15,14 @@
      - This is useful to make sure context is available to review later. It may end up blank if the release goes very smoothly.
      - There are other notes on servicing in this OneNote. It may be useful to review if something goes wrong to see if it's been fixed before.
      - File issues appropriately as you encounter problems, and link to them from the notes. Provide info in the issue rather than in the notes.
-1. - [ ] Retrieve dotnet/installer commit sha from internal communications for the release.
+1. - [ ] Retrieve the final run of the [Stage-DotNet](https://dev.azure.com/dnceng/internal/_build?definitionId=792&_a=summary) pipeline from internal release communications.
+1. - [ ] Run the [source-build-pre-release](https://dev.azure.com/dnceng/internal/_build?definitionId=1188) pipeline. When staging the pipeline run, click "Resources" and provide select the final run of Stage-DotNet mentioned above.
+     - [ ] When the pipeline finishes, look in the logs for the dotnet/installer commit sha that represents the final release.
+     - This pipeline automatically uploads the dotnet source tarball to dotnetclimsrc
 1. - [ ] Ensure the official installer and tarball builds have completed for the release's commit sha.
-     - [ ] [Installer](https://dev.azure.com/dnceng/internal/_build?definitionId=286) (internal link)
-       - [ ] Retrieve the source tarball artifact - `BlobArtifacts/dotnet-sdk-source-6.0.xxx.tar.gz`
      - [ ] [Tarball](https://dev.azure.com/dnceng/internal/_build?definitionId=1011) (internal link)
        - [ ] Ensure the PoisonTests and SdkContentTests are passing.  Warnings indicate a baseline diff and should be inspected carefully.
-1. - [ ] [Internal] Gather smoke-test prereqs
+1. - [ ] [Internal] Gather smoke-test prereqs ([automation tracking issue](https://github.com/dotnet/source-build/issues/3068))
      - [ ] Retrieve smoke-test prereqs artifact for each architecture (e.g. x64 and arm64) from [tarball build](https://dev.azure.com/dnceng/internal/_build?definitionId=1011) (internal link)
        - [ ] x64 - `Build Tarball CentOS7-Offline_Artifacts/dotnet-smoke-test-prereqs.6.0.xxx.tar.gz`
        - [ ] arm64 - `Build Tarball Debian9-Offline_Artifacts/dotnet-smoke-test-prereqs.6.0.xxx.tar.gz`
@@ -32,13 +33,13 @@
        - [ ] Extact arm64 tarball
        - [ ] Copy four `*linux-arm64*` packages to x64 packages
        - [ ] Create new `dotnet-smoke-test-prereqs.6.0.xxx.tar.gz` tarball
-1. - [ ] [Internal] Upload source and smoke-test-prereqs tarball to dotnetclimsrc storage account.
+1. - [ ] [Internal] Upload smoke-test-prereqs tarball to dotnetclimsrc storage account.
 1. - [ ] Notify partners of release.  Include info about how certain we are that this will be the final Microsoft build.
      - [Internal] Send the dotnetclimsrc tarball links to partners.
          - Never overwrite a tarball. At least change the blob storage virtual dir to represent a new build. This can help avoid timing issues and make it more obvious if stale links were accidentally re-sent rather than new ones.
      - [Non-Internal] Send the dotnet/installer commit sha along w/link to publicly built source tarball.  Link to the public instructions for building source-build.
 1. - [ ] SYNC POINT: Wait for Microsoft build release.
-1. - [ ] Upload the source-build artifacts to dotnetcli/source-built-artifacts blob storage.
+1. - [ ] Upload the source-build artifacts to dotnetcli/source-built-artifacts blob storage. ([automation tracking issue](https://github.com/dotnet/source-build/issues/3080))
      - [ ] Retrieve the source-build artifacts from [Tarball](https://dev.azure.com/dnceng/internal/_build?definitionId=1011) (internal link) - `Build Tarball CentOS7-Offline_Artifacts/Private.SourceBuilt.Artifacts.6.0.xxx.tar.gz`
 1. - [ ] Run the [source-build-release pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=1124) (internal link).
      - Set the `SDK Version` parameter.
--- a/.github/ISSUE_TEMPLATE/releases/release_checklist.md
+++ b/.github/ISSUE_TEMPLATE/releases/release_checklist.md
@@ -16,7 +16,7 @@
      - There are other notes on servicing in this OneNote. It may be useful to review if something goes wrong to see if it's been fixed before.
      - File issues appropriately as you encounter problems, and link to them from the notes. Provide info in the issue rather than in the notes.
 1. - [ ] Retrieve the final run of the [Stage-DotNet](https://dev.azure.com/dnceng/internal/_build?definitionId=792&_a=summary) pipeline from internal release communications.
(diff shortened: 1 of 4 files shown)


eng/source-build-pre-release.yml
```diff
@@ -0,0 +1,146 @@
...
+- name: storageAccountName
+  value: 'dotnetclimsrc'
+- name: blobContainerName
+  value: 'source-build'
+- name: blobContainerUploadBaseFilePath
+  value: 'release'
+
+stages:
+
+- stage: GetBuildInfoStage
```
MichaelSimons, line 32: Why a different stage for retrieving the build info?  My concern is that it will take at least one order order of magnitude longer to acquire a build agent than both stages take to run.

eng/source-build-pre-release.yml
```diff
@@ -0,0 +1,146 @@
...
+    - name: InstallerCommit
+      value: $[ stageDependencies.GetBuildInfoStage.GetBuildInfoJob.outputs['ReadReleaseInfoStep.InstallerCommit'] ]
+    - name: SdkVersion
+      value: $[ stageDependencies.GetBuildInfoStage.GetBuildInfoJob.outputs['ReadReleaseInfoStep.SdkVersion'] ]
+    - name: RuntimeVersion
+      value: $[ stageDependencies.GetBuildInfoStage.GetBuildInfoJob.outputs['ReadReleaseInfoStep.RuntimeVersion'] ]
+    - name: ReleaseChannel
+      value: $[ stageDependencies.GetBuildInfoStage.GetBuildInfoJob.outputs['ReadReleaseInfoStep.ReleaseChannel'] ]
+    steps:
+    - script: |
```
MichaelSimons, line 99: Is this step necessary?  If seems like a step used to developer the initial pipeline.

eng/source-build-pre-release.yml
```diff
@@ -0,0 +1,146 @@
...
+        set -euo pipefail
+        az config set extension.use_dynamic_install=yes_without_prompt
+
+        installer_official_runs=$(az pipelines runs list --organization $(azDoOrg) --project $(azDoProject) --pipeline-ids $(installerOfficialCiPipelineId))
+        installer_official_run_id=$(echo $installer_official_runs | jq -r '.[] | select(.sourceVersion == "$(InstallerCommit)") | .id')
+        echo "dotnet-installer-official-ci run ID: ${installer_official_run_id}"
+
+        installer_tarball_build_runs=$(az pipelines runs list --organization $(azDoOrg) --project $(azDoProject) --pipeline-ids $(installerTarabllBuildCiPipelineId))
+        installer_tarball_build_run_id=$(echo $installer_tarball_build_runs | jq -r '.[] | select(.sourceVersion == "$(InstallerCommit)") | .id')
```
MichaelSimons, line 116: How does this behave if there are multiple builds for the same commit?

eng/source-build-pre-release.yml
```diff
@@ -0,0 +1,146 @@
...
+        project: 'internal'
+        pipeline: $(installerOfficialCiPipelineId)
+        runVersion: 'specific'
+        runId: $(InstallerOfficialRunId)
+    - script: |
+        set -euxo pipefail
+        az config set extension.use_dynamic_install=yes_without_prompt
+
+        source_tarball_upload_path="$(blobContainerUploadBaseFilePath)/$(ReleaseChannel)/$(RuntimeVersion)-$(SdkVersion)"
+        source_tarball_filename="dotnet-sdk-source-$(SdkVersion).tar.gz"
```
MichaelSimons, line 141: Since this is using the SdkVersion, should/can the download step do the same?

eng/source-build-pre-release.yml
```diff
@@ -0,0 +1,146 @@
...
+        az storage blob upload --account-name "$(storageAccountName)" --container-name "$(blobContainerName)" --file "$(PIPELINE.WORKSPACE)/${source_tarball_filename}" --name "${source_tarball_upload_path}/${source_tarball_filename}"
```
MichaelSimons, line 142: Does this fail gracefully if the download step fails?  My experience is the download step silently fails when patterns are used.  Is there a way to get it to fail?  If not should this step validate it exists before trying to upload?

.github/ISSUE_TEMPLATE/releases/release_checklist.md
```diff
@@ -15,13 +15,14 @@
      - This is useful to make sure context is available to review later. It may end up blank if the release goes very smoothly.
      - There are other notes on servicing in this OneNote. It may be useful to review if something goes wrong to see if it's been fixed before.
      - File issues appropriately as you encounter problems, and link to them from the notes. Provide info in the issue rather than in the notes.
-1. - [ ] Retrieve dotnet/installer commit sha from internal communications for the release.
+1. - [ ] Retrieve the final run of the [Stage-DotNet](https://dev.azure.com/dnceng/internal/_build?definitionId=792&_a=summary) pipeline from internal release communications.
+1. - [ ] Run the [source-build-pre-release](https://dev.azure.com/dnceng/internal/_build?definitionId=1188) pipeline. When staging the pipeline run, click "Resources" and provide select the final run of Stage-DotNet mentioned above.
```
MichaelSimons, line 19:
```suggestion
1. - [ ] Run the [source-build-pre-release](https://dev.azure.com/dnceng/internal/_build?definitionId=1188) pipeline. When staging the pipeline run, click "Resources" and select the final run of Stage-DotNet mentioned above.
```

.github/ISSUE_TEMPLATE/releases/release_checklist.md
```diff
@@ -15,13 +15,14 @@
...
+     - [ ] When the pipeline finishes, look in the logs for the dotnet/installer commit sha that represents the final release.
+     - This pipeline automatically uploads the dotnet source tarball to dotnetclimsrc
 1. - [ ] Ensure the official installer and tarball builds have completed for the release's commit sha.
```
MichaelSimons, line 22: Shouldn't this step be run prior to uploading the source tarball?

eng/source-build-pre-release.yml
lbussell: I used a different stage because I imagine that in the future multiple stages could run using the variables set by this stage (e.g. a separate stage for updating VMR).

Do you think these should be jobs instaed of stages and just run everything at once? Or do you think running `GetBuildInfoJob` in every stage would be faster?

eng/source-build-pre-release.yml
lbussell: It looks like it will output both IDs.

```bash
> json='[{"id": 1, "commit": "sha1"}, {"id": 2, "commit": "sha2"}, {"id": 3, "commit": "sha2"}]'
> echo $json | jq '.[] | select(.commit == "sha1") | .id'
1
> echo $json | jq '.[] | select(.commit == "sha2") | .id'
2
3
> result=$(echo $json | jq '.[] | select(.commit == "sha2") | .id')
> echo $result
2 3
```

I wonder what we should do in this situation. Select the most recent? Select the most recent that succeeded?

eng/source-build-pre-release.yml
lbussell: It actually does fail silently. Well not silently, but it doesn't have an exit code 1. I will look into this.

.github/ISSUE_TEMPLATE/releases/release_checklist.md
lbussell: I guess this should be another goal of the automation, to check that the installer and tarball builds were successful. I left it below the pre-release pipeline because the pipeline gives you the commit sha, and prevents you from having to look through the manifest.json yourself. It's not the end of the world to check it manually beforehand just for now though.

eng/source-build-pre-release.yml
lbussell: This is also a valid question for what we would do manually.

eng/source-build-pre-release.yml
MichaelSimons: I think these overall these steps aren't going to take much time and provisioning machines for each is going to be wasteful (e.g. plan for machine provisioning to take up to 20 minutes).  I would recommend utilizing as few stages/jobs as possible.

eng/source-build-pre-release.yml
MichaelSimons: Manually, I would expect a human could deduct why two builds existed.  One thing I would suggest is making sure the build passed.  I would say a fallback could be to have a variable to pass in the id.  If specified use the supplied build otherwise fail if there are multiple passing builds.

.github/ISSUE_TEMPLATE/releases/release_checklist.md
MichaelSimons: Agreed we should be checking the success state.

eng/source-build-pre-release.yml
lbussell: In my latest update I reduced everything to one job and multiple stages, and pulled out "Get Build Info" to a template so that it can be used in multiple jobs/pipelines.

eng/source-build-pre-release.yml
lbussell: In my latest change I try to detect if there are more than one build per commit and fail with an error if so. I also added the parameters for optional specific build IDs.

eng/templates/get-build-info.yml
```diff
@@ -0,0 +1,118 @@
+parameters:
+  - name: stageDotNetPipelineResource
```
MichaelSimons, line 2:
```suggestion
  - name: dotnetStagingPipelineResource
```

MichaelSimons (commented):
The new changes look good.  What's the resolution to the to open comments I made previously?

eng/source-build-pre-release.yml
lbussell: I added a check that the file is downloaded before the upload, and another check that the file is listed in blob storage afterwards.

.github/ISSUE_TEMPLATE/releases/release_checklist.md
lbussell: I added a check that the tarball build succeeded before continuing. If the build is red, then we fail the pipeline and require the pipeline run IDs to be manually provided.

crummel (approved):

eng/templates/get-build-info.yml
```diff
@@ -0,0 +1,124 @@
...
+        echo "dotnet-installer-source-build-tarball-build run ID: ${installer_tarball_build_run_id}"
+
+        echo "##vso[task.setvariable variable=InstallerOfficialRunId]${installer_official_run_id}"
+        echo "##vso[task.setvariable variable=InstallerTarballBuildRunId]${installer_tarball_build_run_id}"
+      name: GetAssociatedPipelineRunsStep
+      displayName: Get Associated Pipeline Run IDs
+      env:
+        AZURE_DEVOPS_EXT_PAT: $(System.AccessToken)
+
+  - ${{ if eq(parameters.useSpecificPipelineRunIDs, true) }}:
```
MichaelSimons, line 111: FYI usage of `- ${{ else }}` would make this more readable IMO.

Issue #3080, opened 2022-10-20
Automatically upload Private.SourceBuilt.Artifacts to blob storage on release day


lbussell:
This should go in the `source-build-release` pipeline that is run on release day (we can't upload release artifacts before release day).

PR #3095, opened 2022-10-28
[main] Refactor source-build-release pipeline and upload Private.SourceBuilt archive to blob storage

lbussell:
Fixes https://github.com/dotnet/source-build/issues/3080

Context: https://github.com/dotnet/source-build/issues/3065


eng/source-build-release.yml
```diff
@@ -6,198 +6,236 @@ pool:
...
+    default: false
+  - name: dotnetInstallerOfficialRunID
+    displayName: dotnet-installer-official-ci run ID
+    type: string
+    default: '200XXXX'
+  - name: dotnetInstallerTarballBuildRunID
+    displayName: dotnet-installer-source-build-tarball-build run ID
+    type: string
+    default: '200XXXX'
+  - name: testingMode
```
MichaelSimons, line 47: See previous `isDryRun` comment

eng/templates/variables/installer-pipelines.yml
```diff
@@ -0,0 +1,10 @@
+# Information about dotnet/installer pipelines that won't change and is useful in multiple pipelines
```
MichaelSimons, line 1: What do you see as the scope of these shared variables?  Is it more than just the two release pipelines?  I ask because there appear to be more variables that are shared between the two to declared in each pipeline .e.g blobContainerName

eng/source-build-release.yml
```diff
@@ -6,198 +6,236 @@ pool:
...
+      echo ${{ variables.announcementOrg }}/${{ variables.announcementRepo }} repo ID is ${repo_id}
+
+      query='query { repository(name: "${{ variables.announcementRepo }}", owner: "${{ variables.announcementOrg }}") { discussionCategories(first: 10) { edges { node { id, name } } } } }'
+      echo "${query}"
+      category_id=$( gh api graphql -f query="$query" --template '{{range .data.repository.discussionCategories.edges}}{{if eq .node.name "Announcements"}}{{.node.id}}{{end}}{{end}}' )
+      echo Discussion Category ID is ${category_id}
+
+      echo "##vso[task.setvariable variable=RepoId;]${repo_id}"
+      echo "##vso[task.setvariable variable=DiscussionCategoryId;]${category_id}"
+    displayName: Get category IDs
```
MichaelSimons, line 105: Would this be better named `Get Discussion Category ID`?

eng/source-build-release.yml
```diff
@@ -6,198 +6,236 @@ pool:
...
+      template="$(envsubst < ${template_path}/source-build-release-announcement.md)"
+      # Get the line in the template that is prefixed with "Title:" and remove the prefix
+      title=$(echo "$template" | grep "^Title:" | cut -d " " -f2-)
+      # Get the inverse of the above selection
+      body=$(echo "$template" | grep -v "^Title:")
+
+      query='mutation($RepoId: ID!, $categoryId: ID!, $body: String!, $title: String!) { createDiscussion(input: {repositoryId: $RepoId, categoryId: $categoryId, body: $body, title: $title}) { discussion { url } } }'
+
+      if [ ${{ parameters.testingMode }} = True ]; then
+        echo "Testing mode is set to True, not submitting announcement."
```
MichaelSimons, line 133: For diagnostic purposes, I think it would be useful to emit the announcement content as pipeline output.

eng/source-build-release.yml
lbussell: I had it named `Get Discussion Category ID` but changed the name as I combined the `Get Discussion Category ID` and `Get Repo ID` steps, they are both needed in future steps.

eng/templates/variables/installer-pipelines.yml
lbussell: These are the shared variables that are needed to run `get-build-info.yml`. Unfortunately you can't define `variables` in templates.

I would not consider the blob containers to be the same. They just happen to have the same name, but they're two separate blob containers in different storage accounts.

I'd be interested to hear any other suggestions you have for organizing variables.

MichaelSimons: Yes you raise a good point about the blob containers not needing to be the same.  I have no great suggestions here.

MichaelSimons (approved):

crummel (approved):

PR #3125, opened 2022-11-16
[release/6.0] Remove nuget config changes from humanizer patch

lbussell:
This might be needed in case we don't get a NuGet Security Analysis exception for security-partners-dotnet. Because humanizer has a vulnerable NuGet.config file, we need to remove it during tarball creation. That means the NuGet config doesn't exist at build time, so we need to remove the NuGet.config edits in this patch.

See https://github.com/dotnet/installer/pull/14963 for context.


MichaelSimons:
Closing as obsolete.

Issue #3064, opened 2022-10-13
Automatically create new VMR-lite branches in source-build-release pipeline


lbussell:
This should go in the new source-build-pre-release pipeline.

Use the dotnet source tarball from azure blob storage that will be uploaded by https://github.com/dotnet/source-build/issues/2710 (it could also work if it is uploaded manually). I already have a script I use to run the update manually here: https://gist.github.com/lbussell/fb72c5b32fef03c3f9de2e5b5e466a3c

This may require creating a bot for our internal security partners repo.

The steps required to completely automate this are:

1. Download dotnet source tarball from azure storage
2. Run a script similar to the one above to create a new update branch
3. Push the branch to dotnet-security-partners

PR #3130, opened 2022-11-22
[main] Automatically create new security-partners-dotnet branches with tarball contents in pre-release pipeline

lbussell:
Fixes https://github.com/dotnet/source-build/issues/3064

### DIFF
--- a/eng/source-build-pre-release.yml
+++ b/eng/source-build-pre-release.yml
@@ -71,7 +71,6 @@ jobs:
       runId: $(InstallerOfficialRunId)
 
   - template: templates/steps/upload-to-blob-storage.yml
-    condition: and(succeeded(), ne('${{ parameters.isDryRun }}', 'true'))
     parameters:
       file: '$(PIPELINE.WORKSPACE)/dotnet-sdk-source-$(SdkVersion).tar.gz'
       accountName: '$(storageAccountName)'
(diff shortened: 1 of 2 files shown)


eng/update-vmr.sh
```diff
@@ -0,0 +1,102 @@
...
+# replace the last two characters in sdk_version with xx
+branch_version=$(echo ${sdk_version} | sed 's/..$/xx/')
+target_branch="release/${branch_version}" # e.g. release/6.0.1xx
+
+rm -rf "${vmr_path}"
+git init "${vmr_path}"
+
+pushd "${vmr_path}"
+    git remote add upstream "${upstream_url}"
+    git fetch upstream "${target_branch}"
```
MichaelSimons, line 85: Wouldn't this benefit by including `--depth 1`

eng/source-build-pre-release.yml
```diff
@@ -67,9 +71,37 @@ jobs:
...
+      
+      upstream_with_pat=$(echo $(vmrUpstreamUrl) | sed "s,https://.*@,https://dn-bot:${AZDO_PAT}@,g")
+
+      $(Build.SourcesDirectory)/eng/update-vmr.sh \
+        --releaseChannel $(ReleaseChannel) \
+        --sdkVersion $(SdkVersion) \
+        --upstream ${upstream_with_pat} \
+        --tarball $(Pipeline.Workspace)/dotnet-sdk-source-$(SdkVersion).tar.gz
+
+      month_year=$(date +"%b%Y" -d "+1 month" | sed 's/.*/\L&/') # e.g. aug2022
```
MichaelSimons, line 93: There seems to be some coupling and duplication between this pipeline logic and the script.  e.g. the branch name algorithm is duplicated and the dotnet-vmr path is hard coded across the two.  These can cause maintenance issues in the future if they ever need to be changed.  Consider pushing all of the logic into the script and make this step a simple invocation of the script.  This would solve the coupling as well as make it easier to test locally.  If beneficial for local testing purposes, the push logic could be optional controlled by a parameter.

eng/update-vmr.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+set -euo pipefail
+
+time="$(date +%m%d%y-%H%M)"
```
MichaelSimons, line 4: Is this used?

eng/update-vmr.sh
```diff
@@ -0,0 +1,102 @@
...
+   echo
+   echo "Syntax: ./update-vmr.sh --releaseChannel 7.0 --sdkVersion 7.0.100 --upstream <valid git url> --tarball dotnet-sdk-source.tar.gz"
+   echo
+   echo "options:"
+   echo "--releaseChannel, -r  The .NET SDK release channel (e.g. 6.0)"
+   echo "--sdkVersion, -v      The .NET SDK version (e.g. 6.0.110)"
+   echo "--upstream, -u        A valid git URL to the upstream repo to base the update branch off of"
+   echo "--tarball, -t         The tarball to update the VMR with"
+   echo "--isDryRun, -d        If set, then don't push results to AzDo"
+   echo "--help, -h            (Optional) print this help message and exit"
```
MichaelSimons, line 18: What makes this optional and the rest required?  I'm not seeing any enforcement of the required options.  DryRun shouldn't be required but the rest should.

MichaelSimons (approved):

PR #3132, opened 2022-11-28
[main] Add release info as json to announcement discussion posts

lbussell:
Fixes https://github.com/dotnet/source-build/issues/2888

### DIFF
--- a/eng/source-build-release-announcement.md
+++ b/eng/source-build-release-announcement.md
@@ -9,11 +9,11 @@ Please use the [$TAG tag]($TAG_URL) to source-build .NET version $RUNTIME_VERSIO
 
 @dotnet/distro-maintainers
 
-<!--
+```json
 {
   "Channel": "$RELEASE_CHANNEL",
   "Runtime": "$RUNTIME_VERSION",
   "Sdk": "$SDK_VERSION",
   "InstallerTag": "$TAG"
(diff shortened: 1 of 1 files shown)


crummel (approved):
Does this need any kind of header or something to help people pull the JSON out?  If it's doable without that it looks good to me.

eng/source-build-release-announcement.md
```diff
@@ -7,4 +7,13 @@ Please use the [$TAG tag]($TAG_URL) to source-build .NET version $RUNTIME_VERSIO
 - Release Notes: $RELEASE_NOTES_URL
 - Tag URL: $TAG_URL
 
-@dotnet/distro-maintainers
\ No newline at end of file
+@dotnet/distro-maintainers
+
+<!--
+{
```
MichaelSimons, line 13: I agree with @crummel that a better indication of the machine readable section should be considered.  Avoiding a comment block would make this more visible and promote usage.  e.g. ```json would be better IMO.

lbussell: I see, would we rather have the info be visible in the announcement? The original suggestion was explicitly for a comment.

I can put together an example of how to pull the info from the discussion announcement.

MichaelSimons (approved):

Issue #3184, opened 2022-12-15
Rename src/SourceBuild/tarball directory in dotnet/installer


MichaelSimons:
The src/SourceBuild/tarball directory name no longer makes sense with the introduction of the VMR (Virtual Mono Repo).  This directory should be renamed.

A couple suggestions that I have heard are:

src/VirtualMonoRepo/
src/SourceBuild/

Within this folder there would be content/, patches/ and readme.md

This issue was also discussed in https://github.com/dotnet/source-build/issues/2867#issuecomment-1350697418.

MichaelSimons:
[Triage] We discussed this and feel that src/SourceBuild is the better name.

premun:
I found this reference that might be worth fixing: https://github.com/dotnet/source-build/blob/466a89576b148ea5d540c5ccfafbf9ffbe2d76c3/eng/submit-source-build-release-pr.sh#L36

MichaelSimons:
@premun - is this work done?  Can this be closed?

PR #3189, opened 2022-12-21
Remove a reference to the tarball directory

premun:
https://github.com/dotnet/source-build/issues/3184

### DIFF
--- a/eng/submit-source-build-release-pr.sh
+++ b/eng/submit-source-build-release-pr.sh
@@ -33,7 +33,7 @@ OPTS=$(getopt --options $SHORT --long $LONG --name "$0" -- "$@")
 if [ $? != 0 ] ; then echo "Failed to parse options." >&2 ; exit 1 ; fi
 eval set -- "$OPTS"
 
-global_json_path='src/SourceBuild/tarball/content/global.json'
+global_json_path='src/SourceBuild/content/global.json'
 versions_props_path='eng/Versions.props'
 custom_target_branch=''
 setup_git_auth=''


eng/submit-source-build-release-pr.sh
```diff
@@ -33,7 +33,7 @@ OPTS=$(getopt --options $SHORT --long $LONG --name "$0" -- "$@")
 if [ $? != 0 ] ; then echo "Failed to parse options." >&2 ; exit 1 ; fi
 eval set -- "$OPTS"
 
-global_json_path='src/SourceBuild/tarball/content/global.json'
+global_json_path='src/SourceBuild/content/global.json'
```
lbussell, line 36: We use this pipeline for 6.0 and 7.0 release processes, which still has this path.

Maybe we need to fork `dotnet/source-build` so that we can have updated release pipelines for the VMR process. @MichaelSimons thoughts?

lbussell (changes requested):

eng/submit-source-build-release-pr.sh
premun: Do you always use this script from main or can you maybe run it from the 7.0 branch of source-build?

MichaelSimons: I personally like the idea of maintaining a single copy of the engineering infra we utilize for releases if it reduces the maintenance burden.  If conditioning the logic per branch is not feasible then we should branch.  I think it would be confusing however if we use main for .NET 8.0 and then release/7.0 for both 7.0 and 6.0.

lbussell: @premun could you condition this later in the script on the value of `sdkVersion`?

premun:
I will take care of this as part of https://github.com/dotnet/source-build/issues/3194

PR #3235, opened 2023-02-07
Add GatherPackages to grab smoke-test prereqs.

crummel:
This adds the GatherPackages.csproj used to get the smoke-test prereqs - I thought having this centralized and working for all versions was worthwhile.  I will add instructions for use to the checklist when I make that PR.


MichaelSimons (approved):
I know you mention updating the readme separately because you are making more edits, but IMO it would make sense to update the project reference with this PR so it is "transactional".

Issue #3194, opened 2023-01-03
Ensure release automation for security-partners works in .NET 8.0/VMR


MichaelSimons:
The ~~[recent change to introduce CI](https://github.com/dotnet/installer/pull/15007)~~ [recent pipeline added to mirror changes to the security partners repo](https://github.com/dotnet/source-build/blob/main/eng/source-build-release-mirror.yml) needs to be updates to work in the VMR.  There may be other changes needed with release automation - e.g. https://github.com/dotnet/source-build/blob/main/eng/source-build-pre-release.yml

premun:
I saw the triage recording but I don't quite understand this task. The [linked pipeline](https://github.com/dotnet/installer/blob/release/7.0.1xx/src/SourceBuild/tarball/content/eng/pipelines/security-partners-dotnet.yml) seems to only build the tarball and I thought we have agreed that we don't need that as the VMR is validated in different places.

Or maybe I misunderstand what "release automation" means? Is this issue about mirroring the VMR too?

premun:
I have also seen this one - https://github.com/dotnet/source-build/pull/3187 - should have this one been linked instead?

premun:
Or is this about making sure the whole release checklist is applicable for the new VMR?

MichaelSimons:
@premun - I link the wrong PR in the original description.  I updated the description.  I mean to link to the mirror pipeline.

premun:
I think I have the code ready but to run it, I'd need to wait for the branch snap and some `release/8.0` build

https://github.com/dotnet/source-build/compare/prvysoky/net-8.0?expand=1

premun:
I opened a PR to bring 8.0 up to par with 6.0/7.0


### Next steps
We could make the pipeline multi-stage and compose it of the various pipelines that need to be invoked separately. The stages could be divided by approval gates. So something like:
- **Get info stage** - Current pre-release pipeline
- **Approval gate** - Confirmation that smoke test prereqs are in place
- **Mirror stage** - Current mirroring pipeline
- **Approval gate** - Wait for .NET 8.0 release
- **Release stage** - Current release pipeline

We could then have 1 pipeline run for each release of each version (so a 6.0 run, 7.0 run, 8.0 run).

[...]

PR #3243, opened 2023-02-13
Make the SB mirroring pipeline work for 8.0 preview 1

premun:
There were a couple of minor things such as it could only push to an already existing branch.

I was able to mirror a dev branch: https://dev.azure.com/dnceng/internal/_build/results?buildId=2112291&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e&t=bf6cf4cf-6432-59cf-d384-6b3bcf32ede2

https://github.com/dotnet/source-build/issues/3194

### DIFF
--- a/eng/source-build-pre-release-8.0.yml
+++ b/eng/source-build-pre-release-8.0.yml
@@ -8,7 +8,7 @@ pool:
 resources:
   pipelines:
   - pipeline: dotnet-staging-pipeline-resource
-    source: Stage-DotNet-Test
+    source: Stage-DotNet
 
 parameters:
 - name: useSpecificPipelineRunIDs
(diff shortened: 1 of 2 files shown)


MichaelSimons:
@premun - Can you clarify the intent with this change?  Your description leads me to think the pipeline was updated to create the branch if it doesn't exist.  I am not spotting the change that fixed this, rather it looks like a better UX was implemented to call out the error.

MichaelSimons:
> @MichaelSimons the `|| echo Branch doesn't exist yet` will return 0 and not fail

Oh I get it, the change is hiding the error so that `git push` will create it.  Thanks for helping me understand.

MichaelSimons (approved):

Issue #3272, opened 2023-02-23
Release pipeline job consolidation / speed improvement


premun:
The release pipeline runs for quite a long time due to needing to wait for an agent for each job. There is some room to consolidate jobs so that we cut down the wait time a bit. This will be helpful during re-runs on the release day.

Issue #3271, opened 2023-02-23
Release pipeline should push the release tag into `dotnet/dotnet`


premun:
The release tag (example: https://github.com/dotnet/dotnet/releases/tag/v8.0.100-preview.1) was created and pushed manually during the latest release but this should happen automatically from the pipeline. Ideally, the tag would be signed (verified).

⚠️ For internal flows, this should probably also push the commits that might not be public yet.

Issue #3269, opened 2023-02-23
Release pipeline should deduce or require release tag


premun:
### Context

Currently, the release tag is an optional parameter (SDK version is used otherwise) and there are some problems with this:
- It needs to be set for preview releases
- It is error prone (we already made a mistake during the first release).

### Goals

- Figure out if the release tag value can be deduced from the release manifest and use that.
- Make it required.
- Potentially log it somewhere visible (possibly with other parameters) so that it can be double checked in later stages.

PR #3276, opened 2023-02-27
Automate creation of GitHub tag & release 

premun:
There are many improvements in this PR to the SB release pipeline:
- Release tag now comes from staging pipeline information so preview releases work without a custom tag (https://github.com/dotnet/source-build/issues/3269)
- It is possible to specify a human readable name (e.g. ".NET 8 Preview 1") that is used during the release
    - Used in announcement title
    - Used in release title
- Added automation for creation of tag+release in the dotnet/dotnet repo (https://github.com/dotnet/source-build/issues/3271)
    - We use `pack-sources.sh` to append sources to the release (temporary fix for https://github.com/dotnet/source-build/issues/3265)
    - Release can be created as a draft - the tag is then only published once we publish the draft
    - Example release: https://github.com/dotnet/dotnet/releases/tag/untagged-fb19defbe3cfd793befd
- Fix handling of specific associated installer/VMR builds to now correctly find the SHA it belongs to - the SHA is then used for the GitHub release
- Release stage jobs are consolidated into one to save time (https://github.com/dotnet/source-build/issues/3272)

Example dry-run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2123565
Example release: https://github.com/dotnet/dotnet/releases/tag/untagged-fb19defbe3cfd793befd (unpublished)

### DIFF
--- a/.github/ISSUE_TEMPLATE/releases/release_checklist.md
+++ b/.github/ISSUE_TEMPLATE/releases/release_checklist.md
@@ -58,10 +58,7 @@
            - [ ] Approve the `Approval - Partner notification` approval stage.
       1. - [ ] `Approval - Release` stage
            - [ ] **SYNC POINT**: Wait for Microsoft build release.
-           - [ ] ⚠️ 8.0: In case you decided to not create the GitHub release (you set the pipeline parameter to `skip`), push the release tag to `dotnet/dotnet`
-                - The tag name should match the custom tag used when staging the release pipeline.
-                - The tag should be available in the `dnceng/dotnet-security-partners` repository.
-                - The commit to which this tag belongs is shown in the `Pre-Release` stage (in the `Get Associated Pipeline Run IDs` step).
+           - [ ] ⚠️ 8.0: Verify the release tag in the `dnceng/dotnet-security-partners` repository matches theVMR commit to which is shown in the `Pre-Release` stage (in the `Get Associated Pipeline Run IDs` step).
            - [ ] Approve the `Approval - Release` approval stage.
       1. - [ ] `Release` stage
            - [ ] Verify that the announcement was posted to [dotnet/source-build discussions](https://github.com/dotnet/source-build/discussions) and that the content is correct and all links work.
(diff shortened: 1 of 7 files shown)


eng/templates/stages/release.yml
```diff
@@ -71,243 +106,242 @@ stages:
...
+          if [ ${{ parameters.isPreviewRelease }} = True ]; then
+            prerelease='--prerelease'
+          fi
+
+          draft=''
+          if [ ${{ parameters.createGitHubRelease }} = draft ]; then
+            draft='--draft'
+          fi
+
+          gh release create "$(releaseTag)" "$(SourceTarballPath)#Packed sources" \
```
MichaelSimons, line 150: IMO we should be using the standard GH name here and give an indication it is a tarball.

```suggestion
          gh release create "$(releaseTag)" "$(SourceTarballPath)#Source code (tar.gz)" \
```

.github/ISSUE_TEMPLATE/releases/release_checklist.md
```diff
@@ -58,7 +58,7 @@
            - [ ] Approve the `Approval - Partner notification` approval stage.
       1. - [ ] `Approval - Release` stage
            - [ ] **SYNC POINT**: Wait for Microsoft build release.
-           - [ ] ⚠️ 8.0: Push the release tag to `dotnet/dotnet`
+           - [ ] ⚠️ 8.0: In case you decided to not create the GitHub release (you set the pipeline parameter to `skip`), push the release tag to `dotnet/dotnet`
```
MichaelSimons, line 61: I lean towards removing this.  This is adding unnecessary complexity to the checklist.  If we did this for every optional piece of functionality, the checklist would be unwieldy.

eng/source-build-release-official.yml
```diff
@@ -42,21 +62,18 @@ parameters:
...
-  type: boolean
-  default: false
-- name: customTag
-  displayName: Custom release tag (e.g. v6.0.XYY-source-build)
+- name: createGitHubRelease
+  displayName: '[⚠️ 8.0] Create tag & release in dotnet/dotnet'
   type: string
-  default: ' '
+  values: [ 'skip', 'draft', 'full' ]
+  default: draft
```
MichaelSimons, line 76: Why draft instead of full?  Would it make sense to have a `default` option that is dynamic in nature?  If DryRun is used, the default would be to skip, if not is would default to full.

premun: That's a good idea actually. I named it `auto`. I wonder if I should document the behaviour somewhere else than the code?

.github/ISSUE_TEMPLATE/releases/release_checklist.md
premun: I kept there an instruction verifying whether the tag belongs to the right SHA.

PR #3301, opened 2023-03-04
[main] Use new CI leg name

lbussell:
I have updated the CI leg names in https://github.com/dotnet/installer/pull/15603.

There is still an open question of how we can make sure leg names are shared across these multiple pipelines.

I think that after .NET 8 is released, or whenever we are prebuilt-clean and will likely not require a re-bootstrap, we should get artifacts from the `CentOSStream8_Offline_PreviousSourceBuiltSdk` leg instead of the `CentOSStream8_Offline_MsftSdk` leg. This would prevent us from using any Microsoft-built binaries at any point past the initial bootstrap during the .NET 8.0 servicing.

### DIFF
--- a/eng/templates/stages/release.yml
+++ b/eng/templates/stages/release.yml
@@ -78,6 +78,7 @@ stages:
 
   - ${{ if eq(parameters.dotnetMajorVersion, '8.0') }}:
     - name: sourceBuildArtifactLeg
+      # When .NET 8.0 is released, gather artifacts from the `PreviousSourceBuiltSdk` leg instead
       value: CentOSStream8_Offline_MsftSdk_x64_Artifacts
     - name: sourceBuiltArtifactsFileName
       value: Private.SourceBuilt.Artifacts.8.0.100.centos.8-x64.tar.gz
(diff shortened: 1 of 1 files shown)


premun (approved):

eng/templates/stages/release.yml
```diff
@@ -78,7 +78,8 @@ stages:
 
   - ${{ if eq(parameters.dotnetMajorVersion, '8.0') }}:
     - name: sourceBuildArtifactLeg
-      value: CentOSStream8_Offline_x64_Artifacts
+      # When .NET 8.0 is released, gather artifacts from the `PreviousSourceBuiltSdk` leg instead
```
premun, line 81: Maybe make this a TODO and link to an issue?

MichaelSimons: I personally think an issue should be logged and the comment removed.

crummel (approved):

eng/templates/stages/release.yml
lbussell: Filed https://github.com/dotnet/source-build/issues/3304

PR #3311, opened 2023-03-13
Resolve a Secure Supply Chain issue

premun:
Adds a blank NuGet.config which resolves a problem that was flagged during today's release pipeline run

### DIFF
--- a/test/NuGet.config
+++ b/test/NuGet.config
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+  </packageSources>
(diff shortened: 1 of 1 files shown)


crummel (approved):

MichaelSimons:
Have you verified this does not impact the product build?  My concern is the impact to things like `eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat`.

MichaelSimons:
> Have you verified this does not impact the product build? My concern is the impact to things like `eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat`.

Nevermind, I see this is in the test folder.

PR #3319, opened 2023-03-15
Skip mirroring when the tag already exists

premun:
This will allow re-runs of the pipeline

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2135894&view=logs&j=b6422d7d-1f12-5a27-5bca-0651b2848f2b&t=32705529-1086-50d7-1e7e-88b31414415c&l=35

### DIFF
--- a/eng/templates/stages/mirror.yml
+++ b/eng/templates/stages/mirror.yml
@@ -18,7 +18,7 @@ parameters:
 
 stages:
 - stage: Mirror
-  displayName: Mirror to DSP
+  displayName: Mirror tags & packages
   dependsOn:
   - PreRelease
   - MirrorApproval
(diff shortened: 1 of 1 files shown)


eng/templates/stages/mirror.yml
```diff
@@ -18,14 +18,14 @@ parameters:
 
 stages:
 - stage: Mirror
-  displayName: Mirror to DSP
+  displayName: Mirror tags & packages
```
MichaelSimons, line 21: I realize this is literally mirroring tags but I think this would be more understandable at the stage and job level to use `source` instead of `tags`

MichaelSimons (approved):

PR #3321, opened 2023-03-15
Use runtime version as the default release tag

premun:
https://github.com/dotnet/source-build/issues/3317

### DIFF
--- a/eng/templates/steps/initialize-release-info.yml
+++ b/eng/templates/steps/initialize-release-info.yml
@@ -75,7 +75,7 @@ steps:
       tag="${{ parameters.customTag }}"
       echo "Using custom tag $tag"
     else
-      tag="v$release"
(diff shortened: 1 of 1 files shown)


premun:
@crummel I believe this should fix the issues with 6.0/7.0 we've seen yesterday

premun:
Oh, but this will make the preview release tag to be `8.0.0-preview.2.23128.3`.

I guess for previews we only want to use the release name? Or do we want to start tagging by the runtime name already @MichaelSimons (based on the discussion in #3317)?

MichaelSimons:
> I guess for previews we only want to use the release name? Or do we want to start tagging by the runtime name already @MichaelSimons (based on the discussion in #3317)?

I am in favor of using the runtime name (release name).

MichaelSimons (approved):

PR #3331, opened 2023-03-17
Use runtime version as the release tag for 6.0/7.0

premun:
One final adjustment to match what was discussed in https://github.com/dotnet/source-build/issues/3317#issuecomment-1472501742

### DIFF
--- a/eng/templates/stages/mirror.yml
+++ b/eng/templates/stages/mirror.yml
@@ -1,10 +1,12 @@
 parameters:
 - name: dotnetMajorVersion
-  displayName: Major .NET version being released
   type: string
 - name: releaseBranchName
-  displayName: Release branch name (e.g. release/8.0.1xx-preview1)
   type: string
+- name: isPreviewRelease
+  type: boolean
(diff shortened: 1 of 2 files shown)


eng/templates/stages/mirror.yml
```diff
@@ -36,8 +31,8 @@ stages:
       value: dn-bot@microsoft.com
     - name: destinationUrl
       value: https://dotnet-security-partners@dev.azure.com/dotnet-security-partners/dotnet/_git/dotnet
-    - name: sdkVersion
```
MichaelSimons, line 39: If I understand things correctly, this is changing the tagging for 6.0/7.0 as well.  I think we should retain using the sdk version for the servicing branches.

premun: I changed it so that non-custom-tagged non-preview releases keep the SDK version. But I am not sure what behaviour should we have for 8.0 servicing branches. Should this only happen for 6.0/7.0 then?

MichaelSimons (approved):

PR #3332, opened 2023-03-17
Make SB release PR against main always

premun:
https://github.com/dotnet/installer/pull/15796

### DIFF
--- a/eng/templates/stages/release.yml
+++ b/eng/templates/stages/release.yml
@@ -353,7 +353,7 @@ stages:
             extraArgs+=("--versionsProps" "eng/Versions.props")
           fi
 
-          if [ ${{ parameters.isPreviewRelease }} = True ]; then
+          if [ "${{ parameters.isPreviewRelease }}" = True ]; then
             target_branch='main'
           else
             target_branch="${{ parameters.releaseBranchName }}"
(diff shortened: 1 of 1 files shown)


eng/templates/stages/release.yml
```diff
@@ -367,7 +362,7 @@ stages:
             echo "  --sdkVersion $(sdkVersion)"
             echo "  --title $title"
             echo "  --body $body"
-            echo "  --targetBranch $target_branch"
+            echo "  --targetBranch main"
```
MichaelSimons, line 365: This doesn't look right.  "main" should only be used for preview releases.  The PRs for servicing releases (e.g. 6.0/7.0) should still me made against the servicing branches.

premun: Ah, you're right. Too of a hasty change. Will fix

mthalman (commented):
The [submit script](https://github.com/dotnet/source-build/blob/main/eng/submit-source-build-release-pr.sh ) also needs to be updated to account for the changes in Versions.props from https://github.com/dotnet/installer/pull/15603. That would be specific to the target branch though.

premun:
@mthalman thanks - I wanted to make that change separately as I need more time to read into it

MichaelSimons (approved):

PR #3355, opened 2023-03-31
Remove custom tarball attachment from VMR releases

premun:
https://github.com/dotnet/source-build/issues/3265


premun:
This should get only merged after Preview 3 is out

MichaelSimons (approved):

PR #3396, opened 2023-04-18
Make release JSON metadata not installer-specific

### LINKED ISSUE #3380: Release pipeline: JSON in release announcement not relevant to VMR
When the release pipeline generates a release announcement, it includes JSON to describe state of the release. One of the fields is `InstallerTag`. This refers to the https://github.com/dotnet/installer repo which is relevant for 8.0 releases whose source is in https://github.com/dotnet/dotnet.

Personally, I think the name of the field shouldn't be repo-specific. I think something like this would be more appropriate:

```json
"RepoTag": "https://github.com/dotnet/dotnet/releases/tag/v8.0.100-preview.3.23178.7"
```

This same field name could then be used for 6.0/7.0 servicing builds with a field value that references `installer` instead.

premun:
#3380

### DIFF
--- a/eng/source-build-release-announcement.md
+++ b/eng/source-build-release-announcement.md
@@ -14,7 +14,6 @@ Please use the [$TAG tag]($TAG_URL) to source-build .NET version $RUNTIME_VERSIO
   "Channel": "$RELEASE_CHANNEL",
   "Runtime": "$RUNTIME_VERSION",
   "Sdk": "$SDK_VERSION",
-  "Tag": "$TAG",
   "TagUrl": "$TAG_URL"
 }
 ```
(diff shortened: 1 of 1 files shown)


eng/source-build-release-announcement.md
```diff
@@ -14,6 +14,7 @@ Please use the [$TAG tag]($TAG_URL) to source-build .NET version $RUNTIME_VERSIO
   "Channel": "$RELEASE_CHANNEL",
   "Runtime": "$RUNTIME_VERSION",
   "Sdk": "$SDK_VERSION",
-  "InstallerTag": "$TAG"
+  "Tag": "$TAG",
```
mthalman, line 17: I think the `Tag` field can be excluded. The `TagUrl` field you've included should be sufficient.

eng/source-build-release-announcement.md
```diff
@@ -14,6 +14,7 @@ Please use the [$TAG tag]($TAG_URL) to source-build .NET version $RUNTIME_VERSIO
...
+  "TagUrl": "$TAG_URL"
```
premun, line 18:
```suggestion
  "TagUrl": "$TAG_URL"
```

eng/source-build-release-announcement.md
premun: Great!

mthalman (approved):

Issue #3352, opened 2023-03-30
Pre-release stage should include link to generated PR


mthalman:
The [`Update security-partners-dotnet`](https://github.com/dotnet/source-build/blob/4cd2965e63aa2d23112c4bb03282a4e905977a6f/eng/templates/stages/pre-release.yml#L114-L132) task generates a PR with the changes for the release. It would be helpful if it also included a link to the PR in the output to easily navigate to the PR (and cross-reference for troubleshooting purposes).

premun:
FYI @MichaelSimons, @mthalman we will ask @ali-turan7 to help us with this and onboard him onto the project

ali-turan7:
The response to the request for the establishment of a PR is recorded and parsed. The PR URL is built and printed using the information obtained.

PR #3428, opened 2023-05-03
Dev/ali turan/pr url3352

ali-turan7:
The PR request result has been recorded and parsed. The PR URL was built and printed using the information obtained.
https://github.com/dotnet/source-build/issues/3352

### DIFF
--- a/eng/push-tarball.sh
+++ b/eng/push-tarball.sh
@@ -135,7 +135,7 @@ pushd "$vmr_path"
     if [ -n "$webUrl" ] && [ -n "$pullRequestId" ]; then
       echo "$webUrl"/pullrequest/"$pullRequestId"
     elif [[ "$curlResult" == *"TF401179"* ]]; then
-      echo "##vso[task.logissue type=error]$(jq -r '.message' <<< "$curlResult")"
+      "$(jq -r '.message' <<< "$curlResult")"
     else
       echo "##vso[task.logissue type=error]An unexpected error has occurred during the pr request!"
       jq <<< "$curlResult"
--- a/eng/push-tarball.sh
+++ b/eng/push-tarball.sh
@@ -129,6 +129,16 @@ pushd "$vmr_path"
     git push -u upstream "$new_branch_name"
 
     echo "Creating PR from $new_branch_name to $target_branch"
-    curl -H 'Content-Type: application/json' -d "$data" "$pr_url"
+    curlResult=$(curl -H 'Content-Type: application/json' -d "$data" "$pr_url" 2>/dev/null)
+    webUrl=$(jq -r '.repository.webUrl | values' <<< "$curlResult")
+    pullRequestId=$(jq -r '.pullRequestId | values' <<< "$curlResult")
+    if [ -n "$webUrl" ] && [ -n "$pullRequestId" ]; then
+      echo "$webUrl"/pullrequest/"$pullRequestId"
+    elif [[ "$curlResult" == *"TF401179"* ]]; then
+      echo "##vso[task.logissue type=error]$(jq -r '.message' <<< "$curlResult")"
+    else
+      echo "##vso[task.logissue type=error]An unexpected error has occurred during the pr request!"
+      jq <<< "$curlResult"
+    fi
   fi
 popd


eng/push-tarball.sh
```diff
@@ -129,6 +129,16 @@ pushd "$vmr_path"
...
-    curl -H 'Content-Type: application/json' -d "$data" "$pr_url"
+    curlResult=$(curl -H 'Content-Type: application/json' -d "$data" "$pr_url" 2>/dev/null)
+    webUrl=$(jq -r '.repository.webUrl | values' <<< "$curlResult")
+    pullRequestId=$(jq -r '.pullRequestId | values' <<< "$curlResult")
+    if [ -n "$webUrl" ] && [ -n "$pullRequestId" ]; then
+      echo "$webUrl"/pullrequest/"$pullRequestId"
+    elif [[ "$curlResult" == *"TF401179"* ]]; then
+      "$(jq -r '.message' <<< "$curlResult")"
+    else
+      echo "##vso[task.logissue type=error]An unexpected error has occurred during the pr request!"
```
premun, line 140:
```suggestion
      echo "##vso[task.logissue type=error]An unexpected error has occurred while creating the security partners PR!"
```

eng/push-tarball.sh
```diff
@@ -129,6 +129,16 @@ pushd "$vmr_path"
     git push -u upstream "$new_branch_name"
 
     echo "Creating PR from $new_branch_name to $target_branch"
-    curl -H 'Content-Type: application/json' -d "$data" "$pr_url"
+    curlResult=$(curl -H 'Content-Type: application/json' -d "$data" "$pr_url" 2>/dev/null)
```
premun, line 132: I believe we use `snake_case` in bas (and in the rest of this file) for variable names

eng/push-tarball.sh
```diff
@@ -129,6 +129,16 @@ pushd "$vmr_path"
...
+    webUrl=$(jq -r '.repository.webUrl | values' <<< "$curlResult")
+    pullRequestId=$(jq -r '.pullRequestId | values' <<< "$curlResult")
+    if [ -n "$webUrl" ] && [ -n "$pullRequestId" ]; then
+      echo "$webUrl"/pullrequest/"$pullRequestId"
+    elif [[ "$curlResult" == *"TF401179"* ]]; then
+      "$(jq -r '.message' <<< "$curlResult")"
```
premun, line 138:
```suggestion
      jq -r '.message' <<< "$curlResult"
```

premun, line 138: What does the message say in this case?

Should we just print a warning instead that the release PR already exists? You can print warning the same way the error is printed

eng/push-tarball.sh
tkapin: Since we are redirecting STDERR to /dev/null and are not checking the exit code, how do we learn about an error? IS there reason to discard the STDERR?

eng/push-tarball.sh
```diff
@@ -129,6 +129,16 @@ pushd "$vmr_path"
     git push -u upstream "$new_branch_name"
 
     echo "Creating PR from $new_branch_name to $target_branch"
-    curl -H 'Content-Type: application/json' -d "$data" "$pr_url"
+    curlResult=$(curl -H 'Content-Type: application/json' -d "$data" "$pr_url" 2>/dev/null)
+    webUrl=$(jq -r '.repository.webUrl | values' <<< "$curlResult")
+    pullRequestId=$(jq -r '.pullRequestId | values' <<< "$curlResult")
+    if [ -n "$webUrl" ] && [ -n "$pullRequestId" ]; then
+      echo "$webUrl"/pullrequest/"$pullRequestId"
+    elif [[ "$curlResult" == *"TF401179"* ]]; then
```
tkapin, line 137: 1) Where does this magical constant come from? It would be good to include a brief explanation in a comment so that readers of this code understand where does this come from.

2) Using the glob here is a bit strange as you are efficiently matching the string. Recommend to use the =~ operator for regexp match instead, i.e., `[[ "$curl_result" =~ "TF401179" ]]`

eng/push-tarball.sh
tkapin: Can you think of including some useful information or details in the error (e.g., the url or the PR id?)

eng/push-tarball.sh
ali-turan7: This the message:
"TF401179: An active pull request for the source and target branch already exists."
Sure, I will print the same message as you mentioned.

eng/push-tarball.sh
ali-turan7: I think the PR id and url are available after the PR is created successfully. Additionally, the data that was used to create the PR was previously printed. I am attempting to print the request result, which I presume has all the information, in the following line: But I am really open to suggestions.

eng/push-tarball.sh
ali-turan7: I had assumed that just the request response would be sufficient, but now I am including the errors as well. Thank you :)

eng/push-tarball.sh
premun: The whole response is printed below so I'd just leave it at that.

eng/push-tarball.sh
```diff
@@ -129,6 +129,18 @@ pushd "$vmr_path"
...
-    curl -H 'Content-Type: application/json' -d "$data" "$pr_url"
+    pr_creation_response=$(curl -H 'Content-Type: application/json' -d "$data" "$pr_url")
+    repository_web_url=$(jq -r '.repository.webUrl | values' <<< "$pr_creation_response")
+    pull_request_id=$(jq -r '.pull_request_id | values' <<< "$pr_creation_response")
+    
+    if [ -n "$repository_web_url" ] && [ -n "$pull_request_id" ]; then
+      echo "$repository_web_url/pullrequest/$pull_request_id"
+    # The TF401179 error code indicates that there is already an open pull request for the source and target branches.   
+    elif [[ "$pr_creation_response" =~ "TF401179" ]]; then
+      echo "An active pull request for the source and target branch already exists."
```
premun, line 140: Please make this a warning

premun (approved):
Before we merge, could you maybe make a change in a separate branch that would execute this PR logic even on a dry run and run a test pipeline to see if it works?

ali-turan7:
> Before we merge, could you maybe make a change in a separate branch that would execute this PR logic even on a dry run and run a test pipeline to see if it works?
> 
> Ideally then re-run the stage to see if an already opened PR is detected

Done :)

premun (approved):
I think we can merge then!

premun (approved):

Issue #3353, opened 2023-03-30
Skip creation of PR in pre-release stage if it contains no commits


mthalman:
If a release needs to be iterated upon due to issues that are encountered, it may be necessary to run the Pre-release stage multiple times. In that case, it's possible the PR that gets generated by [`Update security-partners-dotnet`](https://github.com/dotnet/source-build/blob/4cd2965e63aa2d23112c4bb03282a4e905977a6f/eng/templates/stages/pre-release.yml#L114-L132) has already been merged. So when a subsequent run of the Pre-release stage occurs, it will still create the PR even though it contains no commits. I suggest that the PR creation be skipped if it contains no commits, and include a warning message instead.

premun:
FYI @MichaelSimons, @mthalman we will ask @ali-turan7 to help us with this and onboard him onto the project

ali-turan7:
The necessary control is applied by comparing the new and target branches to determine whether a new commit has been made as well as whether a new PR is necessary.

PR #3431, opened 2023-05-05
Skipping PR creation in the absence of any commits

ali-turan7:
It is conceivable that the PR created by Update security-partners-dotnet has already been merged if Pre-release stage is executed more than once. 

Issue Link: https://github.com/dotnet/source-build/issues/3353

Pipeline Link: https://dev.azure.com/dnceng/internal/_build/results?buildId=2174527&view=logs&j=3a04dffd-22cf-5707-fa41-9bd4708cc1db&t=43d9a424-b64c-5d30-c3bd-0fca912bd01a

### DIFF
--- a/eng/push-tarball.sh
+++ b/eng/push-tarball.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -xeuo pipefail
+set -euo pipefail
 
 print-help ()
 {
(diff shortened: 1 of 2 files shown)


eng/push-tarball.sh
```diff
@@ -107,7 +107,12 @@ pushd "$vmr_path"
 
   # Re-runs should expect the branch being merged previously already
```
premun, line 108: Does the comment still apply?

eng/push-tarball.sh
```diff
@@ -107,7 +107,12 @@ pushd "$vmr_path"
...
   git add -f .
-  git diff --staged --quiet || git commit -m "Update to .NET $sdk_version"
+  if git diff --staged --quiet; then
```
premun, line 110: I think this might be doing something else than we want.

- There's the `release/7.0` branch
- We create a new branch called something like `release/7.0-may-2023` above
- We try to open PR from the `may` branch to the `release/7.0` one

If we run this multiple times but not merge the PR, or abandon the previous PR, the new branch might be up to date already but the PR might not be there.

I think we should rather compare the new and the target branch after we commit to see if they differ and only then skip.

premun (commented):
I think this might not be the intended behaviour

premun (commented):
Please update the PR title

eng/push-tarball.sh
ali-turan7: Got the point, thank you :)

premun (approved):

Issue #3423, opened 2023-04-26
Release pipeline: `mono` executable error when mirroring NuGet packages


mthalman:
This error is showing up in the `Publish packages` task:

```
##[error]Error: Unable to locate executable file: 'mono'. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
```

[Build link](https://dev.azure.com/dnceng/internal/_build/results?buildId=2167811&view=logs&j=b6422d7d-1f12-5a27-5bca-0651b2848f2b&t=3ad0151c-6c6b-5312-5ed6-a86e6bc28020) (internal)

Presumably, this is because of the [change in the agent type](https://github.com/dotnet/source-build/pull/3414) which doesn't include the required `mono` executable.

Additionally, I don't think this task should be configured to continue on error. If it fails, we want the pipeline to have an error.

mthalman:
Just installing mono isn't enough. It requires other packages as well. Otherwise, leads to this error:

```
System.IO.FileNotFoundException: Could not load file or assembly 'System.ComponentModel.Composition, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' or one of its dependencies.
File name: 'System.ComponentModel.Composition, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
  at NuGet.CommandLine.Program.MainCore (System.String workingDirectory, System.String[] args) [0x000e3] in <dc1b484fc03b4c9e94db6d2032a7a13d>:0 
```

premun:
@mthalman have you also tried to call the `NuGet authenticate` step? I believe that one gives the build access to feeds in the same AzDO organization.

https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-authenticate-v1?view=azure-pipelines

mthalman:
> @mthalman have you also tried to call the `NuGet authenticate` step? I believe that one gives the build access to feeds in the same AzDO organization.

I tried that but it didn't help.

mthalman:
According to the service connection description, it uses the `dn-bot-all-orgs-artifact-feeds-rw` secret. @premun - Can you confirm this is not expired?

premun:
Will do. Btw @mthalman we've talked with Tomas that we might give Ali the task of onboarding source-build onto secret manager. Might be worth since this is a second time there's a token problem.

premun:
[...]

PR #3441, opened 2023-05-12
Install NuGet and its dependencies in mirror stage of release pipeline

mthalman:
Fixes https://github.com/dotnet/source-build/issues/3423

### DIFF
--- a/eng/templates/stages/mirror.yml
+++ b/eng/templates/stages/mirror.yml
@@ -163,6 +163,10 @@ stages:
       - ${{ else }}:
         - script: sudo apt update && DEBIAN_FRONTEND=noninteractive sudo apt install -y apt-transport-https mono-complete nuget
           displayName: Install NuGet and dependencies
+        - task: NuGetAuthenticate@1
+          displayName: Authenticate to ${{ variables.destinationPackageFeed }}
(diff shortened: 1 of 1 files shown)


MichaelSimons (approved):

premun:
I am not sure if we want mono - isn't it better to install .NET 7 and use `dotnet nuget ...`?

premun:
Again, maybe Arcade here would help

mthalman:
> I am not sure if we want mono - isn't it better to install .NET 7 and use `dotnet nuget ...`?

I've tried that but haven't been unable to get it authenticated to the feed. And even then, I was only able to get this to work by adding the `NuGetAuthenticate` task.

premun:
So .NET 7 SDK + NuGet Authenticate wouldn't work?

That's so weird.. I imagine that's what was on the `1es-ubuntu` image

mthalman:
> So .NET 7 SDK + NuGet Authenticate wouldn't work?
> 
> That's so weird.. I imagine that's what was on the `1es-ubuntu` image

Nope, that didn't work in my testing. I've tried all sorts of combinations and these changes are the only thing that worked.

Issue #3268, opened 2023-02-22
Build tarball artifacts are not versioned correctly


MichaelSimons:
The following build artifacts are not versioned correctly:

1.  Private.SourceBuilt.Artifacts
1.  Private.SourceBuilt.Prebuilts
1.  dotnet-smoke-test-prereqs

If you look at a Preview 1 build these are versioned as 8.0.100. I was the version to indicate they are for a Preview 1 build.  Furthermore when looking at these artifacts in CI, I would expect their version number to be build specific.  Having the version numbers include the build would help in traceability as it would identify which build produced them.  This would improve the dev workflow for uploading these build artifacts to blob storage as it would eliminate the need to rename them to the 0.1.0-8.0.100-n.centos.8-x64 syntax.

NikolaMilosavljevic:
We could easily fix the naming of `Private.SourceBuilt.Artifacts`, to become something like `Private.SourceBuilt.Artifacts.8.0.100-preview.3.23178.1.centos.8-x64.tar.gz`.

What is the expectation for naming of `Private.SourceBuilt.Prebuilts` and `dotnet-smoke-test-prereqs`?

All of these artifacts are produced in separate infra projects, so the solutions would vary, as well as the cost for fixing.

@MichaelSimons

MichaelSimons:
All should be versioned the same.

PR #3451, opened 2023-05-16
Use wildcard for referencing artifacts tarball

mthalman:
Now that https://github.com/dotnet/source-build/issues/3268 is fixed, it causes a mismatch in the expected artifacts tarball file name that is used for the release pipeline. The file name now contains a build number (e.g. `Private.SourceBuilt.Artifacts.8.0.100-preview.4.23260.1.centos.8-x64.tar.gz`) but the release pipeline is expecting `Private.SourceBuilt.Artifacts.8.0.100.centos.8-x64.tar.gz`.

This is fixed by using a wildcard for the version.


MichaelSimons (approved):

premun (approved):

premun:
@ali-turan7 fyi this changed happen, you might want to sync your main

Issue #3447, opened 2023-05-16
Release pipeline: Use of announcement gist results in `TAG_URL` variable not being set


mthalman:
The release pipeline allows you the option of referencing a [gist URL](https://github.com/dotnet/source-build/blob/d4f497340a69a09d6cb58b7fd8eba5a8cc27c74e/eng/source-build-release-official.yml#L72-L75) that will be used as the content for the release announcement instead of automatically generating it. When that parameter is provided, it results in the following error when executing the [`Submit announcement discussion` task](https://github.com/dotnet/source-build/blob/d4f497340a69a09d6cb58b7fd8eba5a8cc27c74e/eng/templates/stages/release.yml#L249-L317) in the release stage:

```
line 62: TAG_URL: unbound variable
```

This is because that variable only gets set as a result of running the [`create-announcement-draft.sh` script](https://github.com/dotnet/source-build/blob/d4f497340a69a09d6cb58b7fd8eba5a8cc27c74e/eng/create-announcement-draft.sh). But that script doesn't get run when an announcement gist URL is provided.

The same applies for the use of the `RELEASE_NOTES_URL` variable.

ali-turan7:
TAG_URL and RELEASE_NOTES_URL variables are removed from being printed in release.yml

PR #3473, opened 2023-05-23
TAG_URL and RELEASE_NOTES_URL variables are removed from being printed in release.yml

ali-turan7:
### Related issue:
https://github.com/dotnet/source-build/issues/3447

### Description
TAG_URL and RELEASE_NOTES_URL variables are removed from being printed in release.yml

### DIFF
--- a/eng/create-announcement-draft.sh
+++ b/eng/create-announcement-draft.sh
@@ -120,7 +120,4 @@ else
   export RUNTIME_VERSION_SHORT="$RUNTIME_VERSION"
 fi
 
-echo "Tag URL: $TAG_URL"
-echo "Release Notes URL: $RELEASE_NOTES_URL"
-
 envsubst < "$template"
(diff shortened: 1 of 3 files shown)


eng/create-announcement-draft.sh
```diff
@@ -120,4 +120,7 @@ else
   export RUNTIME_VERSION_SHORT="$RUNTIME_VERSION"
 fi
 
+echo "Tag URL: $TAG_URL"
```
premun, line 123: We can't do this - look at how the script is used - we use the output of this script for the announcement so these lines would appear in the official announcement.

premun (changes requested):

eng/create-announcement-draft.sh
mthalman: > Maybe we just check if `TAG_URL` is set?

I'm fine with that.

premun: We have looked at it in detail again and it seems this would only get printed when we use gist and we're doing 6.0/7.0. But we only use gists for 8.0 at this time so this would never get printed.

I don't like how it's just relying on the env variables so we decided to just remove these lines as the tag is linked in the announcement itself anyway

premun (approved):

premun (approved):
Please rename the PR accordingly

PR #3477, opened 2023-05-25
Do not publish draft VMR release on dry runs

premun:
There were 35 draft releases published in the repo. I think we should skip on dry runs and if you want to test what the release in `dotnet/dotnet` looks like, you can run a dry run and set the setting to `draft`

Additionally, the GitHub page is buggy so it was hard to delete the draft releases

### DIFF
--- a/eng/templates/stages/release.yml
+++ b/eng/templates/stages/release.yml
@@ -86,8 +86,7 @@ stages:
     - name: sdkArtifactFileName
       value: dotnet-sdk-*.tar.gz
 
-  # GitHub release is skipped for 6.0/7.0
-  # For dry run, we only create a draft release
+  # GitHub release is skipped for 6.0/7.0 and dry runs
   - name: createGitHubRelease
     ${{ if eq(parameters.createGitHubRelease, 'auto') }}:
       ${{ if or(eq(parameters.isDryRun, parameters.dotnetMajorVersion, '6.0'), eq(parameters.dotnetMajorVersion, '7.0')) }}:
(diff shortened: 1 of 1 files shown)


premun:
@mthalman I wanted to get your opinion on this one

ali-turan7 (approved):

eng/templates/stages/release.yml
```diff
@@ -86,17 +86,13 @@ stages:
     - name: sdkArtifactFileName
       value: dotnet-sdk-*.tar.gz
 
-  # GitHub release is skipped for 6.0/7.0
-  # For dry run, we only create a draft release
+  # GitHub release is skipped for 6.0/7.0 and dry runs
   - name: createGitHubRelease
     ${{ if eq(parameters.createGitHubRelease, 'auto') }}:
-      ${{ if or(eq(parameters.dotnetMajorVersion, '6.0'), eq(parameters.dotnetMajorVersion, '7.0')) }}:
+      ${{ if or(eq(parameters.isDryRun, parameters.dotnetMajorVersion, '6.0'), eq(parameters.dotnetMajorVersion, '7.0')) }}:
```
mthalman, line 92: I don't think this syntax is correct unless I'm missing something. It looks like three arguments are being passed to the `eq` function:

```
eq(parameters.isDryRun, parameters.dotnetMajorVersion, '6.0')
```

premun: Oops, should have been this:

```suggestion
      ${{ if or(parameters.isDryRun, eq(parameters.dotnetMajorVersion, '6.0'), eq(parameters.dotnetMajorVersion, '7.0')) }}:
```

mthalman (approved):

Issue #3386, opened 2023-04-13
Provide a mechanism to gather prereq packages required to run smoke-tests offline


MichaelSimons:
The smoke-tests do not support running offline as they have a set of prereq packages.  The source-build infrastructure should provide a mechanism to gather these packages so that the tests can be run offline.  

The prep script script feels like the natural place for this to happen as it's primary responsibility is to gather the prereqs necessary to build.

mthalman:
During development of https://github.com/dotnet/installer/pull/16359, it became clear that versions of dependent packages are being updated frequently enough that makes the maintenance of a baseline file unsustainable.

For that reason, a different approach needs to be taken. Rather than relying on the prep script that would need to be run to gather the smoke test prereqs, the proposal is that a tarball containining these prereqs is provided as part of the release assets. This tarball would contain all the necessary packages for building and running the smoke tests offline. It would need to consolidate all of the restored packages from the various architectures of build legs since there are architecture-specific packages included in the mix.

mthalman:
@omajid / @mirespace - Please read the new approach I've described in https://github.com/dotnet/source-build/issues/3386#issuecomment-1583120095 and let me know your thoughts. Feel free to respond here if further clarification is needed.

Fundamentally, this isn't much different than what we do today for partner hand-off. This will just be automated as part of our release process. And the tarball that is produced will only include packages that are consumed by the smoke tests. Previously, we had been including some [additional packages](https://github.com/dotnet/source-build/blob/b9be5fbf2a037a002713bcb4ffec6527e1520425/test/GatherPackages.csproj#L13C1-L21) that weren't necessarily consumed by our smoke tests; those will no longer be included if the tests don't need them.

dviererbe:
Hi @mthalman 👋🏻, we (Canonical) currently do not run the included smoke-tests automatically and as a consequence do not use the smoke test prereqs tarball either. But getting them working is on our Roadmap; I will put this issue on my TODO for next week.

The approach you proposed sounds good to me 👍🏻, if it should be more convenient for MS, I can also try Red Hat's approach and consume the NuGet feed directly.

mthalman:
Use of the NuGet feed is more tricky because it requires knowing which packages the tests require. More packages exist in the feed than are actually used by the tests. That's one of the reasons why we're going with this tarball approach.

Issue #3497, opened 2023-06-06
Release pipeline: Source tarball should not be uploaded in Pre-Release stage


premun:
### Context

For .NET 6.0/7.0, we upload the `.tar.gz` of sources to a storage account. This should happen in the mirroring stage rather than the pre-release stage.

### Goal

We should try to figure out why we do it like this and if there is no reason for it, move it into the mirroring stage.

MichaelSimons:
[Triage] This is required for the testability.  We prefer to not make this before 8.0 GA in the public pipeline for stability reasons therefore this should be made in the new internal pipeline only.

oleksandr-didyk:
Completed as part of https://github.com/dotnet/arcade-services/issues/2871

PR #3512, opened 2023-06-15
Upload smoke test prereqs tarball

mthalman:
Updates the release pipeline to generate the smoke test prereqs tarball and upload it to blob storage for use by .NET security partners. This is implemented by combining the individual smoke test prereqs tarballs from an x64 and arm64 build leg of the source-build pipeline.

This does not implement logic to provide this tarball for public use. That's in the scope of https://github.com/dotnet/source-build/issues/3015.

Fixes https://github.com/dotnet/source-build/issues/3386

### DIFF
--- a/eng/combine-tarballs.sh
+++ b/eng/combine-tarballs.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+### Usage: combine-tarballs.sh [OPTIONS] [file1] [file2] ... [fileN]
+###
+### Combines the contents of multiple tarballs into a single tarball.
+###
+### Options:
+###   --output <FILE>  The destination path of the output tarball
+###   --help, -h       (Optional) Print this help message and exit
+
+set -euo pipefail
+source="${BASH_SOURCE[0]}"
+
+function print_help () {
+  sed -n '/^### /,/^$/p' "$source" | cut -b 5-
+}
+
+outputPath=''
+
+if [[ $# -gt 0 ]]; then
+  opt="$(echo "$1" | tr "[:upper:]" "[:lower:]")"
+  case "$opt" in
+    -h | --help )
+      print_help
+      exit 0
+      ;;
+    --output )
+      outputPath="$2"
+      shift 2
+      ;;
+    *)
+    ;;
+  esac
+else
+  print_help
+  exit 0
+fi
+
+: "${outputPath:?Missing --output}"
+
+tmpDir=$(mktemp -d)
+
+# The rest of the args are the paths to the input tarball
+for arg in "$@"
+do
+  echo "Extracting '${arg}' to '${tmpDir}'"
+  tar -xf "${arg}" -C $tmpDir
+done
+
+echo "Generating tarball '${outputPath}' from '${tmpDir}'"
+tar -cf "${outputPath}" -C $tmpDir .
+
+rm -rf $tmpDir
--- a/eng/combine-tarballs.sh
+++ b/eng/combine-tarballs.sh
@@ -48,6 +48,6 @@ do
 done
 
 echo "Generating tarball '${outputPath}' from '${tmpDir}'"
-tar -cf "${outputPath}" -C $tmpDir .
+tar --numeric-owner -cf "${outputPath}" -C $tmpDir .
 
 rm -rf $tmpDir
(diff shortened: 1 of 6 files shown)


eng/combine-tarballs.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# The rest of the args are the paths to the input tarball
+for arg in "$@"
+do
+  echo "Extracting '${arg}' to '${tmpDir}'"
+  tar -xf "${arg}" -C $tmpDir
+done
+
+echo "Generating tarball '${outputPath}' from '${tmpDir}'"
+tar -cf "${outputPath}" -C $tmpDir .
```
MichaelSimons, line 51: Should `--numeric-owner` be specified?

eng/templates/stages/mirror.yml
MichaelSimons, line 1: We have reached the point where this stage should be renamed IMO.  It is doing much more than mirroring which hurts the readability of the pipeline.  This pipeline feels more like the pre-release stage to me while the current pre-release is prep-release or init-release.  That will be more so once https://github.com/dotnet/source-build/issues/3497 is resolved.

If you think this is beyond the scope of your changes, and you agree, let's make sure an issue gets logged (I can do it if you don't want to).

mthalman: I've logged https://github.com/dotnet/source-build/issues/3532 for this.

mthalman:
This is dependent on a build job name change from https://github.com/dotnet/installer/pull/16756.

eng/templates/stages/mirror.yml
```diff
@@ -48,17 +52,49 @@ stages:
...
+    - name: runtimeVersion
+      value: $[ stageDependencies.PreRelease.PreRelease.outputs['ReadReleaseInfo.RuntimeVersion'] ]
+    - name: releaseChannel
+      value: $[ stageDependencies.PreRelease.PreRelease.outputs['ReadReleaseInfo.ReleaseChannel'] ]
+    - name: combinedSmokeTestPrereqsTarballPath
+      value: $(Pipeline.Workspace)/dotnet-smoke-test-prereqs-$(sdkVersion).tar.gz
+
+    - ${{ if eq(parameters.dotnetMajorVersion, '6.0') }}:
+      - name: smokeTestPreReqsArtifactNameX64
+        value: Build_Tarball_x64 CentOSStream9-Offline_Artifacts
```
MichaelSimons, line 66: This is going to create a point of maintenance throughout the lifecycle of updating/retiring the distros/versions our CI runs on.  What is the failure experience if this leg no longer exists?  Because the main CI pipeline is defined elsewhere there is no easy way to keep these in sync.

mthalman: If it didn't exist, you'd get an error like `Artifact <name> was not found for build <id>`. If you run a dry run ahead of time (which I do), you'll catch this at that time.

Other options:
* Update the VMR/tarball pipeline to have an additional stage that consolidates the smoke test prereqs from the build legs and outputs them to a well-known artifact that can be consume from the release pipeline.
* Update the VMR/tarball pipeline to publish well-known artifact names specifically for x64 and arm64 from the individual build legs (no additional stage). For example, `CentOSStream9_Offline_MsftSdk_x64` would publish a `SmokeTestPrereqs-x64` artifact and `Debian11_Offline_MsftSdk_arm64` would publish a `SmokeTestPrereqs-arm64` artifact. The release pipeline would consume these as hard-coded artifacts but with more stable names than the build legs. The key is to configure the pipeline so that one and only one build leg for each architecture is the one that publishes the prereqs artifact. That adds some maintenance burden in case of removal/addition of build legs.

From a simplicity standpoint, I consider what I've currently implemented to be the simplest. Yes, it can be broken but the churn on build legs isn't likely to be high and can be caught with dry run. And of course, the other options can cause breakage too if not maintained correctly.

Let me know your thoughts.

MichaelSimons: I have no objections to the proposal.  The CI legs change infrequently and the error message is pretty clear.

MichaelSimons (approved):

mthalman:
Holding off on merging this. It requires https://github.com/dotnet/installer/pull/16855 to be merged and also to flow to release/7.0.1xx branch.

PR #3603, opened 2023-08-29
Remove obsolete copy of release checklist

NikolaMilosavljevic:
We are not updating this copy of the checklist, anymore. Removing it to avoid confusion.


mmitche (approved):

MichaelSimons (approved):
+1 on moving this internal and keeping only one copy.

Issue #3637, opened 2023-09-12
VMR commits are not mirrored on the release day


NikolaMilosavljevic:
We need to mirror all internal VMR commits to GitHub repo, on the release day.

premun:
Thanks @missymessa, moved to `dotnet/source-build`

oleksandr-didyk:
Have a working idea how to deal with the mirroring, need to test it out and create a step in the SB release pipeline for it. Should be complete before 09/28

PR #3638, opened 2023-09-26
Tag build commit before releasing

oleksandr-didyk:
Resolves https://github.com/dotnet/source-build/issues/3637

Tag VMR commit of the build being released and push the said commit before creating a GitHub release.

Even if the commit is coming from the `internal` VMR, pushing the tag would just result in dangling commits at the time of release until the `internal` branch is merged back. Example of a tag with 2 commits from `internal` pushed via a tag - https://github.com/oleksandr-didyk/installer/releases/tag/test-mirror

GitHub CLI can use an existing tag for a release ([command docs](https://cli.github.com/manual/gh_release_create)). Added `--verify-tag` as a safe-guard to make sure the tag is re-used and the command does not create a new one accidentally. Example of a release from a tag mentioned above - https://github.com/oleksandr-didyk/installer/releases

Tested in an AzDO pipeline:
  - if tag exists -> [run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2277880&view=logs&j=b6422d7d-1f12-5a27-5bca-0651b2848f2b&t=cd58d141-4f7b-58b2-52d5-fbefaad1d9e4)
  - if tag exists but links to a different commit -> [run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2277882&view=logs&j=b6422d7d-1f12-5a27-5bca-0651b2848f2b&t=cd58d141-4f7b-58b2-52d5-fbefaad1d9e4)
  - if tag does not exist and `git push` for the new tag fails -> [run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2277910&view=logs&j=b6422d7d-1f12-5a27-5bca-0651b2848f2b&t=cd58d141-4f7b-58b2-52d5-fbefaad1d9e4)

### DIFF
--- a/eng/templates/stages/release.yml
+++ b/eng/templates/stages/release.yml
@@ -158,18 +158,25 @@ stages:
           git config --global user.name "$git_user"
           git config --global user.email "$git_email"
 
-          git tag $releaseTag $dotnetDotnetCommit
-          
           vmr_url=$(echo $vmrPublicUrl | sed "s,https://.*@,https://$(git_user):${GH_TOKEN}@,g")
-          result=$(git push $vmr_url refs/tags/$releaseTag 2>&1)
-          
-          if [[ $? != 0 ]]; then
-            echo "##vso[task.logissue type=error]Failed to create a GitHub tag for the release: $result"
-            exit 1
-          fi
 
-          if [[ "$result" =~ "Everything up-to-date" ]]; then
-            echo "##vso[task.logissue type=warning]GitHub release tag $(releaseTag) already exists ($(vmrPublicUrl)/releases/tag/$(releaseTag))"
+          git fetch $vmr_url tag $releaseTag --no-tags
+          if [[ $? = 0 ]]; then
+            tagged_commit=$(git rev-list -1 refs/tags/$releaseTag)
+            
+            if [[ $tagged_commit != $dotnetDotnetCommit ]]; then
+              echo "##vso[task.logissue type=error]The release tag '$releaseTag' is linked to a different commit. Expected: '$dotnetDotnetCommit'; Actual: '$tagged_commit'"
+              echo "##vso[task.logissue type=error]See the tag at $vmrPublicUrl/releases/tag/$releaseTag for more info."
+              exit 1
+            fi
+          else
+            git tag $releaseTag $dotnetDotnetCommit
+            git push $vmr_url refs/tags/$releaseTag 2>&1
+          
+            if [[ $? != 0 ]]; then
+              echo "##vso[task.logissue type=error]Failed to create a GitHub tag for the release: $result"
+              exit 1
+            fi
           fi
 
           release_json="$(Build.ArtifactStagingDirectory)/release.json"
--- a/eng/templates/stages/release.yml
+++ b/eng/templates/stages/release.yml
@@ -45,6 +45,8 @@ stages:
     value: dotnet
   - name: announcementRepo
     value: source-build
+  - name: publicVmrUrl
+    value: https://github.com/dotnet/dotnet
   - name: storageAccountName
     value: dotnetcli
   - name: blobContainerName
@@ -148,6 +150,18 @@ stages:
             draft='--draft'
           fi
 
+          git tag $releaseTag $dotnetDotnetCommit
+          
+          result=$(git push $publicVmrUrl refs/tags/$releaseTag 2>&1)
+          if [[ $? -ne 0 ]]; then
+            echo "##vso[task.logissue type=error]Failed to create a GitHub tag for the release: $result"
+            exit 1
+          fi
+
+          if [[ "$result" =~ "Everything up-to-date" ]]; then
+            echo "##vso[task.logissue type=warning]GitHub release tag $(releaseTag) already exists ($(publicVmrUrl)/releases/tag/$(releaseTag))"
+          fi
+
           release_json="$(Build.ArtifactStagingDirectory)/release.json"
 
           echo "{}" \
@@ -156,7 +170,7 @@ stages:
             | jq '. += { "tag": "$(releaseTag)" }'                                 \
             | jq '. += { "sdkVersion": "$(sdkVersion)" }'                          \
             | jq '. += { "runtimeVersion": "$(runtimeVersion)" }'                  \
-            | jq '. += { "sourceRepository": "https://github.com/dotnet/dotnet" }' \
+            | jq '. += { "sourceRepository": "$(publicVmrUrl)" }' \
             | jq '. += { "sourceVersion": "$(dotnetDotnetCommit)" }'               \
             > "$release_json"
 
@@ -187,19 +201,15 @@ stages:
             "$release_json#Release manifest"         \
             --repo dotnet/dotnet                     \
             --title "${{ parameters.releaseName }}"  \
-            --target "$(dotnetDotnetCommit)"         \
             --notes-file "$release_notes"            \
+            --verify-tag                             \
             $prerelease                              \
             $draft \
             2>&1)
 
           if [ $? -ne 0 ]; then
-            if [[ "$result" =~ "already exists" ]]; then
-              echo "##vso[task.logissue type=warning]GitHub release tag $(releaseTag) already exists (https://github.com/dotnet/dotnet/releases/tag/$(releaseTag))"
-            else
-              echo "##vso[task.logissue type=error]Failed to create the GitHub release: $result"
-              exit 1
-            fi
+            echo "##vso[task.logissue type=error]Failed to create the GitHub release: $result"
+            exit 1
           fi
 
         displayName: Create GitHub release
--- a/eng/templates/stages/release.yml
+++ b/eng/templates/stages/release.yml
@@ -160,18 +160,17 @@ stages:
 
           vmr_url=$(echo $vmrPublicUrl | sed "s,https://.*@,https://$(git_user):${GH_TOKEN}@,g")
 
-          git fetch $vmr_url tag $releaseTag --no-tags
+          git fetch "$vmr_url" tag "$releaseTag" --no-tags
           if [[ $? = 0 ]]; then
-            tagged_commit=$(git rev-list -1 refs/tags/$releaseTag)
-            
-            if [[ $tagged_commit != $dotnetDotnetCommit ]]; then
+            tagged_commit=$(git rev-list -1 "refs/tags/$releaseTag")           
+            if [[ "$tagged_commit" != "$dotnetDotnetCommit" ]]; then
               echo "##vso[task.logissue type=error]The release tag '$releaseTag' is linked to a different commit. Expected: '$dotnetDotnetCommit'; Actual: '$tagged_commit'"
               echo "##vso[task.logissue type=error]See the tag at $vmrPublicUrl/releases/tag/$releaseTag for more info."
               exit 1
             fi
           else
-            git tag $releaseTag $dotnetDotnetCommit
-            git push $vmr_url refs/tags/$releaseTag 2>&1
+            git tag "$releaseTag" "$dotnetDotnetCommit"
+            git push "$vmr_url" "refs/tags/$releaseTag" 2>&1
           
             if [[ $? != 0 ]]; then
               echo "##vso[task.logissue type=error]Failed to create a GitHub tag for the release: $result"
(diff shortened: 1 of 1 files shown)


eng/templates/stages/release.yml
```diff
@@ -45,6 +45,8 @@ stages:
     value: dotnet
   - name: announcementRepo
     value: source-build
+  - name: publicVmrUrl
+    value: https://github.com/dotnet/dotnet
```
mthalman, line 49: `dotnet/dotnet` is already used here: https://github.com/dotnet/source-build/pull/3638/files#diff-16da42ff0ed618ecf203753bec7e4ebd9c976c28ff0753b0807d1d890d3a05cdR202

I would suggest defining a new variable for `dotnet/dotnet` and reference it from there. Then define another derived variable that adds the `https://github.com/` part.

eng/templates/stages/release.yml
```diff
@@ -148,6 +150,18 @@ stages:
             draft='--draft'
           fi
 
+          git tag $releaseTag $dotnetDotnetCommit
+          
+          result=$(git push $publicVmrUrl refs/tags/$releaseTag 2>&1)
```
mthalman, line 155: How does authentication work here? I see that `GH_TOKEN` is set but is that actually consumed by `git`? I know that `gh` uses it but I'm not sure about `git`.

eng/templates/stages/release.yml
```diff
@@ -148,6 +150,18 @@ stages:
...
+          if [[ $? -ne 0 ]]; then
```
mthalman, line 156: This `-ne` syntax doesn't work with the double brackets.

```suggestion
          if [[ $? != 0 ]]; then
```

eng/templates/stages/release.yml
oleksandr-didyk: Thanks for the suggestion, will improve

eng/templates/stages/release.yml
oleksandr-didyk: I forgot about it honestly, thanks for pointing it out. Will fix

oleksandr-didyk: Added the auth for the push

eng/templates/stages/release.yml
```diff
@@ -148,6 +152,26 @@ stages:
...
+          vmr_url=$(echo $vmrPublicUrl | sed "s,https://.*@,https://$(git_user):${GH_TOKEN}@,g")
+          result=$(git push $vmr_url refs/tags/$releaseTag 2>&1)
+          
+          if [[ $? != 0 ]]; then
+            echo "##vso[task.logissue type=error]Failed to create a GitHub tag for the release: $result"
+            exit 1
+          fi
+
+          if [[ "$result" =~ "Everything up-to-date" ]]; then
+            echo "##vso[task.logissue type=warning]GitHub release tag $(releaseTag) already exists ($(vmrPublicUrl)/releases/tag/$(releaseTag))"
```
premun, line 172: I would not warn about this. Instead, we could verify if the tag points to the same commit. If not, we should throw an error saying that the tag already exists and points to a different SHA.
Otherwise just log that we are good (this is a re-run and we already pushed the tag)

oleksandr-didyk: I thought about this originally but could something like this happen? The tag in question points to a runtime version being released, so we would need to have two different commits of the VMR getting released with the same runtime version

premun: We can get additional commits for SB only that we would be released on top of what we release usually. This happened recently:
https://github.com/dotnet/installer/tree/release/8.0.1xx-preview3-sb

This is also when we then override the build IDs to configure what differs when we're releasing.

That being said, regardless of this flow, if it ever happens that we want to tag something and a tag already points to something else, we should definitely fail hard because this is unusual and we need to know about this.
So this check makes sense to me to have even if it looks like dead code.

oleksandr-didyk: Sure thing, sounds reasonable, will add. Thanks for pointing it out

oleksandr-didyk: Added the check

eng/templates/stages/release.yml
```diff
@@ -148,6 +152,33 @@ stages:
...
 
+          git_user=dotnet-bot
+          git_email=dotnet-bot@microsoft.com
+
+          git config --global user.name "$git_user"
+          git config --global user.email "$git_email"
+
+          vmr_url=$(echo $vmrPublicUrl | sed "s,https://.*@,https://$(git_user):${GH_TOKEN}@,g")
+
+          git fetch $vmr_url tag $releaseTag --no-tags
```
premun, line 163:
```suggestion
          git fetch "$vmr_url" tag "$releaseTag" --no-tags
```

premun, line 163: Curious, what does the `--no-tags` do?

eng/templates/stages/release.yml
```diff
@@ -148,6 +152,33 @@ stages:
...
+          if [[ $? = 0 ]]; then
+            tagged_commit=$(git rev-list -1 refs/tags/$releaseTag)
```
premun, line 165: Why not this?
```suggestion
            tagged_commit=$(git rev-parse "refs/tags/$releaseTag")
```

eng/templates/stages/release.yml
```diff
@@ -148,6 +152,33 @@ stages:
...
+
```
premun, line 166:
```suggestion

```

eng/templates/stages/release.yml
```diff
@@ -148,6 +152,33 @@ stages:
...
+          git config --global user.name "$git_user"
+          git config --global user.email "$git_email"
+
+          vmr_url=$(echo $vmrPublicUrl | sed "s,https://.*@,https://$(git_user):${GH_TOKEN}@,g")
+
+          git fetch $vmr_url tag $releaseTag --no-tags
+          if [[ $? = 0 ]]; then
+            tagged_commit=$(git rev-list -1 refs/tags/$releaseTag)
+            
+            if [[ $tagged_commit != $dotnetDotnetCommit ]]; then
```
premun, line 167:
```suggestion
            if [[ "$tagged_commit" != "$dotnetDotnetCommit" ]]; then
```

eng/templates/stages/release.yml
premun, line 163: Have you tested this? Seems like this command will kill the execution because we have `set -euo pipefail` so it will never reach the if?

eng/templates/stages/release.yml
```diff
@@ -148,6 +152,33 @@ stages:
...
+              echo "##vso[task.logissue type=error]The release tag '$releaseTag' is linked to a different commit. Expected: '$dotnetDotnetCommit'; Actual: '$tagged_commit'"
+              echo "##vso[task.logissue type=error]See the tag at $vmrPublicUrl/releases/tag/$releaseTag for more info."
+              exit 1
+            fi
+          else
+            git tag $releaseTag $dotnetDotnetCommit
+            git push $vmr_url refs/tags/$releaseTag 2>&1
+
```
premun, line 175:
```suggestion

```

eng/templates/stages/release.yml
```diff
@@ -148,6 +152,33 @@ stages:
...
+            tagged_commit=$(git rev-list -1 refs/tags/$releaseTag)
+            
+            if [[ $tagged_commit != $dotnetDotnetCommit ]]; then
+              echo "##vso[task.logissue type=error]The release tag '$releaseTag' is linked to a different commit. Expected: '$dotnetDotnetCommit'; Actual: '$tagged_commit'"
+              echo "##vso[task.logissue type=error]See the tag at $vmrPublicUrl/releases/tag/$releaseTag for more info."
+              exit 1
+            fi
+          else
+            git tag $releaseTag $dotnetDotnetCommit
+            git push $vmr_url refs/tags/$releaseTag 2>&1
```
premun, line 174: Please add an `echo` that we are pushing this. Or does git show that?

eng/templates/stages/release.yml
```diff
@@ -148,6 +152,33 @@ stages:
...
+          if [[ $? = 0 ]]; then
+            tagged_commit=$(git rev-list -1 refs/tags/$releaseTag)
+            
+            if [[ $tagged_commit != $dotnetDotnetCommit ]]; then
+              echo "##vso[task.logissue type=error]The release tag '$releaseTag' is linked to a different commit. Expected: '$dotnetDotnetCommit'; Actual: '$tagged_commit'"
+              echo "##vso[task.logissue type=error]See the tag at $vmrPublicUrl/releases/tag/$releaseTag for more info."
+              exit 1
+            fi
+          else
+            git tag $releaseTag $dotnetDotnetCommit
```
premun, line 173:
```suggestion
            git tag "$releaseTag" "$dotnetDotnetCommit"
```

eng/templates/stages/release.yml
premun, line 174:
```suggestion
            git push "$vmr_url" "refs/tags/$releaseTag" 2>&1
```

premun, line 174: Again, `set -e` here will fail the script

eng/templates/stages/release.yml
```diff
@@ -185,21 +216,17 @@ stages:
           set +e
           result=$(gh release create "$(releaseTag)" \
             "$release_json#Release manifest"         \
-            --repo dotnet/dotnet                     \
+            --repo $vmrPublicRepo                     \
```
premun, line 219:
```suggestion
            --repo $vmrPublicRepo                    \
```

premun (commented):
I think you might need to test these paths as these bash changes are always tricky to get right

eng/templates/stages/release.yml
oleksandr-didyk: It stops `git` form pulling all tags. If I understood correctly it creates a local ref for the tag in question but pulls all associated tags (and lists them once the command is finished).

With this option set it will only pull the mentioned tag and will create a local ref for it

eng/templates/stages/release.yml
oleksandr-didyk: According to [this](https://stackoverflow.com/a/1862542) thread there is a small difference between the two for annotated tags as `rev-parse` shows the SHA1 of the tag and not the commit, leading to a potential diff depending on the type of tag used.

Probably not a big difference in our case, can use `rev-parse` if needed

eng/templates/stages/release.yml
oleksandr-didyk: I did, but without the `set -euo pipefail`. Thanks for pointing it out

eng/templates/stages/release.yml
oleksandr-didyk: I think these spaces make it a bit easier to read, but I'll remove them

eng/templates/stages/release.yml
oleksandr-didyk: It does, but I'll add the `echo`, thanks for suggesting it

eng/templates/stages/release.yml
premun: Ah, yeah this would bite us if ever created an annotated one. Thanks!

eng/templates/stages/release.yml
premun: Thanks

oleksandr-didyk: What would you suggest to do in this case actually? There is also a `set +x` below and I don't think wrapping the code with `set +e` and `set -e` is the way to go

premun: That should not stop the script

oleksandr-didyk: Did that + used `set +e` since there were other commands later that could fail the job

eng/templates/stages/release.yml
oleksandr-didyk: Thanks for pointing it out, added `set +e`

eng/templates/stages/release.yml
```diff
@@ -148,6 +152,32 @@ stages:
             draft='--draft'
           fi
 
+          git_user=dotnet-bot
+          git_email=dotnet-bot@microsoft.com
+
+          git config --global user.name "$git_user"
+          git config --global user.email "$git_email"
+
+          vmr_url=$(echo $vmrPublicUrl | sed "s,https://.*@,https://$(git_user):${GH_TOKEN}@,g")
```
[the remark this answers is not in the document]
oleksandr-didyk: Thanks, no idea why I did not simplify this instantly. 

As for the testing - ran the chunk I added in an AzDO pipeline, will edit the description of the PR with link to the builds

mthalman (approved):

PR #3716, opened 2023-11-03
Fix tag matching logic

mthalman:
When attempting to run the release pipeline for 8.0 GA, an error occurred when attempting to mirror the tag to the partners repo:

```
Tagging 40e7f014ff784457efffa58074549735e30772ae as v8.0.0
Tag v8.0.0 already exists, checking if it points to the same commit
Tag v8.0.0 already exists but does not point to a 40e7f014ff784457efffa58074549735e30772ae, aborting
```

This is because the logic which checks for an existing tag is not constrained enough:

https://github.com/dotnet/source-build/blob/198e13b4ad85563ffa2de1c79734501442257dde/eng/templates/stages/mirror.yml#L153

It simply greps on the tag name. But the repo already contains tags like `refs/tags/v8.0.0-preview.7.23375.6`. So it's matching on those and thinks the `v8.0.0` tag already exists when it does not.

I've updated the logic to match on the exact tag name.

### DIFF
--- a/eng/templates/stages/mirror.yml
+++ b/eng/templates/stages/mirror.yml
@@ -150,10 +150,23 @@ stages:
 
         # Check if the tag already exists and points to the same commit
         sha=$(git rev-parse HEAD)
-        if git ls-remote --tags destination | grep -q "$tag_name"; then
+
+        # When querying for tags we need to ensure match the whole line and not just a substring
+        # of the tag name. This avoids accidentally matching a new tag of v8.0.0 with an existing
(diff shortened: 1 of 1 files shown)


MichaelSimons (approved):

premun (approved):

mthalman:
@oleksandr-didyk - This will need to be ported to the other repo.

oleksandr-didyk:
> @oleksandr-didyk - This will need to be ported to the other repo.

PR - https://dev.azure.com/dnceng/internal/_git/dotnet-release/pullrequest/35064

PR #4123, opened 2024-02-13
Add package dependency flow documentation

MichaelSimons:
Is this documentation useful?  Is there something missing you would like to see added?  Please share your feedback.

### DIFF
--- a/Documentation/package-dependency-flow.md
+++ b/Documentation/package-dependency-flow.md
@@ -1,7 +1,7 @@
 # Package Dependency Flow
 
 This document describes how package dependencies are handled within source build.
-It described the mechanisms that exist to control which package versions are used.
+It describes the mechanisms that exist to control which package versions are used.
 
 ## Origins of Packages
 
@@ -70,7 +70,7 @@ properties for all non-pinned dependencies.
 
 **PackageVersions.Previous.props:** This will contain version properties with the
 package versions from the [previous release of source build](#previous-source-built-packages).
-If a new package exist that has never been release before, it will not have a version
+If a new package exists that has never been released before, it will not have a version
 property defined.
 
 ```xml
@@ -93,7 +93,7 @@ property defined.
 These two version.props files get imported by the arcade source build infrastructure after
 the repo's Version.props file. Therefore the repo's Versions.props property versions
 get overridden by the source build versions. In the case of the `SystemCommandLineVersion`
-example, the current source build version, 2.0.0-beta4, would win
+example, the current source build version, 2.0.0-beta4, would win. This is known as package version lifting since it lifts the originally defined package version to the current source built version. This behavior only applies to source build in the context of the [VMR](https://github.com/dotnet/dotnet) (also see [Repo Level Source Builds](#repo-level-source-builds).
 
 ### Transitive Version Properties
 
@@ -132,7 +132,7 @@ of the Versions.props file.
 
 ### Repo Level Source Builds
 
-The source build package lifting mechanism is not applicable when building a individual
+The source build package lifting mechanism is not applicable when building individual
 repos in source build mode because it doesn't have the context of the other product
 repos or previous source build release. In repo source build mode, the versions of the
 packages declared in the Versions.props are used.
--- a/Documentation/package-dependency-flow.md
+++ b/Documentation/package-dependency-flow.md
@@ -19,9 +19,11 @@ contains [tooling to generate new reference packages](https://github.com/dotnet/
 ### Current Source Built Packages
 
 This refers to all of the packages produced in the current build. The set of packages
-available to each repo varies based on its build order. If repo X has a dependency
-on package Y.1 but the repo Y hasn't been built yet, repo X cannot depend
-on the very latest version of the package.
+available to each repo varies based on its build order. For example the msbuild repo
+can take a dependency on the current version of Microsoft.CodeAnalysis from roslyn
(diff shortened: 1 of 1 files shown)


Documentation/package-dependency-flow.md
```diff
@@ -0,0 +1,140 @@
...
+[source-build-reference-packages](https://github.com/dotnet/source-build-reference-packages).
+This repo contains all of the reference packages used to build the product. The repo
+contains [tooling to generate new reference packages](https://github.com/dotnet/source-build-reference-packages?tab=readme-ov-file#adding-new-packages).
+
+### Current Source Built Packages
+
+This refers to all of the packages produced in the current build. The set of packages
+available to each repo varies based on its build order. If repo X has a dependency
+on package Y.1 but the repo Y hasn't been built yet, repo X cannot depend
+on the very latest version of the package.
```
mthalman, line 24: I think this would be better explained with concrete examples. Something like:

The msbuild repo can take a dependency on the current version of Microsoft.CodeAnalysis from roslyn because roslyn builds before msbuild. Conversely, since roslyn builds before msbuild, roslyn cannot take a dependency on the current version of Microsoft.Build; it can only take a dependency on the previously released version.

Documentation/package-dependency-flow.md
```diff
@@ -0,0 +1,140 @@
+# Package Dependency Flow
+
+This document describes how package dependencies are handled within source build.
+It described the mechanisms that exist to control which package versions are used.
```
mthalman, line 4:
```suggestion
It describes the mechanisms that exist to control which package versions are used.
```

Documentation/package-dependency-flow.md
```diff
@@ -0,0 +1,140 @@
...
+Because the .NET product uses itself to build, the .NET source build product must be
+[bootstrapped](./bootstrapping-guidelines.md). This process allows the packages from the
+previous source build release to be used to build the next version of the product. This
+provides a means for breaking the product's circular dependencies. For example repos like
+[arcade](https://github.com/dotnet/arcade) can self-reference its previous version to
+produce the next version.
+
+When referencing previous source built packages, it is important to not leak these
+previously built packages into the resulting packages/product. This is considered a
+"poison leak" and is not permitted during a source build as it breaks the notion of
```
mthalman, line 37: Add a link to poison info.

Documentation/package-dependency-flow.md
```diff
@@ -0,0 +1,140 @@
...
+...
+```
+
+During a source build, the infrastructure will scan the Version.Details.xml file
+and dynamically create two new Versions.props files containing updated version
+properties for all non-pinned dependencies.
+
+**PackageVersions.Previous.props:** This will contain version properties with the
+package versions from the [previous release of source build](#previous-source-built-packages).
+If a new package exist that has never been release before, it will not have a version
```
mthalman, line 73:
```suggestion
If a new package exists that has never been released before, it will not have a version
```

Documentation/package-dependency-flow.md
```diff
@@ -0,0 +1,140 @@
...
+```xml
+...
+  <SystemCommandLineVersion>2.0.0-beta4</SystemCommandLineVersion>
+...
+```
+
+These two version.props files get imported by the arcade source build infrastructure after
+the repo's Version.props file. Therefore the repo's Versions.props property versions
+get overridden by the source build versions. In the case of the `SystemCommandLineVersion`
+example, the current source build version, 2.0.0-beta4, would win
```
mthalman, line 96:
```suggestion
example, the current source build version, 2.0.0-beta4, would win. This is known as package version lifting since it lifts the originally defined package version to the current source built version. This behavior only applies to source build in the context of the [VMR](https://github.com/dotnet/dotnet) (also see [Repo Level Source Builds](#repo-level-source-builds).
```

Documentation/package-dependency-flow.md
```diff
@@ -0,0 +1,140 @@
...
+because of the property evaluation order. If the desired behavior is for
+`MicrosoftBuildVersion` to be set to the same value as `MicrosoftBuildFrameworkVersion`
+for source build, then you either need to declare the Microsoft.Build dependency
+in the Version.Details.xml file or move the `MicrosoftBuildVersion` assignment outside
+of the Versions.props file.
+
+
+### Repo Level Source Builds
+
+The source build package lifting mechanism is not applicable when building a individual
```
mthalman, line 135:
```suggestion
The source build package lifting mechanism is not applicable when building individual
```

Documentation/package-dependency-flow.md
MichaelSimons: Added.

Documentation/package-dependency-flow.md
MichaelSimons: Updated.

mthalman (approved):

Documentation/package-dependency-flow.md
```diff
@@ -43,32 +43,46 @@ ability to service the product.
...
-  </Dependency>
-...
-```
+1. A corresponding version property is defined in the Versions.props.
+
+    **Versions.props**
+
+    ```xml
+    ...
+      <SystemCommandLineVersion>2.0.0-beta4.24068.1</SystemCommandLineVersion>
```
mthalman, line 67: This property would also need to be referenced in order to have an effect.

MichaelSimons: Added clarity in the opening and closing sections of the _Package Versions_ section.

PR #4539, opened 2024-08-05
Add markdown linter GH action

MichaelSimons:
Fixes https://github.com/dotnet/source-build/issues/2855

1. Adds a markdown linter GH action that runs on PRs.  This action was copied from the [dotnet/runtime repo](https://github.com/dotnet/runtime/blob/main/.github/workflows/markdownlint.yml).  Here is a [sample PR](https://github.com/MichaelSimons/source-build/pull/1) demonstrating the usage.  If you look at the commit history, you will see how the errors are displayed in the source-view - [example](https://github.com/MichaelSimons/source-build/pull/1/commits/fe84b00abd1024f8641a5fb44870edf76ea1a2f7)
2. Fixes several markdown syntax violations

PR #4540, opened 2024-08-06
Reformat MD files to limit line lengths to 80 characters

MichaelSimons:
This is in response to https://github.com/dotnet/source-build/pull/4539/files#r1705996684.  I originally hoped to re-enable the MD013 rule but discovered it applies to code blocks as well.  There are several places where this produces undesirable results therefore I am leaving the rule disabled but still re-formatted the MD files to limit the long lines.  I used the 'Rewrap' vscode extension to reformat the files.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,5 +1,4 @@
 {
     "default": true,
-    "MD013": false, // 80 char line length
     "MD033": false, // no inline HTML
 }
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -2,7 +2,8 @@
     "default": true,
     // line length
     "MD013": {
-        "code_block_line_length": 256
+        "code_block_line_length": 256,
+        "heading_line_length": 96
     },
     "MD033": false, // no inline HTML
 }
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,4 +1,5 @@
 {
     "default": true,
+    "MD013": false, // 80 char line length - produces undesireable results within code blocks
     "MD033": false, // no inline HTML
 }
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,5 +1,8 @@
 {
     "default": true,
-    "MD013": false, // 80 char line length - produces undesireable results within code blocks
+    // line length
+    "MD013": {
+        "code_block_line_length": 256
+    },
     "MD033": false, // no inline HTML
 }
--- a/Documentation/bootstrapping-guidelines.md
+++ b/Documentation/bootstrapping-guidelines.md
@@ -1,32 +1,50 @@
 # Bootstrapping Guidelines to Build .NET from Source
 
-.NET utilizes itself to build therefore in order to build .NET from source, you first need to acquire or build a bootstrapping .NET SDK. This document provides guidance around acquiring and building this bootstrapping .NET SDK.
+.NET utilizes itself to build therefore in order to build .NET from source, you
+first need to acquire or build a bootstrapping .NET SDK. This document provides
+guidance around acquiring and building this bootstrapping .NET SDK.
 
-The version of the SDK used to source build .NET is referred to as "N-1" (e.g. 8.0.100). The version of the SDK produced by source build is referred to as "N" (e.g. 8.0.101). The previous SDK (e.g. N-1) supplies the tools required to build.
+The version of the SDK used to source build .NET is referred to as "N-1" (e.g.
+8.0.100). The version of the SDK produced by source build is referred to as "N"
+(e.g. 8.0.101). The previous SDK (e.g. N-1) supplies the tools required to
+build.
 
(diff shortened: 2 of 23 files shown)


Documentation/bootstrapping-guidelines.md
```diff
@@ -65,60 +85,106 @@ Building .NET for the first time is a two stage process:
...
+You will need to update the RID graph to include your new platform and runtime
+IDs.  See <https://github.com/dotnet/runtime/pull/82382> or
+<https://github.com/dotnet/runtime/pull/75396> for examples.
 
-Building for unsupported architectures require cross-compilaton on the supported platform. Determine the compatible host to build which provides cross-compilation toolchain.  [IBM has published](https://community.ibm.com/community/user/powerdeveloper/blogs/sapana-khemkar/2023/01/13/cross-build-dotnet7-on-x86-ibm-power?CommunityKey=8cc2a1f0-6307-48cb-9178-ace50920244e) a detailed description of how they successfully built .NET 7 for IBM Power.
+Building for unsupported architectures require cross-compilaton on the supported
+platform. Determine the compatible host to build which provides
+cross-compilation toolchain.  [IBM has
+published](https://community.ibm.com/community/user/powerdeveloper/blogs/sapana-khemkar/2023/01/13/cross-build-dotnet7-on-x86-ibm-power?CommunityKey=8cc2a1f0-6307-48cb-9178-ace50920244e)
```
ellahathaway, line 153: Does the line limit not apply here because this is a link?

ellahathaway (commented):
I'm a little disappointed that this rule can't be enabled. Do you have examples of code blocks which currently violate this rule? I'm wondering if we'd be able to adjust them to fit the line length or not.

Edit: https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md013---line-length seems to suggest that `code_blocks` can be excluded from the line length check, depending on which release we're using.

Documentation/bootstrapping-guidelines.md
MichaelSimons: Link urls are excluded from the limit.

MichaelSimons:
> Do you have examples of code blocks which currently violate this rule?

Yes, you can see the attempts in this PR - https://github.com/dotnet/source-build/actions/runs/10273766628/job/28428985773

A good example is the xml code snippet here - https://github.com/dotnet/source-build/blob/main/Documentation/poison-report-format.md#interpreting-the-poison-report.  We are intending to give a snippet of the poison report.  Reformatting it (specifically line 12) doesn't feel right to me.  Other places we have code blocks that includes commands to execute.  Breaking these apart across multiple will hurt the UX when they are intended to be copy/paste.  I could specify the code block line length to a large value so that we can have the rule enabled.  Thoughts?

ellahathaway:
> I could specify the code block line length to a large value so that we can have the rule enabled. Thoughts?

This sounds good to me!

.markdownlint.json
```diff
@@ -1,5 +1,9 @@
 {
     "default": true,
-    "MD013": false, // 80 char line length
+    // line length
+    "MD013": {
+        "code_block_line_length": 256,
+        "heading_line_length": 96
```
MichaelSimons, line 6: This was necessary for https://github.com/dotnet/source-build/blob/main/Documentation/planning/multi-sdk-band-support.md#a-distro-maintainer-wishing-to-build-the-initial-release-n00-of-a-nxx-band-sdk

ellahathaway (approved):

mthalman:
Is there precedence for 80 characters? It seems quite unnecessarily small.

MichaelSimons:
> Is there precedence for 80 characters? It seems quite unnecessarily small.

This is a fairly widely used practice.  I personally think it is becoming a bit dated.  From the internet:

> Rows longer than 80 characters should be split into multiple lines where appropriate. This helps make it easier for reviewers to make review comments and for the contributors to see where the requested changes are (without scrolling). It is also easier to see the lines in split view mode when the lines don’t wrap on the screen. It will also improve the readability when using git diff.

MichaelSimons:
I will also note all of the MD written by @dagood followed this practice already.

dagood:
FWIW, my new favorite Markdown style is to use one line per sentence. 😄 Very nice diffs for typical edits, usually short enough to not be all that annoying in the occasional non-wrapping viewer, and actually quite nice for the authoring/editing process, mechanically. Downside: less readable (IMO) before Markdown rendering is applied, which might actually influence how you write. (Still getting used to it.)

A random other thing that bugs me about 80 columns is this:

```
Refer to the [build
instructions](https://github.com/dotnet/installer/blob/main/README.md#build-net-from-source-source-build)
to review how to build the .NET SDK from source.
```

I think it's signficantly more readable to chop, but I don't think I've seen an editor/wrapper that does it this way:

```
Refer to the
[build instructions](https://github.com/dotnet/installer/blob/main/README.md#build-net-from-source-source-build)
to review how to build the .NET SDK from source.
```

MichaelSimons:
> my new favorite Markdown style is to use one line per sentence.

I've seen this.  I have to say I am not the biggest fan. 😄 

I agree the way links are handled by tooling is generally not the way I would prefer.

PR #4567, opened 2024-08-20
Switch markdownlint GH action to use super-linter

MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker/issues/5316

Switched the MarkDown linter GH action to use the super-linter. This gives us the advantage of enabling other linters via a "flip of a switch". The configuration will be more standardized with the super-linter versus having separate GH actions/linters. The super-linter used the markdownlint tool we are already using so there should be no functional changes.


.github/workflows/lint-code-base.yml
```diff
@@ -0,0 +1,31 @@
+name: Lint Code Base
+
+on:
+  pull_request: null
```
NikolaMilosavljevic, line 4: Is this a change in behavior? Previously, linter was supposedly `on` for PRs with a list of file patterns.

MichaelSimons: It is on for PRs with no specific file patterns (e.g. all PRs).  The idea is it allows for the easy enablement of other linters supported by the super-linter.  e.g. scripting, json, yml, etc.

ellahathaway (approved):

NikolaMilosavljevic (approved):

PR #4759, opened 2024-11-26
Update patching guidelines

premun:


Documentation/patching-guidelines.md
```diff
@@ -39,6 +39,8 @@ create one.
 > git format-patch --zero-commit --no-signature -1 -o "$PATCH_DIR/<repo>"
 > ```
 
+**Note:** Please make sure there are no other already existing patches which target same files as the newly added patch.
```
premun, line 42:
```suggestion
**Note:** Please make sure there are no other already existing patches
which target same files as the newly added patch.
```

Documentation/patching-guidelines.md
```diff
@@ -39,6 +39,9 @@ create one.
 > git format-patch --zero-commit --no-signature -1 -o "$PATCH_DIR/<repo>"
 > ```
 
+**Note:** Please make sure there are no other already existing patches
```
MichaelSimons, line 42: Does this break synchronization?  It has been necessary in the past.

premun: It breaks the synchronization, we've only had it once. Why would it be necessary?

MichaelSimons: > Why would it be necessary?

Historically there have been numerous times where a single file is touched by multiple patches.  Patches are typically kept separate per the time they were made and their scope.  This makes patch removal much easier as the backports flow in.

Can't get backport flow soon enough...

premun: In which order would the patches be applied? Alphabetically? I know that the one case I mention was one patch adding a file and another patching it

MichaelSimons: alphabetically - that is why the numeric prefix is included in the guidelines.

premun: https://github.com/dotnet/arcade-services/pull/4188

PR #5630, opened 2026-08-05
Clarify feature band source-build inputs

mthalman:
Higher feature-band source builds can fail when inputs are selected by recency rather than by the versions configured in the VMR revision being built. This clarifies how distro maintainers should interpret lifecycle-relative "current" and "previous" inputs while reproducing a specific branch, tag, or commit. It also provides guidance on cross-build scenarios.

### DIFF
--- a/Documentation/feature-band-source-building.md
+++ b/Documentation/feature-band-source-building.md
@@ -809,8 +809,8 @@ version (= 10.0.10)`
   higher-band source can fail even when each package is valid on its own. For
   example, supplying `10.0.110` inputs to the `v10.0.200` source revision
   configured for `10.0.103` SDK and PSB inputs plus `10.0.104` shared
-  components caused `Microsoft.NETCore.Platforms` resolution to fail. That
-  failure was an unsupported input matrix, not a general 2xx defect.
+  components causes `Microsoft.NETCore.Platforms` resolution to fail because
+  the inputs did not match the versions expected by the build.
 - **Resolution**: Re-read `PrivateSourceBuiltSdkVersion`,
   `PrivateSourceBuiltArtifactsVersion`, and `MicrosoftNETSdkPackageVersion`
   from the exact VMR revision being built. Replace all three inputs as a
--- a/Documentation/feature-band-source-building.md
+++ b/Documentation/feature-band-source-building.md
@@ -801,8 +801,8 @@ not produce shared components`
   `build.sh` script and pass the extracted artifact directory for the
   required 1xx version.
 
-**Error**: Package-source mapping or package-version resolution failures after
-supplying newer 1xx inputs
+**Error**: `NU1102: Unable to find package Microsoft.NETCore.Platforms with
+version (= 10.0.10)`
 
 - **Explanation**: The SDK, PSB artifacts, and shared components form a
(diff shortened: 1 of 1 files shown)


Documentation/feature-band-source-building.md
```diff
@@ -243,17 +280,17 @@ Note: Only the 1xx SDK is guaranteed to build the shared runtime components
 
 - **Bootstrap (any version)**: Two-stage process using Microsoft source-built
   2xx artifacts + Microsoft 2xx SDK + prep script
-- **Initial Release (N.0.200)**: Current source-built 1xx artifacts + current
-  source-built 1xx SDK
+- **Initial Release (N.0.200)**: Current source-built 1xx shared component
+  artifacts + previous source-built 1xx artifacts and SDK
```
lbussell: The `2xx` section and the `3xx/4xx` sections are very similar. It seems like the `2xx` sections `3xx/4xx` could be combined into a `2xx+` section and use the text from the `3xx/4xx` section. Is my understanding correct?

Feel free to dismiss if this is outside the scope of these changes.

Documentation/feature-band-source-building.md
```diff
@@ -754,7 +807,23 @@ not produce shared components`
   from the 1xx band because they provide the necessary runtime and
   foundational components for the SDK.
 - **Resolution**: Include the `--with-shared-components` parameter for the
-  `build.sh` script.
+  `build.sh` script and pass the extracted artifact directory for the
+  required 1xx version.
+
+**Error**: Package-source mapping or package-version resolution failures after
```
lbussell: **nit:** other troubleshooting entries use specific error text here whereas this one doesn't look like a specific error.

Documentation/feature-band-source-building.md
```diff
@@ -754,7 +807,23 @@ not produce shared components`
...
+supplying newer 1xx inputs
+
+- **Explanation**: The SDK, PSB artifacts, and shared components form a
+  supported version matrix. Mixing the latest 1xx inputs with older
+  higher-band source can fail even when each package is valid on its own. For
+  example, supplying `10.0.110` inputs to the `v10.0.200` source revision
+  configured for `10.0.103` SDK and PSB inputs plus `10.0.104` shared
+  components caused `Microsoft.NETCore.Platforms` resolution to fail. That
+  failure was an unsupported input matrix, not a general 2xx defect.
```
lbussell:
```suggestion
  components causes `Microsoft.NETCore.Platforms` resolution to fail because
  the inputs did not match the versions expected by the build.
```

lbussell (approved):

Documentation/feature-band-source-building.md
mthalman: Fixed in cc280dae

Documentation/feature-band-source-building.md
mthalman: Fixed in 6751ffa6

Documentation/feature-band-source-building.md
mthalman: Fixed in b30876ef

lbussell (approved):

PR #5653, opened 2026-08-21
Fix markdown link checker action reference

mthalman:
## Problem

The **Check Markdown Links** workflow fails before creating any jobs because it pins `umbrelladocs/action-linkspector` to a commit SHA. The dotnet enterprise Actions policy only permits this action through the `umbrelladocs/action-linkspector@v1` reference.

## Solution

Change the action reference to the enterprise-allowed `@v1` tag. Add a comment documenting that the tag is intentional and should not be replaced with a SHA unless the enterprise policy changes.

### DIFF
--- a/.github/workflows/check-markdown-links.yml
+++ b/.github/workflows/check-markdown-links.yml
@@ -17,7 +17,8 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Check markdown links
-        uses: umbrelladocs/action-linkspector@568ec8d29fa92b31fd9ea5381e155c51e922af83 # v1.5.5
+        # The enterprise Actions allowlist requires this tag; do not pin a SHA unless the policy changes.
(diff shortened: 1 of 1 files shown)


lbussell:
Can we update the policy instead? We should be able to pin it to the sha. /cc @blowdart

blowdart:
Ah the problem is the allow list of actions?

That should be updated to limit on versions. So either the allow list needs updating to have the sha, or to just allow umbrelladocs/action-linkspector@*.

Obviously being stricter and using the sha in the allow list is safer, however that would mean any time a workflow gets updated with a new version a repo admin will have to update the allow list, which is a huge toil. I'd suggest using the @* approach here as a compromise.

mthalman:
@mmitche - Can you comment on the best way to proceed here?

mmitche:
@blowdart I'm fine with that as long as these workflows never have anything but PR-read.

mmitche:
Otherwise, a force-push of a tag can change behavior without review.

mmitche (approved):

